PR #42, opened 2015-01-22
add support for citext (case insensitive text extension)

troyk:
Support citext http://www.postgresql.org/docs/9.4/static/citext.html

citext is a commonly installed extension for Postgres and while it appears to work without explicit inclusion, the array variant `citext[]` does not (casts to binary instead of list).  This pull request remedies this.  I did not add a test because it would require the installation of the citext extension.  Open to any thoughts or feedback, have tested in my app with postgres versions 9.1-9.4.

Thanks,

Troy


josevalim:
:+1:

PR #45, opened 2015-02-03
Add test for date value of -1 (1999-12-31)

hamiltop:
Test for situation in https://github.com/ericmj/postgrex/issues/44


ericmj:
:heart:

PR #47, opened 2015-02-05
citext[] type causes 'no extension found for oid' error

troyk:
added citextsend to @senders to fix.

The encoder was already there, just no entry in @senders


josevalim:
:shipit:

PR #75, opened 2015-05-29
Use `to_char_list/1` when possible

whatyouhide:
Instead of checking if the hostname that we pass to `:gen_tcp.connect/3` is a binary and converting it to a charlist if it is (`String.to_char_list/1`), we can call `to_char_list/1` which converts anything to a charlist according to the `List.Chars` protocol.

Avoiding `to_char_list/1` may have been intentional though; if this is the case, feel free to close this PR (and maybe explain to me the reason behind this choice so that I can learn something myself :)).


josevalim:
:+1:

PR #102, opened 2015-10-17
Async query

gjaldon:
Closes https://github.com/ericmj/postgrex/issues/101

Here's the PR for async queries. Thanks a lot for the help, @fishcakez!

### DIFF
--- a/lib/postgrex/connection.ex
+++ b/lib/postgrex/connection.ex
@@ -108,7 +108,8 @@ defmodule Postgrex.Connection do
   """
   def async_query(pid, statement, params) do
     message = {:query, statement, params}
-    process = GenServer.whereis(pid) || raise "No process is associated with #{inspect pid}"
+    process = GenServer.whereis(pid) ||
+      raise ArgumentError, "No process is associated with #{inspect pid}"
     monitor = Process.monitor(process)
     from = {self(), monitor}
 
--- a/lib/postgrex/connection.ex
+++ b/lib/postgrex/connection.ex
@@ -85,12 +85,10 @@ defmodule Postgrex.Connection do
     message = {:query, statement, params}
     timeout = opts[:timeout] || @timeout
     case Connection.call(pid, message, timeout) do
-      %Postgrex.Result{} = res ->
-        {:ok, res}
-      %Postgrex.Error{} = err ->
-        {:error, err}
       {:exit, reason} ->
         exit({reason, {__MODULE__, :query, [pid, statement, params, opts]}})
+      result ->
+        result
     end
   end
 
@@ -260,7 +258,7 @@ defmodule Postgrex.Connection do
     end
   end
 
-  def handle_cast({{:query, _, _} = command, {_, monitor} = from}, s) do
+  def handle_cast({{:query, _, _} = command, {_, _} = from}, s) do
     handle_call(command, from, s)
   end
 
--- a/lib/postgrex/connection.ex
+++ b/lib/postgrex/connection.ex
@@ -95,6 +95,28 @@ defmodule Postgrex.Connection do
   end
 
   @doc """
+  Works like `query/3` and `query/4` but is asynchronous. This returns a `%Task{}`
+  and is useful for starting multiple async queries and then waiting on the result
+  later on.
+
+  When server name passed has no process associated, it will raise an error.
+
+  ## Examples
+
+      task = %Task{} = Postgrex.Connection.async_query(pid, "SELECT title FROM posts", [])
+      Task.await(task)
+  """
+  def async_query(pid, statement, params) do
+    message = {:query, statement, params}
+    process = GenServer.whereis(pid) || raise "No process is associated with #{inspect pid}"
+    monitor = Process.monitor(process)
+    from = {self(), monitor}
+
+    :ok = Connection.cast(pid, {message, from})
+    %Task{ref: monitor}
+  end
+
+  @doc """
   Runs an (extended) query and returns the result or raises `Postgrex.Error` if
   there was an error. See `query/3`.
   """
@@ -235,6 +257,10 @@ defmodule Postgrex.Connection do
     else
       {:noreply, s}
     end
+  end
+
+  def handle_cast({{:query, _, _} = command, {_, monitor} = from}, s) do
+    handle_call(command, from, s)
   end
 
   @doc false
--- a/lib/postgrex/connection.ex
+++ b/lib/postgrex/connection.ex
@@ -99,6 +99,9 @@ defmodule Postgrex.Connection do
 
   When server name passed has no process associated, it will raise an error.
 
+  This function expects a pid as first argument for Elixir 1.0. In Elixir 1.1, a pid
+  or a name could be passed as first argument.
+
   ## Examples
 
       task = %Task{} = Postgrex.Connection.async_query(pid, "SELECT title FROM posts", [])
@@ -106,8 +109,14 @@ defmodule Postgrex.Connection do
   """
   def async_query(pid, statement, params) do
     message = {:query, statement, params}
-    process = GenServer.whereis(pid) ||
-      raise ArgumentError, "No process is associated with #{inspect pid}"
+    process = cond do
+        is_pid(pid) ->
+          pid
+        function_exported?(GenServer, :whereis, 1) ->
+          GenServer.whereis(pid) || raise ArgumentError, "no process is associated with #{inspect pid}"
+        true ->
+          raise ArgumentError, "requires Elixir 1.1 when passing server name as first argument"
+      end
     monitor = Process.monitor(process)
     from = {self(), monitor}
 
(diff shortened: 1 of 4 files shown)


josevalim:
Looks great! I think though we should use exit(:noproc) instead of raising,
or whatever gen server does!

On Saturday, October 17, 2015, Gabriel Jaldon <notifications@github.com>
wrote:

> Closes #101 <https://github.com/ericmj/postgrex/issues/101>
>
> Here's the PR for async queries. Thanks a lot for the help, @fishcakez
> <https://github.com/fishcakez>!
> ------------------------------
> You can view, comment on, or merge this pull request online at:
>
>   https://github.com/ericmj/postgrex/pull/102
> Commit Summary
>
>    - Support for async_query/3
>    - Testing async query
>
> File Changes
>
>    - *M* lib/postgrex/connection.ex
>    <https://github.com/ericmj/postgrex/pull/102/files#diff-0> (27)
>    - *M* test/query_test.exs
>    <https://github.com/ericmj/postgrex/pull/102/files#diff-1> (16)
>    - *M* test/test_helper.exs
>    <https://github.com/ericmj/postgrex/pull/102/files#diff-2> (6)
>
> Patch Links:
>
>    - https://github.com/ericmj/postgrex/pull/102.patch
>    - https://github.com/ericmj/postgrex/pull/102.diff
>
> —
> <https://github.com/ericmj/postgrex/pull/102>.
>


-- 


*José Valim*
www.plataformatec.com.br
Skype: jv.ptec
Founder and Director of R&D

fishcakez:
A GenServer.cast raises an ArgumentError when a local name has no process.
There are no side effects when cast raises in this way so we don't need to
exit/1. The same is true here, there are no side effects, so I think it is
good to raise an ArgumentError as it is ok to rescue.

josevalim:
Thanks James!

-- 


test/query_test.exs
```diff
@@ -494,4 +494,20 @@ defmodule QueryTest do
       assert_receive [[:void]], 1000
     end)
   end
+
+  test "async query", context do
+    task1 = async_query("SELECT true, false", [])
+    task2 = async_query("SELECT 42", [])
+
+    %{rows: rows} = Task.await(task1)
```
fishcakez: The return value should be {:ok, struct} | {:error, struct}, so may need to refactor how replies are sent. The reply/2 needs to send the actual result of query/4

gjaldon: Just pushed a fix for this @fishcakez!

fishcakez: I'm not familiar enough with postgrex to review that last one sorry

lib/postgrex/connection.ex
```diff
@@ -85,16 +85,38 @@ defmodule Postgrex.Connection do
     message = {:query, statement, params}
     timeout = opts[:timeout] || @timeout
     case Connection.call(pid, message, timeout) do
-      %Postgrex.Result{} = res ->
-        {:ok, res}
-      %Postgrex.Error{} = err ->
-        {:error, err}
       {:error, kind, reason, stack} ->
         :erlang.raise(kind, reason, stack)
```
fishcakez: Raising an exception with a stack trace from another process? Please let's not do this as the stack trace is lying and going to confuse. Not related to this issue though.

test/query_test.exs
gjaldon: Ok no prob! Will wait for @ericmj's review then.

lib/postgrex/connection.ex
gjaldon: Ok so I should just remove this then and just return the result of `Connection.call/3`? How will it confuse, btw @fishcakez?

ericmj: Maybe we should just log the error with the stacktrace at the source.

mix.exs
```diff
@@ -4,7 +4,7 @@ defmodule Postgrex.Mixfile do
   def project do
     [app: :postgrex,
      version: "0.9.2-dev",
-     elixir: "~> 1.0",
+     elixir: "~> 1.1",
```
ericmj: I'm not sure we want to push postgrex to 1.1 yet, it just came out. Can you reimplement `GenServer.whereis`?

josevalim: Please let's not reimplement GenServer.whereis. What if we leave it as 1.0 and say that particular feature requires 1.1?

fishcakez: Given that `process` will nearly always be a `pid` we could do:
```elixir
cond do
  is_pid(process) -> pid
  function_exported?(GenServer, :whereis, 1) -> GenServer.whereis(pid)
  true -> raise "Requires 1.1"
end
```

josevalim: Sounds like a good idea. Honestly, we can just call GenServer.whereis unless we have a process.

josevalim: And make it clear the docs. "Expects a pid for Elixir 1.0. A pid or a name can be given from Elixir 1.1."

lib/postgrex/connection.ex
fishcakez: I submitted a fix #103

gjaldon:
@ericmj ready for review. I rebased this PR on top of master and made this PR Elixir-1.0 compatible as you guys suggested.

lib/postgrex/connection.ex
```diff
@@ -85,16 +85,46 @@ defmodule Postgrex.Connection do
...
+  """
+  def async_query(pid, statement, params) do
+    message = {:query, statement, params}
+    process = cond do
+        is_pid(pid) ->
+          pid
+        function_exported?(GenServer, :whereis, 1) ->
+          GenServer.whereis(pid) || raise ArgumentError, "No process is associated with #{inspect pid}"
+        true ->
+          raise ArgumentError, "Requires Elixir 1.1 when passing server name as first argument"
```
ericmj: Start error messages with lower case.

ericmj:
Left a small comment. Ready to merge when it's fixed and @josevalim or @fishcakez sign off.

gjaldon:
Error messages format has been fixed!

lib/postgrex/connection.ex
```diff
@@ -85,16 +85,46 @@ defmodule Postgrex.Connection do
...
+
+  This function expects a pid as first argument for Elixir 1.0. In Elixir 1.1, a pid
+  or a name could be passed as first argument.
+
+  ## Examples
+
+      task = %Task{} = Postgrex.Connection.async_query(pid, "SELECT title FROM posts", [])
+      Task.await(task)
+  """
+  def async_query(pid, statement, params) do
```
fishcakez: Can we get a spec on this and docs on the possible results of the task? Its important to document the `{:exit, reason}` result as it can't happen in query/3.

lib/postgrex/connection.ex
```diff
@@ -85,16 +85,46 @@ defmodule Postgrex.Connection do
...
     timeout = opts[:timeout] || @timeout
     case Connection.call(pid, message, timeout) do
-      %Postgrex.Result{} = res ->
-        {:ok, res}
-      %Postgrex.Error{} = err ->
-        {:error, err}
       {:exit, reason} ->
         exit({reason, {__MODULE__, :query, [pid, statement, params, opts]}})
+      result ->
+        result
```
fishcakez: Will need this change in `listen`/`unlisten` to as the `%Postgrex.Error{}` can will now be sent as `{:error, %Postgrex.Error{}}`. There should be a test for these if its possible.

fishcakez:
Here we send a reply on tcp/ssl error: https://github.com/ericmj/postgrex/blob/fb1867540619e669b1a746f17d64c22e1af48540/lib/postgrex/utils.ex#L8, we need to wrap the error exception in an `:error` tuple.

lib/postgrex/connection.ex
```diff
@@ -85,16 +85,46 @@ defmodule Postgrex.Connection do
...
+        function_exported?(GenServer, :whereis, 1) ->
+          GenServer.whereis(pid) || raise ArgumentError, "no process is associated with #{inspect pid}"
+        true ->
+          raise ArgumentError, "requires Elixir 1.1 when passing server name as first argument"
+      end
+    monitor = Process.monitor(process)
+    from = {self(), monitor}
+
+    :ok = Connection.cast(pid, {message, from})
+    %Task{ref: monitor}
```
josevalim: Please add a TODO above this line with something like this:

    # TODO: Add :owner field to task once we support Elixir 1.2 onwards

ericmj: We can support both with

```elixir
if Map.has_key?(%Task{}, :owner) do
  defp new_key(ref), do: %Task{ref: ref, owner: self}
else
  defp new_key(ref), do: %Task{ref: ref}
end
```

ericmj:
Where are we on this PR?

fishcakez:
I left some comments about spec's and listen/unlisten error handling.

josevalim:

PR #110, opened 2015-11-15
Put formatted reason before raw term in error messages

fishcakez:


josevalim:

PR #128, opened 2015-12-10
Use DBConnection

fishcakez:
This is WIP.

Issues:

- [x] There are failing tests because `Postgrex.Connection.stop/1` has not been replaced. I could not come up with a good method to stop a supervisor without requiring Elixir 1.2.
- [x] The parameter map is always sent to the client process and back but rarely used. Previously they were not sent. What are the use cases for this function?
- [x] Timeouts are not handled correctly as a socket receive timeout of 5000ms is always used, which means a longer timeout will not be honoured. This can be solved by moving the timeout from the protocol's state to its request "status" term.

There are two optimisations in this change because the `DBConnection` API is not exactly equivalent to the old `Postgrex.Connection`:

* Notifications are sent as soon as they are received and are not accumulated.
* Type information for a query can be looked up without blocking the connection. I am unsure how useful it is as the client will likely be holding on to the connection about to execute.

Will need to do benchmark/profile to check performance does not regress before merging.

I have not added any new features in this PR and would prefer these to come one by one in separate PRs. The features (pooling, transactions, network failure handling, idle timeout) require implementing simple callbacks, documenting options and/or wrapping `DBConnection` calls so are almost free.

### DIFF
--- a/lib/postgrex/connection.ex
+++ b/lib/postgrex/connection.ex
@@ -32,8 +32,6 @@ defmodule Postgrex.Connection do
     * `:password` - User password (default PGPASSWORD);
     * `:parameters` - Keyword list of connection parameters;
     * `:timeout` - Connect timeout in milliseconds (default: `#{@timeout}`);
-    * `:idle_timeout` - Idle timeout to ping postgres to maintain a connection
-    (default: `#{@idle_timeout}`)
     * `:ssl` - Set to `true` if ssl should be used (default: `false`);
     * `:ssl_opts` - A list of ssl options, see ssl docs;
     * `:socket_options` - Options to be given to the underlying socket;
@@ -41,6 +39,14 @@ defmodule Postgrex.Connection do
     * `:extensions` - A list of `{module, opts}` pairs where `module` is
       implementing the `Postgrex.Extension` behaviour and `opts` are the
       extension options;
+    * `:idle_timeout` - Idle timeout to ping postgres to maintain a connection
+    (default: `#{@idle_timeout}`)
+    * `:backoff_start` - The first backoff interval when reconnecting (default:
+    `200`);
+    * `:backoff_max` - The maximum backoff interval when reconnecting (default:
+    `15_000`);
+    * `:backoff_type` - The backoff strategy when reconnecting, `:stop` for no
+    backoff and to stop (see `:backoff`, default: `:jitter`)
     * `:transactions` - Set to `:strict` to error on unexpected transaction
     state, otherwise set to `naive` (default: `:naive`);
     * `:pool_mod` - The pool module to use, see `DBConnection`, it must be
--- a/lib/postgrex/connection.ex
+++ b/lib/postgrex/connection.ex
@@ -7,14 +7,9 @@ defmodule Postgrex.Connection do
   you need to start a separate (notifications) connection.
   """
 
-  use Connection
-  alias Postgrex.Protocol
-  require Logger
+  alias Postgrex.Query
 
   @timeout 5000
-
-  defstruct [parameters: nil, protocol: nil, queue: :queue.new(), client: nil,
-             timer: nil]
 
   ### PUBLIC API ###
 
@@ -40,19 +35,7 @@ defmodule Postgrex.Connection do
   """
   @spec start_link(Keyword.t) :: {:ok, pid} | {:error, Postgrex.Error.t | term}
   def start_link(opts) do
-    Connection.start_link(__MODULE__, Postgrex.Utils.default_opts(opts))
-  end
-
-  @doc """
-  Stop the process and disconnect.
-
-  ## Options
-
-    * `:timeout` - Call timeout (default: `#{@timeout}`)
-  """
-  @spec stop(pid, Keyword.t) :: :ok
-  def stop(pid, opts \\ []) do
-    Connection.call(pid, :stop, opts[:timeout] || @timeout)
+    DBConnection.start_link(Postgrex.Protocol, Postgrex.Utils.default_opts(opts))
   end
 
   @doc """
@@ -82,7 +65,8 @@ defmodule Postgrex.Connection do
   """
   @spec query(pid, iodata, list, Keyword.t) :: {:ok, Postgrex.Result.t} | {:error, Postgrex.Error.t}
   def query(pid, statement, params, opts \\ []) do
-    run(pid, &Protocol.query(&1, statement, params, &2), opts)
+    query = %Query{name: "", statement: statement, params: params}
+    DBConnection.query(pid, query, opts)
   end
 
   @doc """
@@ -91,10 +75,8 @@ defmodule Postgrex.Connection do
   """
   @spec query!(pid, iodata, list, Keyword.t) :: Postgrex.Result.t
   def query!(pid, statement, params, opts \\ []) do
-    case query(pid, statement, params, opts) do
-      {:ok, res}    -> res
-      {:error, err} -> raise err
-    end
+    query = %Query{name: "", statement: statement, params: params}
+    DBConnection.query!(pid, query, opts)
   end
 
   @doc """
@@ -115,7 +97,7 @@ defmodule Postgrex.Connection do
   """
   @spec prepare(pid, iodata, iodata, Keyword.t) :: {:ok, Postgrex.Query.t} | {:error, Postgrex.Error.t}
   def prepare(pid, name, statement, opts \\ []) do
-    run(pid, &Protocol.prepare(&1, name, statement, &2), opts)
+    DBConnection.prepare(pid, %Query{name: name, statement: statement}, opts)
   end
 
   @doc """
@@ -124,10 +106,7 @@ defmodule Postgrex.Connection do
   """
   @spec prepare!(pid, iodata, iodata, Keyword.t) :: Postgrex.Query.t
   def prepare!(pid, name, statement, opts \\ []) do
-    case prepare(pid, name, statement, opts) do
-      {:ok, query}  -> query
-      {:error, err} -> raise err
-    end
+    DBConnection.prepare!(pid, %Query{name: name, statement: statement}, opts)
   end
 
   @doc """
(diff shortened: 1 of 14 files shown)


josevalim:

josevalim:
>  There are failing tests because Postgrex.Connection.stop/1 has not been replaced. I could not come up with a good method to stop a supervisor without requiring Elixir 1.2.

Chants: one two! one two! one two!

We gotta give what the crowds ask for.

ericmj:
> The parameter map is always sent to the client process and back but rarely used. Previously they were not sent. What are the use cases for this function?

Not sure I understand, what was changed and why? The parameters are used to get connection parameters which can be useful to get information like the server timezone.

fishcakez:
The full parameter map is passed to the client process on every checkout.
So when doing queries it is still sent and if using sbroker passed via the
broker. On master it is not always sent and fetched only on explicit
request.

By use cases, I rather meant to find out how often it is called and if it
should work without a pool checkout. If it is not very often I will add
get/format_status to DBConnection. This will provide parameters, socket
details and other connection information that will be nicely formatted in
observer and other introspection tools. I planned to add that at some point
but we could use it to fetch parameters without always including them in
the state. It would make the parameter call more expensive though.

It is difficult to abstract using callbacks in two processes to something
that is generic so state is only accessed in one process at a time in
DBConnection.

ericmj:
Sorry, I don't understand this at all. Are you talking about connection parameters? 

> The full parameter map is passed to the client process on every checkout.

What is the reason for this? Why was it changed from a function doing a `GenServer.call` like it was previously?

fishcakez:
> What is the reason for this? Why was it changed from a function doing a GenServer.call like it was previously?

There are no `GenServer.call`s in `DBConnection` because only one process can hold the state at a time. This means there isn't a good way to handle a call as the state could be checked out in a client when the call arrives. However `DBConnection.Connection` will log an info message if a `handle_info` message arrives while the process is checked out. Socket messages don't arrive when checked out because the socket is set to `active: false` and the message queue flushed for socket messages during the `checkout` callback.

DBConnection is designed like this so that the state (i.e. Postgrex.Protocol.state) is fetched once at the start of a transaction and then stored in the client process until commit/rollback is handled. The callbacks (e.g. Postgrex.Protocol.execute/3) are all called in the client process and state changes are saved in the pdict. This means all the logic and state is held in one process for transactions and transactions are independent of pooling and disconnect/reconnection attempts. This independence makes the logic much simpler. I think the pdict usage is well worth it.

I should have mentioned this design decision as it turns out to be an optimisation for transactions or prepare/execute/close because state is copied once end the socket only needs to be deactivated once when doing mutplie calls during a pool checkout, such as a transaction. However this PR does not add support for transactions or pooling so I forgot to mention it.

fishcakez:
I will fix the three issue tomorrow using a newer release of DBConnection and keep the requirement at Elixir 1.0.

* `stop/1` will be implemented by writing a custom supervisor to sit on top of a pool which would normally have a Supervisor as the top process.
* Use a global read/write concurrent ets table to store parameters using a (monitor) reference as key stored in protocol state.
* timeout fix as stated in initial comment.

fishcakez:
I haven't been able to fix `stop/1`. Even when doing as described there is still a problem. If the named process is not the top level process (i.e. the pid returned by start_link is not the named process by one below is) it is unclear which process should be called. This can not be solved with 1.2 either :(. The other two issues have been resolved.

The ability to use a prepared query prepared by another process has been added in 9e7b93c, which means using prepared queries with a pool just works. The method is efficient in that describing the query is skipped and the type information is reused. However it is inefficient in that a failed execute has to be sent and received to know the query needs to be prepared. This can be made more efficient at a later time. The (query) parameters have been outside the query struct to support this and will be needed for Ecto.

josevalim:
Can we leave stop support outside of DBConnection and just call it directly
from Ecto? Or can we add it and say it only works with Elixir 1.2?



-- 


fishcakez:
Ecto doesnt need stop because it puts the pool under its own supervisor,
Ecto stops its supervisor using an exit signal from parent

fishcakez:
1.2 has the same issue as above though too

josevalim:
So why do we need stop? Can't we tell people to supervise it or something
they need it?

On Wednesday, December 16, 2015, James Fish <notifications@github.com>
wrote:

> Ecto doesnt need stop because it puts the pool under its own supervisor,
> Ecto stops its supervisor using an exit signal from parent
>
> On 16 December 2015 at 20:30, José Valim <notifications@github.com
> <javascript:_e(%7B%7D,'cvml','notifications@github.com');>> wrote:
>
> > Can we leave stop support outside of DBConnection and just call it
> directly
> > from Ecto? Or can we add it and say it only works with Elixir 1.2?
> >
> >
> >
> > --
> >
> >
> > *José Valimwww.plataformatec.com.br
> > <http://www.plataformatec.com.br/>Founder and Director of R&D*
> >
> > —
> > Reply to this email directly or view it on GitHub
> > <https://github.com/ericmj/postgrex/pull/128#issuecomment-165230400>.
> >
>
> —
> <https://github.com/ericmj/postgrex/pull/128#issuecomment-165236647>.
>


-- 


fishcakez:
@josevalim, @ericmj and I discussed on IRC and its gone.

josevalim:
Glad to hear @fishcakez! The best way to solve problems ;)

fishcakez:
Moebius currently uses stop/1 so we may need to provide a simple one for
one pool as well as ownership pool. I am hoping that the pooling will be
acceptable to them though and we don't need to worry about it.

josevalim:
@fishcakez if they start the db_connection inside their own simple_one_for_one supervisor, wouldn't that be enough?

fishcakez:
@josevalim if they start one like that how will they stop it? Safely?

josevalim:
terminate_child?

On Thursday, December 17, 2015, James Fish <notifications@github.com> wrote:

> @josevalim <https://github.com/josevalim> if they start one like that how
> will they stop it? Safely?
>
> —
> <https://github.com/ericmj/postgrex/pull/128#issuecomment-165572445>.
>


-- 


fishcakez:
There is no guarantee to call it unfortunately though. It is not too hard to add support but tests would need some work. Maybe it would be motivation to clean up the tests as they have got in bad shape favouring short term wins :(. I would prefer to wait and see.

josevalim:
It will be safe with the owner feature. If the parent crashes before, we
will notice and act on it.

Also, take your time to clean stuff up or remove things you don't want to
support because it is a "short term win". This happens a lot: sometimes we
are just gathering the requirements and they pile up, at some point we need
to look back and think about what makes sense. That's certainly what
happened with Ecto and in fact what led to DBConnection. :)

On Thursday, December 17, 2015, James Fish <notifications@github.com> wrote:

> There is no guarantee to call it unfortunately though. It is not too hard
> to add support but tests would need some work. Maybe it would be motivation
> to clean up the tests as they have got in bad shape favouring short term
> wins :(. I would prefer to wait and see.
>
> —
> <https://github.com/ericmj/postgrex/pull/128#issuecomment-165581601>.
>


-- 


fishcakez:
I've added transactions and all the configurable features DBConnection provides. DBConnection provides some extra low level calls (prepare_execute, execute_close) but I have skipped those for now. It is easy to add if someone wants them later down the line.

josevalim:
Is this good to go then? :dancers:

fishcakez:
I would like to:
- [ ] Fix registered naming poolboy in DBConnection (it crashes)
- [ ] Check it works on the moebius test suite. I have 6 failing tests on a branch using sbroker to track down. All related to documents and transactions. I am using Postgrex transactions in strict mode, which behaves differently to the current transactions.
- [ ] Benchmark to check no regression (query/3 uses a different message order, which I hope does not effect performance)
- [ ] Release new DBConnection and updated mix.exs

fishcakez:
Fixed a bug where ArgumentErrors weren't being raised and added to_string implementation for Query for concise logging.

fishcakez:
Using this branch and a DBConnection fork of Ecto reduces basic query (`Repo.insert!/1)` time by ~15% and transaction overhead (`Repo.transaction(fn -> nil end)`) time by ~50% using a local postgres server with poolboy.

The large reduction in transaction overheard comes from more efficient use of postgres (removed 2 round trips) and more efficient pooling (removed 2 calls, 2 casts and 2 `:inet.setopts/2`). It should be possible to make similar gains for cached queries inside transactions that require minimal decoding by using prepared queries.

josevalim:
This is beautiful! I can't wait to get this merged and start the final push towards Ecto 2.0.

fishcakez:
This is ready for review.

lib/postgrex/connection.ex
```diff
@@ -140,33 +172,45 @@ defmodule Postgrex.Connection do
...
   ## Options
 
-    * `:timeout` - Call timeout (default: `#{@timeout}`)
+    * `:pool_timeout` - Time to wait in the queue for the connection
+    (default: `#{@pool_timeout}`)
+    * `:queue` - Whether to wait for connection in a queue (default: `true`);
+    * `:timeout` - Execute request timeout (default: `#{@timeout}`);
     * `:decode`  - Decode method: `:auto` decodes the result and `:manual` does
     not (default: `:auto`)
+    * `:decode_mapper` - Fun to map each row in the result to a term, see
```
josevalim: Do we care about making this public? Would using a mapper offer any advantage over `decode: :manual`? If so, maybe we should get rid of `decode: :manual` altogether?

lib/postgrex/connection.ex
```diff
@@ -177,312 +221,118 @@ defmodule Postgrex.Connection do
...
 
-  defp handle_next({:checkout, timeout}, from, buffer, s) do
-    handle_checkout(timeout, from, buffer, s)
+    * `:pool_timeout` - Time to wait in the queue for the connection
+    (default: `#{@pool_timeout}`)
+    * `:queue` - Whether to wait for connection in a queue (default: `true`);
+    * `:timeout` - Transaction timeout (default: `#{@timeout}`);
+    * `:pool` - The pool module to use, must match that set on
+    `start_link/1`, see `DBConnection`
+    * `:proxy` - The proxy module for the request, if any, see
```
josevalim: :heart:

mix.exs
```diff
@@ -27,6 +27,7 @@ defmodule Postgrex.Mixfile do
     [{:ex_doc, only: :dev},
      {:earmark, only: :dev},
      {:decimal, "~> 1.0"},
+     {:db_connection, "~> 0.1.6"},
      {:connection, "~> 1.0"}]
```
josevalim: Should we remove this dep?

fishcakez: We can't, its used for notification process

lib/postgrex/connection.ex
fishcakez: We dont need manual anymore. If we can make this backwards incompatible
change may be able to tidy up encode/decode a bit.

If we were to need it, likely it would make sense to add to db_connection.

mix.exs
josevalim: Ah, I see. :D

lib/postgrex/connection.ex
josevalim: So kill it with :fire:. :)

fishcakez:
I've cleaned up the encode/decode so the db_connection protocol is always used.

PR #139, opened 2016-01-07
Delay row decoding

fishcakez:
Only traverse and reverse the row once, instead of twice.


josevalim:

PR #140, opened 2016-01-08
Skip describe when executing described query

fishcakez:
Previously when executing a query prepared on another connection, the connection would try to execute, prepare and then execute the query. This took 3 round trips. This change does prepare and execute in 1 round trip when the query was already described. Therefore when we use prepared cached queries in Ecto we only need to do one 2 round trip to postgres per cached query.

Previously when executing a prepared query on a connection with a different query prepared with the same name, the different query would be executed. This change ensures the names and statements match before executing.


fishcakez:
The ets query table is supposed to guarantee that if the query is in the table then it is prepared on backend. However if the query isn't in the table is does not guarantee that the query is not prepared on the backend. If we want to limit the number of prepared queries per connection we will first need to provide this guarantee.

josevalim:

PR #144, opened 2016-01-11
Test savepoint transactions

fishcakez:
The tests don't actually use savepoints yet, which means savepoint transactions aren't usable in postgrex itself.


fishcakez:
I've fixed the tests and added docs.

josevalim:

fishcakez:
> The tests don't actually use savepoints yet, which means savepoint transactions aren't usable in postgrex itself.

Clearly the tests show it is usable, I just derped earlier trying to nest savepoint transaction inside a normal one :D.

PR #148, opened 2016-01-16
Optimize tag decoding

josevalim:
The first commit moves the enum based implementation into binary decoding. The new implementation assumes that once an integer is found in the tag, only integers will be expected. I.e. we can have "INSERT 0 1" and "SELECT 1" and "ROLLBACK" but never "COMMAND 0 ANOTHER 1". @ericmj is this assumption correct?

The second commit inlines common commands: insert, select, update, delete, begin, rollback and commit.

### DIFF
--- a/lib/postgrex/protocol.ex
+++ b/lib/postgrex/protocol.ex
@@ -897,17 +897,25 @@ defmodule Postgrex.Protocol do
   defp tag(:ssl), do: :ssl
 
   defp decode_tag(tag) do
-    words = :binary.split(tag, " ", [:global])
-    words = Enum.map(words, fn word ->
-      case Integer.parse(word) do
-        {num, ""} -> num
-        :error -> word
-      end
-    end)
+    decode_tag(tag, "")
+  end
 
-    {command, nums} = Enum.split_while(words, &is_binary(&1))
-    command = Enum.join(command, "_") |> String.downcase |> String.to_atom
-    {command, List.last(nums)}
+  defp decode_tag(<<?\s, i, t::binary>>, acc) when i in ?0..?9,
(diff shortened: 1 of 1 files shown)


josevalim:
The motivation is that @fishcakez found decode_tag sometimes was taking up to 17% of the stack when profiling.

fishcakez:
It might be that tracing interferes with the real data. I have rerun the benchmark I used before and got between 2.5% and 5% reduction in time using this branch. So likely I need to find a better benchmark. Once we do a benchmark with a remote postgres its hopefully that the performance difference on master compared to v0.10.0 or this branch become meaningless.

josevalim:
In any case, I think we should at least merge the first commit. The implementation is better. The second commit is inlining and therefore is optional.

fishcakez:
Yeah definitely I was just wondering about the large % of time and not being able to say exactly how much reduction it gave. The docs say the format is `COMMAND`, `COMMAND rows` or  `INSERT oid rows`.

fishcakez:
http://www.postgresql.org/docs/9.5/static/protocol-message-formats.html See CommandComplete. I think we should just special case the listed commands with a row value, otherwise just do lowercase -> to_atom (if not a common command).

fishcakez:
Also lets use `:erlang.binary_to_integer` instead because we never need the tail except for `INSERT`. For `INSERT` we can just skip the integers until we get a space then to the BIF on the tail.

josevalim:
@fishcakez thanks. pushed a new implementation!

fishcakez:
The stack looks much better now. The failure on the branch is because its no based on master. :shipit:

josevalim:
Merging this since @fishcakez validated the assumption was correct. :)

PR #151, opened 2016-01-17
Handle parse error when part of handle_execute(_close)

### LINKED ISSUE #149: Unexpected msg_ready packet
When using the Ecto ownership pool an unexpected msg_ready packet be received when rolling back the transaction: https://travis-ci.org/elixir-lang/ecto/jobs/102873680#L651

This issues blocks `0.11.0`

fishcakez:
Previously we did not handle the case where a parse error occurs and a `sync` has already been sent. Another `sync` was sent and only one `ready` was received. This left another `ready` packet in the buffer and crashes the next request.

Closes #149.


josevalim:
:shipit:

PR #152, opened 2016-01-17
Only describe queries that are already prepared

### LINKED ISSUE #150: Only describe when preparing an already prepared statement
When preparing a query that already exists on the connection an error occurs. Instead we should only describe the query. See: https://travis-ci.org/elixir-lang/ecto/jobs/102873680#L642

This issues blocks `0.11.0`

fishcakez:
This change means the connection prepared query cache is now in sync with the backend and error handling (of duplicate prepared queries) relies on the backend's error (along with other errors). There is a special case where we must delete unnamed queries if an error occurs.

Closes #150.


josevalim:
:shipit:

PR #156, opened 2016-01-31
Support PgBouncer transaction/statement pooling

### LINKED ISSUE #155: Add an option to always use unnamed queries
When using pgbouncer and transaction pooling the pg backend is only the same for the duration of a transaction and so named prepared queries will not work over multiple transactions. Even when forcing unnamed queries we can still cache the parameter/row descriptions and do one round trip when using Ecto's cache.

We will also need to stop closing unnamed queries in `execute_close` as the close may go to a different backend. This is ok because unnamed queries can be rebound without closing them.

We will also need to use unnamed queries for transaction/savepoints when forcing unnamed queries.

/cc @stevedomin

fishcakez:
Closes #155

### DIFF
--- a/README.md
+++ b/README.md
@@ -90,6 +90,17 @@ query("select nextval($1::text::regclass)", ["some_sequence"])
 query("select nextval($1)", [sequence_oid])
 ```
 
+## PgBouncer
+
+When using PgBouncer with transaction or statement pooling named prepared
+queries can not be used because the bouncer may route requests from the same
+postgrex connection to different PostgreSQL backend processes and discards named
+queries after the transactions closes. To force unnamed prepared queries:
+
+```elixir
+Postgrex.start_link(prepare: :unnamed)
+```
+
 ## Contributing
 
 To contribute you need to compile Postgrex from source and test it:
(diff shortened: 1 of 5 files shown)


josevalim:

fishcakez:
When using this branch, PgBouncer with transaction pooling and Ecto (unchanged) the only test failure is: https://github.com/elixir-lang/ecto/blob/master/integration_test/pg/storage_test.exs#L40. The message is "ERROR (protocol_violation) custom reason". This means we are still caching the query descriptions and doing single roundtrips on cached queries without blocking the connection to encode parameters.

fishcakez:
Do we want this for Ecto 2.0? Should we fix the test in Ecto?

I am on the fence about this because I know the issue exists for any user of PgBouncer (or similar) but have not been able to locate anyone using it with Ecto. It would be great if a user of PgBouncer, or similar tool, could step forward to request this feature.

josevalim:
@fishcakez  honestly, I would merge this. If only the storage test is failing in Ecto, then surely, someone can fix the storage test, but this is a long way to go. :+1:

fishcakez:
The failing test in Ecto is only when running tests with extra config `prepare: :unnamed` and directing Ecto at pgbouncer, not postgres. So I am not worried about the failing test there.

josevalim:
so let's ship it!

fishcakez:
Yeah we had a conflict, just waiting for CI to pass.

stevedomin:
This is really nice, thanks for taking care of that one @fishcakez!

PR #166, opened 2016-02-15
Include more information in error message, closes #134

### LINKED ISSUE #134: Better error message on constraint violation
It is quite frequent that I get error messages like this:

```
** (Postgrex.Error) ERROR (not_null_violation): null value in column "position_id" violates not-null constraint
```

Since I have a rather complicated schema where more than one table has a column named "position_id", I end up spending some time hunting to discover where exactly the constraint is being violated.

Would it be feasible to add the table name to the error message? Something in the lines of

```
** (Posgrex.Error) ERROR (not_null_violation): null value in column "position_id" of table "mytable" violates not-null constraint.
```

If this is not too hard to implement I would be happy to try sending a PR.

Also, thanks for the great project & effort ! ;D

josevalim:


test/error_test.exs
```diff
@@ -0,0 +1,17 @@
+defmodule ErrorTest do
+  use ExUnit.Case, async: true
+
+  alias Postgrex, as: P
+
+  @tag min_pg_version: "9.3"
+  test "connection works after failure in executing state" do
```
fishcakez: this is great addition but we should rename this test and maybe move it to query_test.exs.

PR #171, opened 2016-02-22
Use PGPORT as an integer.

lunks:
This fixes :inet_tcp.getserv/1 receiving a binary which it has no matching available.

If there is a smarter way of doing this, let me know! This is my first piece of code in Elixir. :tada:

### DIFF
--- a/lib/postgrex/utils.ex
+++ b/lib/postgrex/utils.ex
@@ -52,8 +52,14 @@ defmodule Postgrex.Utils do
     |> Keyword.put_new(:username, System.get_env("PGUSER") || System.get_env("USER"))
     |> Keyword.put_new(:password, System.get_env("PGPASSWORD"))
     |> Keyword.put_new(:hostname, System.get_env("PGHOST") || "localhost")
-    |> Keyword.put_new(:port, System.get_env("PGPORT"))
+    |> Keyword.put_new(:port, parse_port_number(System.get_env("PGPORT")))
     |> Enum.reject(fn {_k, v} -> is_nil(v) end)
+  end
+
+  defp parse_port_number(nil), do: nil
+
+  defp parse_port_number(port) when is_binary(port) do
+    String.to_integer(port)
   end
 
   @doc """
--- a/test/login_test.exs
+++ b/test/login_test.exs
@@ -136,4 +136,34 @@ defmodule LoginTest do
 
     assert_receive {:ok, %Postgrex.Result{}}
   end
+
+  test "translates port to integer" do
+    opts = []
+    previous_port = System.get_env("PGPORT")
+    try do
+      set_port_number("12345")
+      assert 12345 == P.Utils.default_opts(opts)[:port]
+    after
+      set_port_number(previous_port)
+    end
+  end
+
+  test "ignores pgport if non existent" do
+    opts = []
+    previous_port = System.get_env("PGPORT")
+    try do
+      System.delete_env("PGPORT")
+      assert nil == P.Utils.default_opts(opts)[:port]
+    after
+      set_port_number(previous_port)
+    end
+  end
+
+  defp set_port_number(nil) do
+    System.delete_env("PGPORT")
+  end
+
+  defp set_port_number(port) when is_binary(port) do
+    System.put_env("PGPORT", port)
+  end
 end


CHANGELOG.md
```diff
@@ -1,5 +1,8 @@
 # Changelog
 
+* Bug fixes
+  * Correctly handles PGPORT environment variable.
+
```
ericmj: Don't add things to the changelog because it usually gives us merge conflicts.

lib/postgrex/utils.ex
```diff
@@ -47,15 +47,23 @@ defmodule Postgrex.Utils do
...
-    |> Keyword.put_new(:port, System.get_env("PGPORT"))
+    |> Keyword.put_new(:username, environment.get_env("PGUSER") || environment.get_env("USER"))
+    |> Keyword.put_new(:password, environment.get_env("PGPASSWORD"))
+    |> Keyword.put_new(:hostname, environment.get_env("PGHOST") || "localhost")
+    |> Keyword.put_new(:port, set_port(environment.get_env("PGPORT")))
     |> Enum.reject(fn {_k, v} -> is_nil(v) end)
   end
 
+  defp set_port(port) when is_binary(port) do
+    elem(Integer.parse(port), 0)
```
ericmj: Use String.to_integer instead.

lib/postgrex/utils.ex
```diff
@@ -47,15 +47,23 @@ defmodule Postgrex.Utils do
...
+  def default_opts(opts, environment \\ System) do
     opts
-    |> Keyword.put_new(:username, System.get_env("PGUSER") || System.get_env("USER"))
-    |> Keyword.put_new(:password, System.get_env("PGPASSWORD"))
-    |> Keyword.put_new(:hostname, System.get_env("PGHOST") || "localhost")
-    |> Keyword.put_new(:port, System.get_env("PGPORT"))
+    |> Keyword.put_new(:username, environment.get_env("PGUSER") || environment.get_env("USER"))
+    |> Keyword.put_new(:password, environment.get_env("PGPASSWORD"))
+    |> Keyword.put_new(:hostname, environment.get_env("PGHOST") || "localhost")
+    |> Keyword.put_new(:port, set_port(environment.get_env("PGPORT")))
```
ericmj: You can call `System.put_env/2` in a test and call `System.delete_env/1` after instead of adding an optional argument.

test/default_opts_test.exs
```diff
@@ -0,0 +1,19 @@
...
+    def get_env(_) do
+      nil
+    end
+  end
+
+  test "translates port to integer" do
+    opts = []
+    assert 5432 == U.default_opts(opts, Environment)[:port]
+  end
+end
```
ericmj: Please move this test to `test/login_test.exs`.

lib/postgrex/utils.ex
ericmj: To make sure it always runs you can do:

```elixir
test "foo" do
  ...
after
  System.delete_env(...)
end
```

lunks: I can't change the environment variables because it will break users which
actually use these variables (like me).

ericmj: Then set it to the previous value if it was already set.

test/default_opts_test.exs
lunks: Alright, I'll move it once we're satisfied with the test itself. :+1:

lib/postgrex/utils.ex
lunks: So, I'm trying to evaluate what to do instead of this. I have a few issues:

* Using `after` I can't grab the previous set variable inside the function it seems, like this:
```elixir
  test "translates port to integer" do
    opts = []
    previous_port = System.get_env("PGPORT")
    assert 5432 == U.default_opts(opts, Environment)[:port]
  after
    set_back_pgport(previous_port)
  end
```

I'm getting a `undefined function previous_port/0` error.

If I don't use an `after` block, I'm assuming it might not get set back if the test fails based on your reasoning on why to use `after`.

One other concern is that env vars being directly set will possibly leak onto other tests as they run concurrently, won't they?

lexmag: We need to set `previous_port` before `try` like:
```elixir
  test "translates port to integer" do
    opts = []
    previous_port = System.get_env("PGPORT")
    try do
      assert 5432 == U.default_opts(opts, Environment)[:port]
    after
      set_back_pgport(previous_port)
    end
  end
```

lunks:
@lexmag @ericmj seems to be working. Anything apart from moving to `login_test`?

test/default_opts_test.exs
```diff
@@ -0,0 +1,34 @@
...
+    after
+      set_port(previous_port)
+    end
+  end
+
+  defp set_port(port) when is_binary(port) do
+    System.put_env("PGPORT", port)
+  end
+
+  defp set_port(port) when is_nil(port) do
```
lexmag: We can make this clause first and match on `nil` directly:
```elixir
defp set_port(nil) do
  System.delete_env("PGPORT")
end
defp set_port(port) when is_binary(port) do
  System.put_env("PGPORT", port)
end
```

lib/postgrex/utils.ex
```diff
@@ -52,10 +52,18 @@ defmodule Postgrex.Utils do
     |> Keyword.put_new(:username, System.get_env("PGUSER") || System.get_env("USER"))
     |> Keyword.put_new(:password, System.get_env("PGPASSWORD"))
     |> Keyword.put_new(:hostname, System.get_env("PGHOST") || "localhost")
-    |> Keyword.put_new(:port, System.get_env("PGPORT"))
+    |> Keyword.put_new(:port, set_port(System.get_env("PGPORT")))
     |> Enum.reject(fn {_k, v} -> is_nil(v) end)
   end
 
+  defp set_port(port) when is_binary(port) do
```
lexmag: The `set_port` name is a bit confusing, good in test but not here.
Maybe `port_number` or `port_value`?

lib/postgrex/utils.ex
```diff
@@ -52,10 +52,18 @@ defmodule Postgrex.Utils do
...
+    String.to_integer(port)
+  end
+
+  defp set_port(port) when is_nil(port) do
```
lexmag: This clause also can be first.

lib/postgrex/utils.ex
lunks: Agreed, changed both in test and in here.

ericmj:
Looks great, squash the commits and move the test. Then we are good to merge.

lunks:
All done. Thanks for the feedbacks, @ericmj and @lexmag

ericmj:
Thanks!

PR #188, opened 2016-05-18
Convert the port number to integer

britto:
In addition to the `PGPORT` fallback, now the given port number is also converted from string to integer.

Related: elixir-lang/ecto#544 and #171.


josevalim:

PR #198, opened 2016-06-07
Prepare fix for Ecto sandbox inconsistencies

fishcakez:
* Allows `mode: :savepoint` on prepare/execute/close that acts like `ON_ERROR_ROLLBACK` in psql. Using this means that transaction won't fail if an error is returned. If request is successful then does not add roundtrips to the database. On error adds an extra roundtrip.

* Track postgresql backend status and rollback to savepoint if failed after savepoint transaction

See https://github.com/elixir-lang/ecto/issues/1459

### DIFF
--- a/lib/postgrex.ex
+++ b/lib/postgrex.ex
@@ -103,6 +103,8 @@ defmodule Postgrex do
     `start_link/1`, see `DBConnection`
     * `:null` - The atom to use as a stand in for postgres' `NULL` in encoding
     and decoding;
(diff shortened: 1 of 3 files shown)


fishcakez:
The `mode: :savepoint` feature could be used outside of the Ecto sandbox and in a "savepoint transaction" inside the Ecto sandbox.

josevalim:
Wow, beautiful. 

PR #199, opened 2016-06-07
Add transaction mode options

fishcakez:
When beginning the outter transaction allows setting isolation level, read only and deferrable using prepared named or unnamed queries. Uses `nil` not to set a mode to use session defaults.

Prepares fix for: https://github.com/elixir-lang/ecto/issues/1460

Based on #198 to prevent conflicts and awaiting its merge.


fishcakez:
Merged manually.

PR #207, opened 2016-06-17
Fix timetz type

ericmj:


fishcakez:
I am sorry to bother you @laut but could you review this because you are a got to person for time related problems :smile:.

lau:
@fishcakez not at all, it is no bother :)

I don't see the point of the postgres `timetz` type and would recommend people to **never** use it.

Even the [postgres documentation](https://www.postgresql.org/docs/9.5/static/datatype-datetime.html) does not see the point: 😄 

> The type time with time zone is defined by the SQL standard, but the definition exhibits properties which lead to questionable usefulness.

I might even want to go as far as to suggest to issue run time warnings every time that type was used telling people to change it to a normal `time`. But I can see that if e.g. people have a legacy database with the `timetz` type that it would be useful for Postgrex to support it in some way. From that perspective this PR that assumes UTC seems reasonable.

fishcakez:
Thanks @lau. In the long run we can't map it to an elixir type so I don't think we should support this type. Note that it is currently broken and no one has filed an issue. If someone opens an issue we could revisit this. Wdyt @ericmj?

ericmj:
How come we cannot map it in the long run? How is it different than `timestamptz`?

It's not the driver's job to inform users which type they should or shouldn't use. If we can support the type in a reasonable way we should do it.

fishcakez:
@ericmj I meant there wasn't a corresponding Elixir struct, like Date, DateTime, NativeDateTime - not that we couldn't turn it into a struct. Since @lau says it's a reasonable behaviour I think we could merge this then.

lau:
It is different from timestamptz because it is a lot less useful. You could see it as hour, minute, seconds in UTC. But since there is no date, you cannot really convert it to any other timezone. E.g. if you have 13:00:00 UTC, what is that in Paris? Depends on the date. The same thing even in Reykjavík that is currently on UTC. Because Reykjavík was not always at 0 offset from UTC so without a date you can't assume that 13:00:00 UTC is 13:00:00 in Reykjavík. A datetime in UTC on the other hand can be converted to a datetime in any timezone.

So I just think the type is useless and a waste of time. A normal "time" could do the same thing, but with less risk of problems caused by postgres doing some kind of conversion.

That said I can see that it makes sense to support it. And always treating it as UTC makes the most sense. 👍

fishcakez:
So postgresql gives offset from UTC without date. Does that mean we can't
reliably convert to UTC? Or do we trust offset? Offset is in microseconds
looking at the PR?

lau:
@fishcakez If you have an offset, you can calculate the time in UTC. Just by simply addition or subtraction. E.g. 13:00+01:00 is 12:00+00:00. As long as we are just talking offsets, it is fine.

But an offset is just not a meaningful timezone in the sense of a timezone being the rules at a particular place. For instance in Paris in the last few years it has been +1 or +2 depending if DST is in place or not. But if you go back far enough the offsets were different. So "12:00 in UTC" cannot be converted to "Europe/Paris" because we don't know the date.

Issue #212, opened 2016-07-06
Changing the structure of the database on the fly causes: ERROR (feature_not_supported): cached plan must not change result type


janjiss:
Okay, so this might qualify as a bug or might not, but we have a shared database between two apps. The structure does not actually change for us, but I was not able to reproduce the bug with barebones app without changing the structure of the table, hence the example changes the type of the column.

When we change the structure of the table outside of phoenix, db connections in the pool throw an error of:  ERROR (feature_not_supported): cached plan must not change result type.

New connections are not effected

### Steps to reproduce:

mix phoenix.new hello
cd hello
mix
mix phoenix.gen.json Hello hello hello:string```

Add the resource to your api scope in web/router.ex:

    resources "/hello", HelloController, except: [:new, :edit]

mix ecto.migrate
mix ecto.schema.dump

go to localhost:4000/hello

take following code from priv/repo/structure.sql that was generated by mix ecto.schema.dump:
```SQL
CREATE TABLE hello (
    id integer NOT NULL,
    hello character varying(255),
    inserted_at timestamp without time zone NOT NULL,
    updated_at timestamp without time zone NOT NULL
);
```
and change it to:
```SQL
CREATE TABLE hello (
    id integer NOT NULL,
    hello integer,
    inserted_at timestamp without time zone NOT NULL,
    updated_at timestamp without time zone NOT NULL
);
```
Execute the query in your favourite PG client.

Refresh localhost:4000/hello and you will get: 

ERROR (feature_not_supported): cached plan must not change result type

NOTE: You might need to hit refresh several times for the query to be executed in one of the old connections or limit your pool size to 1. 

If you kill Elixir.Hello.Repo in observer or restart the app, it fixes the problem. 
The connection is not restarted after we receive this error. 

Similar issue in ActiveRecord: https://github.com/rails/rails/issues/1892

### Environment

* Elixir version (elixir -v): Elixir 1.3.1
* Phoenix version (mix deps): 
 connection 1.0.3 (Hex package) (mix)
  locked at 1.0.3 (connection) 3145f741
  ok
* fs 0.9.1 (Hex package) (rebar)
  locked at 0.9.2 (fs) ed17036c
  ok
* gettext 0.11.0 (Hex package) (mix)
  locked at 0.11.0 (gettext) 80c1dd42
  ok
* ranch 1.2.1 (Hex package) (rebar)
  locked at 1.2.1 (ranch) a6fb992c
  ok
[...]

PR #221, opened 2016-07-17
Detect stale types when executing prepared queries

fishcakez:
Match cached queries by name and referenced rather than name and statement.

Initial work towards #212. This PR means we should always error if types change.


fishcakez:
Notable changes:
* pgbouncer/`prepared: :unnamed` users will always "suffer" 2 roundtrips
* Can prepare multiple statements with the same name

fishcakez:
The failing tests on 8.4 are because of unsupported query syntax used in tests, will fix later.

josevalim:
Wow, great job @fishcakez!

PR #223, opened 2016-07-22
Remove 'use Behaviour' as deprecated

gmile:


gmile:
The time hasn't come yet.

josevalim:
I would say this is fine because using `@callback` is backwards compatible and it will work on Elixir v1.0.

PR #224, opened 2016-07-29
Only sync with backend once when doing query/4

### LINKED ISSUE #218: Raising errors causes queries not to be logged
We have a limitation today in that, if preparing or executing the query fails, the query is no logged. This makes it really hard to debug errors, specially if the error is related to parameter encoding/decoding as said here: https://groups.google.com/forum/#!topic/elixir-ecto/m8TdUtGgsX0

I would like to fix this and I can think of two options:

1. Fix Postgrex so it avoids raising as much as possible and instead if returns `{:error, Exception.t}`. This will improve known situations but still make it impossible for unknowns.

2. Fix DBConnection to also invoke the logger in case of errors.

I am more inclined to do 2 but I would like @fishcakez feedback before moving on.

### LINKED ISSUE #219: Prepare unnamed will parse query twice when using query/4
* Need to flush instead of sync in prepare when using DBConnection's prepare_execute
* Need to close on any exception from describe and encode when using DBConnection's prepare_execute

Currently prepare always syncs, which means we lose the active connection with pgbouncer between prepare and execute. We then get a new one (but could be the same) on execute and re-parse. This requires closing in DBConnection before fixing here. Likely the DBConnection side makes sense to handle at the same time as #218.

fishcakez:
* Parse query on backend once (instead of twice) when unnamed
* Savepoint once (instead of twice) when savepointing queries

Requires latest DBConnection!

Closes #219. Closes #218.

### DIFF
--- a/lib/postgrex.ex
+++ b/lib/postgrex.ex
@@ -114,7 +114,11 @@ defmodule Postgrex do
   @spec query(conn, iodata, list, Keyword.t) :: {:ok, Postgrex.Result.t} | {:error, Postgrex.Error.t}
   def query(conn, statement, params, opts \\ []) do
     query = %Query{name: "", statement: statement}
-    case DBConnection.prepare_execute(conn, query, params, defaults(opts)) do
+    opts =
+      opts
+      |> defaults()
+      |> Keyword.put(:function, :prepare_execute)
+    case DBConnection.prepare_execute(conn, query, params, opts) do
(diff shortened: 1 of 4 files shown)


fishcakez:
Oops bumped connection by mistake, fixed. Latest connection 1.0.4 requires Elixir ~> 1.2 due to bug fix with logging.

josevalim:
:shipit:

PR #215, opened 2016-07-09
add support for redshift

talentdeficit:
the query used in type bootstrapping can return null when made against aws redshift
where an empty array would be returned by postgrex

PR #235, opened 2016-09-01
Allow access to Amazon Redshift

dparnell:
Hi,

  this patch fixes access to Amazon Redshift.

Daniel

### DIFF
--- a/lib/postgrex/types.ex
+++ b/lib/postgrex/types.ex
@@ -199,6 +199,10 @@ defmodule Postgrex.Types do
     []
   end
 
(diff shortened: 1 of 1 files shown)


fishcakez:
We had exactly the same before in #215. It was probably a mistake not to merge that if this is actually enough to make it work. Unfortunately we can't test it. Of course merging would not mean we support redshift or plan to support it. I have manually merge #215.

dparnell:
I think a note that it seems to work with redshift, but that you only support official postgres would be enough

PR #236, opened 2016-09-05
Raise DBConnection.ConnectionError on connection error

### LINKED ISSUE #234: Switch tcp/ssl Postgrex.Error errors to raising DBConnection.ConnectionError
There are two types of `Postgrex.Error` at the moment. Errors from postgres and tcp/ssl errors. If a tcp/ssl error occurs we will disconnect and we can not reconnect then `DBConnection` will raise a `DBConnection.ConnectionError` if another process tries to use the connection. Therefore raising a `DBConnection.ConnectionError` on tcp/ssl error would be consistent.

The semantics of rescue encourage catching specific errors. If `Postgrex.Error` is being rescued it is likely because of an anticipated pg error, such as a constraint violation. If the user wishes to catch a tcp/ssl error they will already need to rescue `DBConnection.ConnectionError` and may want to handle this in a different part of their code. For example handle the `Postgrex.Error` as close as possible to the query it occurs but handle any `DBConnection.ConnectionError` in one place by putting all `Postgrex` calls inside a try and rescuing `DBConnection.ConnectionError`. Therefore making it easy to differentiate between when a business logic error occurs and when another service is down.

At the moment we have good behaviour for handling when a database is down in Postgrex/Ecto. However we do not have a good story for other applications calling Postgrex/Ecto. They face exactly the same problems as connection processes because their target might not be available (Postgrex raises). For example a `GenServer` in another application might make a `Postgrex.query` or `Postgrex.prepare` inside its `init/1` callback. If `Postgrex` can not connect to postgres then this error may quickly bubble up the supervision tree. In this situation it is unlikely that bubbling up the supervision tree will help because postgres is probably down. The expection being when `Postgrex` is pointing at the wrong ip/port. It should be easy to differentiate when there is a bug (pg query error) and when the database is down so its easy to use a circuit breaker or other patterns when the remote service is down.

The semantics inside a transaction are also slightly awkward with the current behaviour. If a `Postgrex.Error` occurs inside a transaction, it might be assumed that the transaction could be rolled back to a prior savepoint. However this roll back will also fail, losing the initial error.

fishcakez:
Fixes #234 .

### DIFF
--- a/lib/postgrex.ex
+++ b/lib/postgrex.ex
@@ -73,11 +73,15 @@ defmodule Postgrex do
 
   @doc """
   Runs an (extended) query and returns the result as `{:ok, %Postgrex.Result{}}`
-  or `{:error, %Postgrex.Error{}}` if there was an error. Parameters can be
-  set in the query as `$1` embedded in the query string. Parameters are given as
-  a list of elixir values. See the README for information on how Postgrex
+  or `{:error, %Postgrex.Error{}}` if there was a database error. Parameters can
+  be set in the query as `$1` embedded in the query string. Parameters are given
+  as a list of elixir values. See the README for information on how Postgrex
   encodes and decodes Elixir values by default. See `Postgrex.Result` for the
   result data.
+
+  This function may still raise an exception if there is an issue with types
+  (`ArgumentError`), connection (`DBConnection.ConnectionError`), ownership
+  (`DBConnection.OwnershipError`) or other error (`RuntimeError`).
 
   ## Options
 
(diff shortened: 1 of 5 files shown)


lib/postgrex.ex
```diff
@@ -121,12 +125,10 @@ defmodule Postgrex do
...
         {:ok, result}
-      {:error, %ArgumentError{} = err} ->
-        raise err
-      {:error, %RuntimeError{} = err} ->
-        raise err
-      {:error, _} = error ->
+      {:error, %Postgrex.Error{}} = error ->
         error
+      {:error, err} ->
+        raise err
```
fishcakez: We will need to carry forward this change to ecto

josevalim:
You are [the best](https://youtu.be/5vRlJrkxsqo?t=25). :shipit:

whatyouhide:
Sorry for the noise, just chiming in to thank for that video, can't believe I never saw it, gonna use it for everything from now on 🚀 😄

PR #239, opened 2016-09-20
Make postgrex respect env var PGDATABASE like libpq.

scouten:
I'd like to be able to configure my database name in production via environment variable and this would help. (FWIW I'm an Elixir and Ecto newbie. Happy to make changes if needed.)

### DIFF
--- a/test/login_test.exs
+++ b/test/login_test.exs
@@ -97,6 +97,14 @@ defmodule LoginTest do
     assert {:ok, %Postgrex.Result{}} = P.query(pid, "SELECT 123", [])
   end
 
+  test "env var default db name" do
+    System.put_env("PGDATABASE", "postgrex_test")
+    opts = []
+    assert {:ok, pid} = P.start_link(opts)
+    assert {:ok, %Postgrex.Result{}} = P.query(pid, "SELECT 123", [])
+    System.delete_env("PGDATABASE")
+  end
+
   test "sync connect" do
     opts = [ database: "postgres", sync_connect: true ]
     assert {:ok, pid} = P.start_link(opts)
--- a/test/login_test.exs
+++ b/test/login_test.exs
@@ -98,11 +98,26 @@ defmodule LoginTest do
   end
 
   test "env var default db name" do
-    System.put_env("PGDATABASE", "postgrex_test")
-    opts = []
-    assert {:ok, pid} = P.start_link(opts)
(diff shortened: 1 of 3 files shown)


scouten:
Shamelessly modeled after #11.

test/login_test.exs
```diff
@@ -97,6 +97,14 @@ defmodule LoginTest do
     assert {:ok, %Postgrex.Result{}} = P.query(pid, "SELECT 123", [])
   end
 
+  test "env var default db name" do
+    System.put_env("PGDATABASE", "postgrex_test")
+    opts = []
```
fishcakez: `sync_connect: true` please so `start_link` blocks to connect

test/login_test.exs
```diff
@@ -97,6 +97,14 @@ defmodule LoginTest do
...
+    assert {:ok, pid} = P.start_link(opts)
+    assert {:ok, %Postgrex.Result{}} = P.query(pid, "SELECT 123", [])
+    System.delete_env("PGDATABASE")
```
fishcakez: If the test fails we won't reset, can you copy the dance done with `try do..after.. end` and `PGPORT` later in this file?

scouten:
@fishcakez: Sure thing. It may take me a day or two to get to it, but happy to make the changes.

fishcakez:
@scouten this is a feature you want so do as you wish :smile. It took a week for this to get reviewed so we were very slow on this end, sorry.

scouten:
@fishcakez: I've updated per your suggestions. Please re-review. Thanks!

test/login_test.exs
```diff
@@ -97,6 +97,42 @@ defmodule LoginTest do
...
+
+        assert_receive {:EXIT, _, {%Postgrex.Error{postgres: %{code: :invalid_catalog_name}}, [_|_]}}
+      end
+
+      capture_log fn ->
+        opts = [ backoff_type: :stop ]
+        {:ok, pid} = P.start_link(opts)
+        assert_receive {:EXIT, ^pid, {%Postgrex.Error{postgres: %{code: :invalid_catalog_name}}, [_|_]}}
+      end
+    after
```
fishcakez: We only need to test this once, either with or without `:sync_connect`. I think the `:sync_connect example is clearer as we would only needs to assert on the `start_link` return value. No need to check for the `:EXIT`.

scouten:
@fishcakez: Thanks for the quick review. Just pushed another update.

fishcakez:
Thank you :heart:

Issue #136, opened 2016-01-03
bootstrap_query cost too much time


DavidAlphaFox:
Dear author,
I use postgresx 0.10.0 with PostgreSQL 9.4.4.
I found that the bootstrap_query cost too much time, then the server side close the connection.

I saw this in postgresSQL:

LOG:  statement: SELECT t.oid, t.typname, t.typsend, t.typreceive, t.typoutput, t.typinput,
           t.typelem, coalesce(r.rngsubtype, 0), ARRAY (
      SELECT a.atttypid
      FROM pg_attribute AS a
      WHERE a.attrelid = t.typrelid AND a.attnum > 0 AND NOT a.attisdropped
      ORDER BY a.attnum
    )
    FROM pg_type AS t
    LEFT JOIN pg_range AS r ON r.rngtypid = t.oid
    WHERE
      t.typname::text = ANY ((ARRAY['citext', 'hstore', 'jsonb', 'json'])::text[]) OR
      t.typsend::text = ANY ((ARRAY['array_send', 'range_send', 'record_send', 'cidsend', 'xidsend', 'regtypesend', 'regclasssend', 'regoperatorsend', 'regopersend', 'regproceduresend', 'regprocsend', 'oidsend', 'macaddr_send', 'cidr_send', 'inet_send', 'unknownsend', 'tidsend', 'enum_send', 'interval_send', 'timestamptz_send', 'timestamp_send', 'timetz_send', 'time_send', 'date_send', 'uuid_send', 'numeric_send', 'float8send', 'float4send', 'int8send', 'int4send', 'int2send', 'byteasend', 'varcharsend', 'textsend', 'bpcharsend', 'boolsend', 'void_send'])::text[]) OR
      t.typreceive::text = ANY ((ARRAY[])::text[]) OR
      t.typoutput::text = ANY ((ARRAY[])::text[]) OR
      t.typinput::text = ANY ((ARRAY[])::text[]);
LOG:  could not receive data from client: Operation timed out
LOG:  connection received: host=10.99.0.11 port=26380

I don't know why postgrex need this, and I want to know how to fix it.

josevalim:
We need the types information so we know how to decode types. I would try a couple things:

1. Does the issue happen on Postgrex 0.9.0?
2. Have you configured the timeouts in your database in any way?

DavidAlphaFox:
[...]

PR #241, opened 2016-09-26
Use types table lock when rebootstrapping

fishcakez:
Possible fix for #136.

Avoid doing multiple rebootstraps when multiple connections connect concurrently. We only need to do this once because the connections would get the same results. Therefore only the first connection
reboostraps, and others wait for it to finish. If it fails the next connection will try, just like the initial bootstrap.

Similarly on initial bootstrap we skip rebootstrapping on connections that were waiting for initial bootstrap.

### DIFF
--- a/lib/postgrex/protocol.ex
+++ b/lib/postgrex/protocol.ex
@@ -76,8 +76,8 @@ defmodule Postgrex.Protocol do
 
     types_key = if types?, do: {host, port, Keyword.fetch!(opts, :database), decode_bin, custom}
     status = %{opts: opts, types_key: types_key, types_ref: nil,
-               types_table: nil, extensions: extensions, prepare: prepare,
-               ssl: ssl?}
+               types_table: nil, build_types: nil, extensions: extensions,
+               prepare: prepare, ssl: ssl?}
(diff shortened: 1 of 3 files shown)


lib/postgrex/protocol.ex
```diff
@@ -535,24 +536,24 @@ defmodule Postgrex.Protocol do
...
     extension_opts = Types.prepare_extensions(extensions, parameters)
     types = Types.build_types(rows)
     Types.associate_extensions_with_types(table, extension_keys, extension_opts, types)
-    Postgrex.TypeServer.unlock(ref)
     bootstrap_sync_recv(s, status, buffer)
   end
 
-  defp bootstrap_sync_recv(s, status, buffer) do
+  defp bootstrap_sync_recv(s, %{types_ref: ref} = status, buffer) do
+    Postgrex.TypeServer.unlock(ref)
```
fishcakez: This might get called multiple times so we should move it back to bootstrap_types/4.

fishcakez:
Merged manually https://github.com/elixir-ecto/postgrex/commit/4967683e696b3ccb5c40731927b972e114a9a2dc.

PR #246, opened 2016-10-01
Implement bit string encoding/decoding

dantswain:
Maps to/from Elixir bit strings as per #245

This certainly can be cleaned up a bit, but I wanted to get some code up for discussion.

I want to take another look at this after I've done some clean up, but I'm still not convinced that bit strings are the cleanest implementation here.  In fact, for the decode implementation I end up building a list and then converting that back into a bit string.


dantswain:
Also still need to make sure this covers `varbit`.

dantswain:
Tests failed on Elixir 1.0.5 because `Integer.digits/2` is unavailable there.  Should be easy enough to work around if that's still needed after cleanup.

dantswain:
Wow, I'm not sure how I managed to complicate this so much on the first go.  The last commit is much simpler.

Would it be beyond the scope of Postgrex to add a BitString module to help work with these values?  E.g., `"10" = BitString.to_s(<< 1::1, 0::1 >>)`, `true = BitString.bit_set?(<< 1::1, 0::1 >>, 0)`, etc.

lib/postgrex/extensions/bit_string.ex
```diff
@@ -0,0 +1,24 @@
+defmodule Postgrex.Extensions.BitString do
+  use Postgrex.BinaryExtension,
+    [receive: "bit_recv", send: "bit_send", type: "bit",
+     receive: "varbit_recv", send: "varbit_send", type: "varbit"]
```
fishcakez: We just use `send` entries because that describes the wire encoding. Using `type` can lead to issues with custom types. Postgrex can't handle the situation where a `send` and `receive` use non-matching encoding so we skip `receive`.

test/bit_string_test.exs
```diff
@@ -0,0 +1,77 @@
...
+         0::1,0::1,0::1,0::1,0::1,0::1,0::1,0::1,
+         1::1,0::1,1::1>>])
+    assert [["1000000110000000101"]] ==
+      query("SELECT $1::bit(19)::text",
+        [<<1::1,0::1,0::1,0::1,0::1,0::1,0::1,1::1,
+         1::1,0::1,0::1,0::1,0::1,0::1,0::1,0::1,
+         1::1,0::1,1::1>>])
+  end
+
+  test "persist bit string", context do
```
fishcakez: We are ok to skip this test, we don't do it for types unless there are issues with persisting.

test/test_helper.exs
```diff
@@ -55,6 +55,8 @@ DROP TYPE IF EXISTS missing_comp;
 CREATE TABLE altering (a int2);
 
 CREATE TABLE calendar (a timestamp without time zone, b timestamp with time zone);
+
+CREATE TABLE bitstring_test (a bit(5), b bit(10), c varbit)
```
fishcakez: :scissors:

test/bit_string_test.exs
```diff
@@ -0,0 +1,77 @@
+defmodule BitStringTest do
+  use ExUnit.Case, async: true
```
fishcakez: Please move this to `test/query_test.exs` where most other tests are. calendar types have their own test file but they are tricky and we need a way to disable those for old versions of elixr.

test/bit_string_test.exs
```diff
@@ -0,0 +1,77 @@
...
+  import Postgrex.TestHelper
+  alias Postgrex, as: P
+
+  setup do
+    opts = [database: "postgrex_test", backoff_type: :stop,
+            extensions: [{Postgrex.Extensions.BitString, []}]]
```
fishcakez: We want to avoid requiring the extension explicitly. Please add the extension module to the attribute in `Postgrex.Utils`.

fishcakez:
> Would it be beyond the scope of Postgrex to add a BitString module to help work with these values?

I think it would be given we are providing a direct mapping to a native type.

lib/postgrex/utils.ex
```diff
@@ -26,7 +26,8 @@ defmodule Postgrex.Utils do
     Postgrex.Extensions.Timestamp,
     Postgrex.Extensions.Interval,
     Postgrex.Extensions.UUID,
-    Postgrex.Extensions.Void]
+    Postgrex.Extensions.Void,
+    Postgrex.Extensions.BitString]
```
fishcakez: alphabetical order please

fishcakez:
Thank you :heart:. I made a few changes in https://github.com/elixir-ecto/postgrex/commit/d5c627d486c8a20728e56443424dc85b1a260ce5 to use iodata to reduce copying and added support for the `:decode_binary` option.

PR #252, opened 2016-10-15
Update ex_doc to 0.14.3 and minor docs update for start_link

wojtekmach:
This PR does two things:

1. Updates ex_doc to 0.14.3, so that we can cross link to DBConnection docs
2. Updates `Postgrex.start_link` to mention `after_connect` as dicussed on: https://github.com/elixir-ecto/db_connection/issues/63#issuecomment-253910368

Rendered:

rendered

### DIFF
--- a/lib/postgrex.ex
+++ b/lib/postgrex.ex
@@ -63,7 +63,8 @@ defmodule Postgrex do
     and decoding (default: `nil`);
 
   `Postgrex` uses the `DBConnection` framework and supports all `DBConnection`
-  options. See `DBConnection` for more information.
+  options like `:idle`, `:after_connect` etc.
+  See `DBConnection.start_link/2` for more information.
   """
   @spec start_link(Keyword.t) :: {:ok, pid} | {:error, Postgrex.Error.t | term}
   def start_link(opts) do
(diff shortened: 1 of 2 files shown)


josevalim:

lib/postgrex.ex
```diff
@@ -63,7 +63,8 @@ defmodule Postgrex do
     and decoding (default: `nil`);
 
   `Postgrex` uses the `DBConnection` framework and supports all `DBConnection`
-  options. See `DBConnection` for more information.
+  options like `:idle`, `:after_connect` etc.
+  See `DBConnection.start_link/2` for more information.
   """
```
wojtekmach: btw, as discussed elixir-ecto/db_connection#63, I think it would be good to show case how to actually use after_connect in practice, but not sure how best to add it here. perhaps as an example?

```markdown

### Examples

    iex> Postgrex.start_link(database: "postgres")
    {:ok, #PID<0.69.0>}

Run a query when connection is established:

    defmodule MyMod do
      def set_timezone(conn) do
        {:ok, _} = Postgrex.query(conn, "SET TIME ZONE 'UTC';", [])
      end
    end

    iex> {:ok, pid} = Postgrex.start_link(database: "postgres", after_connect: &MyMod.set_timezone/1)
    {:ok, #PID<0.69.0>}
```

josevalim:
An example sounds great.

-- 

PR #253, opened 2016-10-15
Add example for using after_connect hook

wojtekmach:


lib/postgrex.ex
```diff
@@ -65,6 +65,23 @@ defmodule Postgrex do
...
+      {:ok, #PID<0.69.0>}
+
+  Run a query after connection has been established:
+
+      defmodule MyMod do
+        def set_timezone(pid) do
+          {:ok, _} = Postgrex.query(pid, "SET TIME ZONE 'UTC';", [])
+        end
+      end
+
```
fishcakez: Could we simplify this to `&Postgrex.query!(&1, "SET TIME ZONE 'UTC';", [])` so that the example is a simple as psosible. Note also that it is not `pid` in an after_connect.

wojtekmach: :+1: Fixed in 24f71ffabda60d234034654b32bf7e88bef70654

lib/postgrex.ex
```diff
@@ -68,22 +68,16 @@ defmodule Postgrex do
 
   ## Examples
 
-      iex> {:ok, pid} = Postgrex.start_link(database: "postgres")
+      iex> {:ok, conn} = Postgrex.start_link(database: "postgres")
```
fishcakez: Please leave this as a pid because it is always a pid.

lib/postgrex.ex
```diff
@@ -364,7 +358,7 @@ defmodule Postgrex do
 
   ## Example
 
-      {:ok, res} = Postgrex.transaction(pid, fn(conn) ->
+      {:ok, res} = Postgrex.transaction(conn, fn(conn) ->
```
fishcakez: Please leave this as `pid` because if it is a pid then the `conn` in `fn(conn)` won't be `pid` and the shadowing of `conn` is confusing. This occurs elsewhere too.

wojtekmach:
re: pid -> conn. I noticed that in some places in the docs the 1st argument was called `conn`, not `pid` (e.g.: [here](https://github.com/elixir-ecto/postgrex/blob/dfaa93c6b2f0ff513b698e13d8b898b78280f315/lib/postgrex.ex#L107)), so that's why I thought unifying it might be a good idea.

But yeah, in some of these examples (e.g. transaction) shadowing is a problem, so I'll just revert the latest commit and update the `after_connect` per your initial comment.

Let me know if you'd like me to change pid->conn in places where shadowing is not a problem

fishcakez:
Thank you :heart:

PR #256, opened 2016-11-06
Use prepare_stream and stream from DBConnection

fishcakez:
* No longer require stream to be prepared before use
* Always bind params in a stream's start fun
* Always execute portal in a stream's first next fun
* Report `:num_rows` for each result in a stream
* Identify streams in logs with cursor/copy struct
* Drop support for `:copy_data` parameter

### DIFF
--- a/lib/postgrex.ex
+++ b/lib/postgrex.ex
@@ -401,7 +401,7 @@ defmodule Postgrex do
   end
 
   @doc """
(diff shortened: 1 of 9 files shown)


josevalim:
:+1:

fishcakez:
Updated for improved DBConnection API.

PR #257, opened 2016-11-11
Fail all waiting connections on bootstrap failure

fishcakez:
Adds more timeout control, logging on handshake timeout and all waiting connections fail if the bootstrapper fails. This means all connections backoff for a bit so we don't hog the backend. Hopefully it will succeed later. I don't think this will solve the problem but it should provide a better ops experience.


fishcakez:
@safwank is it possible you could try to reproduce #136 and if you can, could try to use this branch to see if you can avoid the situation? You could see a benefit from trying different `handshake_timeout` values but we should prevent the spiral of failure.

PR #280, opened 2017-01-01
allow user specified variant protocols to support redshift and cockroachdb

talentdeficit:


talentdeficit:
this is non-functional and just for discussion, see #279

fishcakez:
This would require documenting functions that are currently `@doc false` and I am not sure if we are ready to get undressed in front of the world like that yet :S. Maybe for 1.0 but not yet.

It would prevent us from supporting super extensions for the text format because we could not introduce delimiters (and any other requirements) to the bootstrap query/type info struct. However we might already have blocked that feature by requiring the leading length when decoding, so this could be ok.

talentdeficit:
you don't need to document `Types.bootstrap_query` you just need to declare an interface/type signature for the bootstrap query

for introducing delimiters, couldn't you just kick out to the variant module for that also?

talentdeficit:
closing in favour of disabling re-bootstrap

PR #283, opened 2017-01-07
Test against elixir 1.4.0

ddrmanxbxfr:
Elixir 1.4 is now released in stable :tada: .

We can test against it instead of the release candidate.

Thanks!


josevalim:

PR #284, opened 2017-01-07
Base implementation of Tsvector & Lexeme support

ddrmanxbxfr:
Hi,

This PR adds the support for Tsvector extensions.

It's usually used with full text search features that are built in inside postgres.

I wanted to be able to use manually written tsvector and tsquery directory from ecto. However, no extension was available for it, so I implemented one.

If this seems good, I will look to implement tsquery after this PR merges as both are usually used together...

Thanks!

Links to the official postgres doc : https://www.postgresql.org/docs/8.3/static/datatype-textsearch.html

The query tests are the ones from postgresql directly here : https://github.com/postgres/postgres/blob/89fcea1ace40bc025beea2758a80bcd56a319a6f/src/test/regress/sql/tstypes.sql

### DIFF
--- a/README.md
+++ b/README.md
@@ -49,6 +49,7 @@ iex> Postgrex.query!(pid, "INSERT INTO comments (user_id, text) VALUES (10, 'hey
     enum            "ok" ***
     bit             << 1::1, 0::1 >>
     varbit          << 1::1, 0::1 >>
+    tsvector        [%Postgrex.Lexeme{positions: [{1, :A}], word: "a"}]
 
 \* [Decimal](http://github.com/ericmj/decimal)
 
--- a/lib/postgrex/builtins.ex
+++ b/lib/postgrex/builtins.ex
@@ -257,6 +257,23 @@ defmodule Postgrex.Circle do
 
   defstruct [
     center: nil,
-    radius: nil 
+    radius: nil
+  ]
+end
+
+defmodule Postgrex.Lexeme do
+  @moduledoc """
+  Struct for Postgres Lexeme (A Tsvector type is composed of multiple lexemes)
+
+  ## Fields
+    * `word`
+    * `positions`
+  """
+
+  @type t :: %__MODULE__{word: String.t, positions: [{pos_integer, :A | :B | :C | nil}]}
+
+  defstruct [
+    word: nil,
+    positions: nil
   ]
 end
--- a/lib/postgrex/extensions/tsvector.ex
+++ b/lib/postgrex/extensions/tsvector.ex
@@ -1,0 +1,62 @@
+defmodule Postgrex.Extensions.TSVector do
+  @moduledoc false
+
+  import Postgrex.BinaryUtils, warn: false
+  use Bitwise
+  use Postgrex.BinaryExtension, [send: "tsvectorsend"]
+
+  def encode(_) do
+    quote location: :keep do
+      values when is_list(values) ->
+        encoded_tsvectors = unquote(__MODULE__).encode_tsvector(values)
+        <<byte_size(encoded_tsvectors)::int32, encoded_tsvectors::binary>>
+      other ->
+        raise ArgumentError, Postgrex.Utils.encode_msg(other, "a list of tsvectors")
+    end
+  end
+
+  def decode(_) do
+    quote do
+      <<len::int32, value::binary-size(len)>> ->
+        <<nb_lexemes::int32, words::binary>> = value
+        unquote(__MODULE__).decode_tsvector_values(words)
+    end
+  end
+
+  ## Helpers
+
+  def encode_tsvector(values) do
+    <<length(values)::int32, encode_lexemes(values)::binary>>
+  end
+
+  defp encode_lexemes(values) do
+    Enum.map(values, fn(x) -> encode_positions(x) end) |> IO.iodata_to_binary
+  end
+
+  defp encode_positions(%Postgrex.Lexeme{word: word, positions: positions}) do
+    positions = Enum.map(positions, fn({position, weight}) -> <<encode_weight_binary(weight)::2, position::14>> end)
+    [[word, 0], <<length(positions)::16>>, positions]
+  end
+
+  def decode_tsvector_values("") do
+    []
+  end
+
+  def decode_tsvector_values(words) do
+    [word, <<positions_count::16, rest::binary>>] = :binary.split(words, [<<0>>])
+    positions_bytes = positions_count * 2
+    <<positions::binary-size(positions_bytes), remaining_data::binary>> = rest
+    positions = for <<weight::2, position::14 <- positions>>, do: {position, decode_weight(weight)}
+    [%Postgrex.Lexeme{word: word, positions: positions} | decode_tsvector_values(remaining_data)]
+  end
+
+  defp encode_weight_binary(:A) do 3 end
+  defp encode_weight_binary(:B) do 2 end
+  defp encode_weight_binary(:C) do 1 end
+  defp encode_weight_binary(nil) do 0 end
+
+  defp decode_weight(0), do: nil
+  defp decode_weight(1), do: :C
+  defp decode_weight(2), do: :B
+  defp decode_weight(3), do: :A
+end
--- a/test/tsvector_test.exs
+++ b/test/tsvector_test.exs
@@ -1,0 +1,53 @@
+defmodule TsvectorTest do
+  use ExUnit.Case, async: true
+  import Postgrex.TestHelper
+  alias Postgrex, as: P
+
+  setup do
+    opts = [database: "postgrex_test", backoff_type: :stop]
+    {:ok, pid} = P.start_link(opts)
+    {:ok, [pid: pid]}
+  end
+
+  test "encode basic tsvector", context do
+    assert [["'1'"]] = query("SELECT $1::tsvector::text", [[%Postgrex.Lexeme{positions: [], word: "1"}]])
+    assert [[[%Postgrex.Lexeme{positions: [], word: "1"}]]] = query("SELECT $1::tsvector", [[%Postgrex.Lexeme{positions: [], word: "1"}]])
+    assert [["'1' 'hello'"]] = query("SELECT $1::tsvector::text", [[%Postgrex.Lexeme{positions: [], word: "1"}, %Postgrex.Lexeme{positions: [], word: "hello"}]])
+    assert [[[%Postgrex.Lexeme{positions: [], word: "1"}, %Postgrex.Lexeme{positions: [], word: "hello"}]]] = query("SELECT $1::tsvector", [[%Postgrex.Lexeme{positions: [], word: "1"}, %Postgrex.Lexeme{positions: [], word: "hello"}]])
+  end
+
+  test "encode tsvector with positions", context do
+    assert [[[%Postgrex.Lexeme{positions: [{1, nil}], word: "1"}]]] = query("SELECT $1::tsvector", [[%Postgrex.Lexeme{positions: [{1, nil}], word: "1"}]])
+  end
+
+  test "encode tsvector with multiple positions", context do
+    assert [[[%Postgrex.Lexeme{positions: [{1, nil}, {2, nil}], word: "1"}]]] = query("SELECT $1::tsvector", [[%Postgrex.Lexeme{positions: [{1, nil}, {2, nil}], word: "1"}]])
+  end
+
+  test "encode tsvector with position and weight", context do
+    assert [[[%Postgrex.Lexeme{positions: [{1, :A}], word: "car"}]]] = query("SELECT $1::tsvector", [[%Postgrex.Lexeme{positions: [{1, :A}], word: "car"}]])
+    assert [[[%Postgrex.Lexeme{positions: [{1, :B}], word: "car"}]]] = query("SELECT $1::tsvector", [[%Postgrex.Lexeme{positions: [{1, :B}], word: "car"}]])
+    assert [[[%Postgrex.Lexeme{positions: [{1, :C}], word: "car"}]]] = query("SELECT $1::tsvector", [[%Postgrex.Lexeme{positions: [{1, :C}], word: "car"}]])
+  end
+
+  test "encode tsvector with multiple positions and weights", context do
+        assert [[[%Postgrex.Lexeme{positions: [{1, :A}, {2, nil}, {3, :B}], word: "car"}]]] = query("SELECT $1::tsvector", [[%Postgrex.Lexeme{positions: [{1, :A}, {2, nil}, {3, :B}], word: "car"}]])
+  end
+
+  test "decode basic tsvectors", context do
+    assert [[[%Postgrex.Lexeme{positions: [], word: "1"}]]] = query("SELECT '1'::tsvector", [])
+    assert [[[%Postgrex.Lexeme{positions: [], word: "1"}]]] = query("SELECT '1 '::tsvector", [])
+    assert [[[%Postgrex.Lexeme{positions: [], word: "1"}]]] = query("SELECT ' 1'::tsvector", [])
+    assert [[[%Postgrex.Lexeme{positions: [], word: "1"}]]] = query("SELECT ' 1 '::tsvector", [])
+  end
+
+  test "decode tsvectors with multiple elements", context do
+    assert [[[%Postgrex.Lexeme{positions: [], word: "1"}, %Postgrex.Lexeme{positions: [], word: "2"}]]] = query("SELECT '1 2'::tsvector", [])
+    assert [[[%Postgrex.Lexeme{positions: [], word: "1 2"}]]] = query("SELECT '''1 2'''::tsvector", [])
+  end
+
+  test "decode tsvectors with multiple positions and elements", context do
+    assert [[[%Postgrex.Lexeme{positions: [{8, nil}], word: "a"}, %Postgrex.Lexeme{positions: [{1, nil}, {2, :C}, {3, :B}, {4, :A}, {5, nil}], word: "w"}]]] = query("SELECT '''w'':4A,3B,2C,1D,5 a:8'::tsvector", [])
+    assert [[[%Postgrex.Lexeme{positions: [{3, :A}], word: "a"}, %Postgrex.Lexeme{positions: [{2, :A}], word: "b"}]]] = query("SELECT 'a:3A b:2a'::tsvector", [])
+  end
+end
--- a/lib/postgrex/utils.ex
+++ b/lib/postgrex/utils.ex
@@ -36,6 +36,7 @@ defmodule Postgrex.Utils do
     Postgrex.Extensions.Timestamp,
     Postgrex.Extensions.TimestampTZ,
     Postgrex.Extensions.TimeTZ,
+    Postgrex.Extensions.TSVector,
     Postgrex.Extensions.UUID,
     Postgrex.Extensions.VoidBinary,
     Postgrex.Extensions.VoidText]


ddrmanxbxfr:
The CI failure seems to be related in a intermitent bug, we can retry it once #283 merges if we want to be sure.

lib/postgrex/extensions/tsvector.ex
```diff
@@ -0,0 +1,45 @@
+defmodule Postgrex.Extensions.Tsvector do
+  @moduledoc false
+  @behaviour Postgrex.Extension
+
+  def init(_), do: {}
+
+  def matching(_), do: [send: "tsvectorsend"]
```
fishcakez: If the format is `:text` should match on `:output` because that specifies the text format, `:send` is for `:binary` format. However we want to use `:binary` format because that composes with "super" extensions (array etc) and is more efficient. It would be very unlikely that we would support a `:text` format extension as a built in at this time.

fishcakez:
> The CI failure seems to be related in a intermitent bug, we can retry it once #283 merges if we want to be sure.

It looks like a race condition in the tests that was introduced with the new type handling. Don't worry about this :D.

lib/postgrex/extensions/tsvector.ex
ddrmanxbxfr: Alright, I'll have a look at the C code of postgresql and implement the binary parsing of TSVector.

I'll poke you once ready as it may take some reading. afaik, the decoding part requires a small algorithm.

ddrmanxbxfr:
I reworked the code during the weekend and now support the Encoding and Decoding with binary 🎉 
I also added tests for the encode and decode part.

@fishcakez would you mind re-reviewing this PR with binary support now ?

Thanks!

lib/postgrex/extensions/tsvector.ex
```diff
@@ -0,0 +1,127 @@
...
+
+  ## Helpers
+
+  def encode_tsvector(values) do
+    << length(values) :: int32, encode_lexemes(values) :: binary >>
+  end
+
+  defp encode_lexemes(values) do
+    Enum.map(values, fn(x) ->
+        case x do
```
josevalim: There is no need for case here. You can pattern on the lexeme struct in the function head.

lib/postgrex/extensions/tsvector.ex
```diff
@@ -0,0 +1,127 @@
...
+          %Postgrex.Lexeme{word: word, positions: positions} ->
+            %{word: "#{word}\0", positions: positions}
+          end
+        end)
+      |> Enum.map(fn(x) -> encode_positions(x[:word], x[:positions]) end)
```
josevalim: There is no need for this map. You can call encode_positions with the word and positions directly in the map above.

lib/postgrex/extensions/tsvector.ex
```diff
@@ -0,0 +1,127 @@
...
+      |> Enum.reduce("", fn(x, acc) -> acc <> x end)
+  end
+
+  defp encode_positions(bytes, position) do
+      positions = Enum.map(position, fn(x) -> << encode_weight(x)::2, encode_position(x)::14 >> end)
+                  |> Enum.reduce("", fn(x, acc) -> acc <> x end)
```
josevalim: Instead of `Enum.reduce`, you should call `IO.iodata_to_binary`, it is going to be much faster. In fact, it is best to avoid `<>` since it copies the binaries and instead build lists of binaries and lists (called io lists) and then call IO.iodata_to_binary at the end. So for example, `encode_positions` could look like this:

```elixir
positions = Enum.map(position, fn(x) -> << encode_weight(x)::2, encode_position(x)::14 >> end)
[bytes, <<length(position) :: 16>>, positions]
```

And then encode_lexemes would look like this:

```elixir
values
|> Enum.map(fn %Postgrex.Lexeme{word: word, positions: positions} ->
  encode_positions(word, positions)
end)
|> IO.iodata_to_binary
```

lib/postgrex/extensions/tsvector.ex
ddrmanxbxfr: Indeed! Didn't knew I could pattern match at the function level. Really cool feature.

lib/postgrex/extensions/tsvector.ex
ddrmanxbxfr: Yeup, removed ! :)

lib/postgrex/extensions/tsvector.ex
ddrmanxbxfr: This is a good improvement.

I was able to remove all the Enum.reduce in this file as they were all doing the same operation 👍 .

ddrmanxbxfr:
Implemented the improvements proposed by @josevalim.

Removing all the list copy were indeed a good improvement 👍 . 

I was able to remove all the Enum.reduce in favor of IO.iodata_to_binary and simplify two or three functions.

Thanks to all.

josevalim:
@ddrmanxbxfr Actually, I just realized that the encode functions allow you to return an iodata, so you don't even need to call iodata_to_binary anywhere. Just remove it and you should still be good.

lib/postgrex/extensions/tsvector.ex
```diff
@@ -0,0 +1,115 @@
...
+      <<len :: int32, value :: binary-size(len)>> ->
+        <<nb_lexemes :: int32, words :: binary>> = value
+        unquote(__MODULE__).decode_tsvector_values(:binary.bin_to_list(words))
+    end
+  end
+
+  ## Helpers
+
+  def encode_tsvector(values) do
+    << length(values) :: int32, encode_lexemes(values) :: binary >>
```
josevalim: Please no spaces after `<<` and before `>>`. Please check other occurrences in the file.

lib/postgrex/extensions/tsvector.ex
```diff
@@ -0,0 +1,115 @@
...
+  end
+
+  def encode_lexemes(values) do
+    Enum.map(values, fn(%Postgrex.Lexeme{word: word, positions: positions}) ->
+      encode_positions("#{word}\0", positions)
```
josevalim: Following the iodata idea, you want `"#{word}\0"` to be `[word, 0]`.

josevalim:
@ddrmanxbxfr I have added some more comments but I also have a question. Instead of encoding positions as "1A", "2B", wouldn't it be better to encode them as `{position :: integer, weight :: integer}`? This way both encoding and decoding become much simpler and I believe it is a more interesting format for users to work with as well. Is there any reason why we use "1A"?

ddrmanxbxfr:
It's the standard format that is mentioned for text and in the
documentation of postgres.

I'm fine with both format so I'll switch to the

%{position::integer, weight::integer} format

The important is that we still support weights with letters either charlist
or string. As the weight with numbers are only used internally in postgres.


On 10 Jan 2017 4:27 a.m., "José Valim" <notifications@github.com> wrote:

@ddrmanxbxfr <https://github.com/ddrmanxbxfr> I have added some more
comments but I also have a question. Instead of encoding positions as "1A",
"2B", wouldn't it be better to encode them as {position :: integer, weight
:: integer}? This way both encoding and decoding become much simpler and I
believe it is a more interesting format for users to work with as well. Is
there any reason why we use "1A"?

—
You are receiving this because you were mentioned.
Reply to this email directly, view it on GitHub
<https://github.com/elixir-ecto/postgrex/pull/284#issuecomment-271526722>,
or mute the thread

josevalim:
> %{position::integer, weight::integer} format

Just to be clear, I proposed a tuple format, such as `{position, integer}` and not `%{position: integer, weight: integer}`. I see no reason for it to be a map if it contains only two elements.

> The important is that we still support weights with letters either charlist or string.

Are they always a single letter?

josevalim:
Btw, still regarding this:

> The important is that we still support weights with letters either charlist or string.

I would wait for @ericmj's feedback when he comes back from holidays. If the goal of Postgrex is to provide a low-level wrapper around PostgreSQL APIs, then I would support only integer weights.

ddrmanxbxfr:
Yes they are always a single letter. Do note that a word can have multiple
positions / weight combinaison.

If we only support int then people will need to read postgres source code
to know the mapping and convert the query they need to do.

josevalim:
I see. So it may be worth keeping it as a char list as today. The last element of the char list will always be an integer. We can optimize the code based on those assumptions which seems to be always true.

ddrmanxbxfr:
Have a look at encode_weight functions. I'm pattern matching with the ASCII
code of the letters we need to support :)

lib/postgrex/extensions/tsvector.ex
```diff
@@ -0,0 +1,115 @@
...
+    end)
+    |> IO.iodata_to_binary
+  end
+
+  defp encode_positions(bytes, position) do
+      positions = Enum.map(position, fn(x) ->
+        chars = to_char_list(x)
+        << encode_weight(chars)::2, encode_position(chars)::14 >>
+      end)
+                  |> IO.iodata_to_binary
```
ericmj: Weird indentation.

lib/postgrex/extensions/tsvector.ex
```diff
@@ -0,0 +1,115 @@
...
+
+  defp encode_weight(data) do
+    Enum.find(data, -1, fn(x) -> (x >= 65 && x <= 67) || (x >= 97 && x <= 99) end)
+    |> encode_weight_binary
+  end
+
+  defp encode_position(data) do
+   Enum.filter(data, fn(x) -> x >= 48 && x <= 57 end)
+   |> :string.to_integer
+   |> elem(0)
```
ericmj: `List.to_integer`.

lib/postgrex/extensions/tsvector.ex
ddrmanxbxfr: Alright, removed all of them

lib/postgrex/extensions/tsvector.ex
ddrmanxbxfr: Yeup that works too 👍

lib/postgrex/extensions/tsvector.ex
ddrmanxbxfr: Inlined it, thanks!

lib/postgrex/extensions/tsvector.ex
ddrmanxbxfr: I've converted the whole encode part to {position, weight} with charlist support.

It allows us to trim all the Enum.filter part in this code.

ddrmanxbxfr:
@josevalim @fishcakez 

I have reworked the lexeme formats to make it more lean and try to optimize it a bit.

Here are the changes :

1. Remove all the string parsing with Enum.filter to work with charlist and pattern matching instead.
2. Remove all the iodata_to_binary that I could
3. Fix the linting on binary formats so we don't have no space in them
4. Rework the positions and weight format to look like so...
```
%Postgrex.Lexeme{positions: [{1, 'A'}, {2, nil}, {3, 'B'}], word: "car"}
```

It does look more lean, thanks to all the contributions so we can make it real good.

lib/postgrex/extensions/tsvector.ex
```diff
@@ -0,0 +1,96 @@
...
+  defp encode_weight_binary('A') do 3 end
+  defp encode_weight_binary('B') do 2 end
+  defp encode_weight_binary('C') do 1 end
+  defp encode_weight_binary('a') do 3 end
+  defp encode_weight_binary('b') do 2 end
+  defp encode_weight_binary('c') do 1 end
+  defp encode_weight_binary(3) do 3 end
+  defp encode_weight_binary(2) do 2 end
+  defp encode_weight_binary(1) do 1 end
+  defp encode_weight_binary(_) do 0 end
```
fishcakez: Should this be matching `nil`? If so then we want `@null` (it'll be registered automatically) as we use that for all `nil`.

lib/postgrex/extensions/tsvector.ex
```diff
@@ -0,0 +1,96 @@
...
+
+  defp decode_weight(1) do 'C' end
+  defp decode_weight(2) do 'B' end
+  defp decode_weight(3) do 'A' end
+  defp decode_weight(_) do nil end
```
fishcakez: We should not have catch all that ignores data. What values could this be?

test/tsvector_test.exs
```diff
@@ -0,0 +1,61 @@
...
+  require IEx
+
+  setup do
+    opts = [database: "postgrex_test", backoff_type: :stop]
+    {:ok, pid} = P.start_link(opts)
+    {:ok, [pid: pid]}
+  end
+
+  test "encode basic tsvector", context do
+    assert [[[%Postgrex.Lexeme{positions: [], word: "1"}]]] = query("SELECT $1::tsvector", [[%Postgrex.Lexeme{positions: [], word: "1"}]])
```
fishcakez: Please cast to text in the query so that the result is a string. Otherwise our encoding test relies on the decoding being correct.

lib/postgrex/extensions/tsvector.ex
```diff
@@ -0,0 +1,96 @@
+defmodule Postgrex.Extensions.Tsvector do
```
fishcakez: Please rename TSVector as it is Text Search Vector

lib/postgrex/extensions/tsvector.ex
```diff
@@ -0,0 +1,96 @@
...
+
+  defp extract_positions(bytes) do
+    Enum.chunk(bytes, 2)
+    |> Enum.map(fn(x) ->
+        <<raw_word_meta::uint16>> = :erlang.list_to_bitstring(x)
+        {raw_word_meta &&& 0x3fff, decode_weight(raw_word_meta >>> 14)}
+       end)
+  end
+
+  defp encode_weight_binary('A') do 3 end
```
fishcakez: Why charlist? Perhaps `:A` or `"A"`?

README.md
```diff
@@ -49,6 +49,7 @@ iex> Postgrex.query!(pid, "INSERT INTO comments (user_id, text) VALUES (10, 'hey
     enum            "ok" ***
     bit             << 1::1, 0::1 >>
     varbit          << 1::1, 0::1 >>
+    tsvector        [%Postgrex.Lexeme{positions: ["8"], word: "a"}]
```
fishcakez: Please update to new format :)

lib/postgrex/builtins.ex
```diff
@@ -257,6 +257,23 @@ defmodule Postgrex.Circle do
...
+defmodule Postgrex.Lexeme do
+  @moduledoc """
+  Struct for Postgres Lexeme (A Tsvector type is composed of multiple lexemes)
+
+  ## Fields
+    * `word`
+    * `positions`
+  """
+
+  @type t :: %__MODULE__{word: String.t, positions: [%{}]}
```
fishcakez: `:positions` is a list of 2-tuples now?

test/tsvector_test.exs
```diff
@@ -0,0 +1,61 @@
+defmodule TsvectorTest do
+  use ExUnit.Case, async: true
+  import Postgrex.TestHelper
+  alias Postgrex, as: P
+  require IEx
```
fishcakez: :scissors:

lib/postgrex/builtins.ex
ddrmanxbxfr: Yeup. Which seem to make more sense than using a string that we parse.

{position,weight}

test/tsvector_test.exs
ddrmanxbxfr: Done

lib/postgrex/extensions/tsvector.ex
ddrmanxbxfr: Alright

lib/postgrex/extensions/tsvector.ex
ddrmanxbxfr: Because it's always only 1 character ? 

I'm far from an elixir expert, would using "A" or :A make more sense performance wise ?

lib/postgrex/extensions/tsvector.ex
ddrmanxbxfr: I've read a bit more on the postgresql source code and I found out we can only have 1 2 3 and 0 as weight for decoding and it should be fine.

So i'll remove the catch all 👍

test/tsvector_test.exs
ddrmanxbxfr: Added 2 tests using cast to text in the query to make sure the encoding is correct.

Thanks!

lib/postgrex/extensions/tsvector.ex
ddrmanxbxfr: Yeup thanks !

lib/postgrex/extensions/tsvector.ex
fishcakez: Well :A might be faster because comparison of atoms is as quick as it gets but it string or atoms felt more appropriate from language perspective. Atom because there are only 3 named weights. String because its used more than char list. Lets see what another maintainer thinks.

README.md
ddrmanxbxfr: Done !

lib/postgrex/builtins.ex
knewter: I think the point was that this type says it's a list of maps still

fishcakez: Yeah. `[%{}]` means a list of __empty__ maps if you didn't know @ddrmanxbxfr.

ddrmanxbxfr: TIL [%{}] a list of empty maps ! Thanks @fishcakez @knewter 

I've changed it for Tuple.t

lib/postgrex/extensions/tsvector.ex
ddrmanxbxfr: @josevalim Can we have your input on this ?

Thanks <3

lib/postgrex/builtins.ex
fishcakez: It should be `[{position, weight}, ...]` (I think must be non-empty list?), where `position` and `weight` describe the types, e.g. `[{pos_integer, :A | :B | :C | nil}, ...]`

lib/postgrex/extensions/tsvector.ex
```diff
@@ -0,0 +1,96 @@
...
+  use Postgrex.BinaryExtension, [send: "tsvectorsend"]
+
+  def encode(_) do
+    quote location: :keep do
+      values when is_list(values) ->
+        encoded_tsvectors = unquote(__MODULE__).encode_tsvector(values)
+        <<byte_size(encoded_tsvectors)::int32, encoded_tsvectors::binary>>
+      other ->
+        raise ArgumentError,
+        Postgrex.Utils.encode_msg(other, "a list of tsvectors")
```
josevalim: Should be indented with two spaces or be on the same line as previous line.

lib/postgrex/extensions/tsvector.ex
josevalim: I like the atoms idea.

lib/postgrex/extensions/tsvector.ex
```diff
@@ -0,0 +1,96 @@
...
+    end
+  end
+
+  def decode(_) do
+    quote do
+      <<len::int32, value::binary-size(len)>> ->
+        <<nb_lexemes::int32, words::binary>> = value
+        unquote(__MODULE__).decode_tsvector_values(:binary.bin_to_list(words))
```
josevalim: I think converting everything to a list is unnecessary. You can call `decode_tsvector_values(words)` and do the whole decoding using binaries:

```elixir
def decode_tsvector_values("") do
  []
end

def decode_tsvector_values(words) do
  [word, <<positions_count::16, rest::binary>>] = :binary.split(words, <<0>>)
  positions_bytes = positions_count * 2
  <<positions::binary-size(positions_bytes), rest::binary>> = rest
  positions = for <<weight::2, position::14 <- positions>>, do: {position, decode_weight(weight)}
  [%Postgrex.Lexeme{word: word, positions: positions} | decode_tsvector_values(rest)]
end

defp decode_weight(0), do: nil
defp decode_weight(1), do: :C
defp decode_weight(2), do: :B
defp decode_weight(3), do: :A
```

That should cover most of decoding but I haven't tested it.

ddrmanxbxfr: I converted it to use the binary format everywhere. It should be an improvment.

Thanks!

lib/postgrex/extensions/tsvector.ex
ddrmanxbxfr: Inlined it !

lib/postgrex/extensions/tsvector.ex
ddrmanxbxfr: Let's go with the atoms then... 👍

ddrmanxbxfr:
@fishcakez I have reworked the decoding part and converted the weights formats to atoms. It should be a bit better.

Could you have a look at it ?

Thanks!

josevalim:
Please ignore the elixir-bot notifications. It seems it is using the latest
Credo version and we don't agree with all of those warnings... :)
-- 


ddrmanxbxfr:
@fishcakez ping ?

fishcakez:
@ddrmanxbxfr sorry I will catch up on this tomorrow :)

josevalim:

PR #295, opened 2017-02-15
allows naming Postgrex.Notifications server. Closes #293

### LINKED ISSUE #293: Naming the Postgrex.Notifications server
As described by https://elixirforum.com/t/supervising-postgrex-notifications-process/722:

> Hello!
> 
> I have a use case involving a reactive application that reacts to postgres events using NOTIFY. I found this [example](https://elixirforum.com/clicks/track?url=https%3A%2F%2Fmedium.com%2F%40kaisersly%2Fpostgrex-notifications-759574f5796e%23.vpm3ju3ji&post_id=4620&topic_id=722) on how to set it up using postgrex, but I found it weird that it was starting the Postgrex.Notifications process as a child within the Posts.PgListener process (the last code example).
> 
> What would be the advantages to do this as opposed to having the Supervisor start and own the Postgrex.Notifications process?
> 
> Also, I was facing some challenges supervising it on the top level because the current implementation doesn't seem to support the name binding. In situations like these, is it more preferable to update the implementation to support name binding, or is there a way to extract the pid out of the supervisor (so that other processes can talk to it)?
> 
> Thanks!

Given that no one came up with a solution, shouldn't Postgrex.Notifications.start_link support the name option?

capitalist:


fishcakez:
Thank you :heart:

PR #296, opened 2017-02-15
Adds support for .pgpass files. Closes #289

### LINKED ISSUE #289: use .pgpass file on connection
Putting in a feature request to support the .pgpass file, as described here:  https://www.postgresql.org/docs/current/static/libpq-pgpass.html

According to the docs for start_link:

```
:hostname - Server hostname (default: PGHOST env variable, then localhost);
:port - Server port (default: PGPORT env variable, then 5432);
:database - Database (default: PGDATABASE env variable; otherwise required);
:username - Username (default: PGUSER env variable, then USER env var);
:password - User password (default: PGPASSWORD env variable); 
```

Would be great if I could also rely on the .pgpass file, as most other languages' psql connectors can do.

Brand new to elixir, just testing things out, so I can't contribute code for this.

capitalist:
Supports the format as described here.
https://www.postgresql.org/docs/current/static/libpq-pgpass.html

### DIFF
--- a/config/config.exs
+++ b/config/config.exs
@@ -2,5 +2,4 @@ use Mix.Config
 
 if Mix.env == :test do
   config :ex_unit, :assert_receive_timeout, 1000
-  config :postgrex, :pgpass, "test/support/pgpass"
 end
--- a/config/config.exs
+++ b/config/config.exs
@@ -2,4 +2,5 @@ use Mix.Config
 
 if Mix.env == :test do
   config :ex_unit, :assert_receive_timeout, 1000
+  config :postgrex, :pgpass, "test/support/pgpass"
 end
--- a/lib/postgrex.ex
+++ b/lib/postgrex.ex
@@ -35,8 +35,8 @@ defmodule Postgrex do
     * `:hostname` - Server hostname (default: PGHOST env variable, then localhost);
     * `:port` - Server port (default: PGPORT env variable, then 5432);
     * `:database` - Database (default: PGDATABASE env variable; otherwise required);
-    * `:username` - Username (default: PGUSER env variable, then USER env var);
-    * `:password` - User password (default: PGPASSWORD env variable);
+    * `:username` - Username (default: PGUSER env variable, then checks ~/.pgpass, then USER env var);
+    * `:password` - User password (default: PGPASSWORD env variable, then checks ~/.pgpass);
     * `:parameters` - Keyword list of connection parameters;
     * `:timeout` - Socket receive timeout when idle in milliseconds (default:
     `#{@timeout}`);
--- a/lib/postgrex.ex
+++ b/lib/postgrex.ex
@@ -39,7 +39,7 @@ defmodule Postgrex do
     * `:password` - User password (default: PGPASSWORD env variable, then checks pgpass);
     * `:passfile` - A credentials file using the .pgpass file format:
     https://www.postgresql.org/docs/current/static/libpq-pgpass.html
-    (default: ~/.pgpass on unix %APPDATA%/postgresql/pgpass.conf on win32)
+    (default: PGPASSFILE env variable, then ~/.pgpass on unix %APPDATA%/postgresql/pgpass.conf on win32)
     * `:parameters` - Keyword list of connection parameters;
     * `:timeout` - Socket receive timeout when idle in milliseconds (default:
     `#{@timeout}`);
--- a/lib/postgrex.ex
+++ b/lib/postgrex.ex
@@ -35,8 +35,11 @@ defmodule Postgrex do
     * `:hostname` - Server hostname (default: PGHOST env variable, then localhost);
     * `:port` - Server port (default: PGPORT env variable, then 5432);
     * `:database` - Database (default: PGDATABASE env variable; otherwise required);
-    * `:username` - Username (default: PGUSER env variable, then checks ~/.pgpass, then USER env var);
-    * `:password` - User password (default: PGPASSWORD env variable, then checks ~/.pgpass);
+    * `:username` - Username (default: PGUSER env variable, then USER env var);
+    * `:password` - User password (default: PGPASSWORD env variable, then checks pgpass);
+    * `:passfile` - A credentials file using the .pgpass file format:
+    https://www.postgresql.org/docs/current/static/libpq-pgpass.html
+    (default: ~/.pgpass on unix %APPDATA%/postgresql/pgpass.conf on win32)
     * `:parameters` - Keyword list of connection parameters;
     * `:timeout` - Socket receive timeout when idle in milliseconds (default:
     `#{@timeout}`);
--- a/lib/postgrex/pgpass.ex
+++ b/lib/postgrex/pgpass.ex
@@ -50,7 +50,7 @@ defmodule Postgrex.Pgpass do
 
   defp readable_pgpass_file? do
     with {:ok, stat}     <- File.stat(@pgpass_path),
-         0o0600          <- stat.mode &&& 0o0777, # don't read pgpass without proper permissions
+         # 0o0600          <- stat.mode &&& 0o0777, # don't read pgpass without proper permissions
     do: true,
     else: (_ -> false)
   end
--- a/lib/postgrex/pgpass.ex
+++ b/lib/postgrex/pgpass.ex
@@ -1,0 +1,51 @@
+defmodule Postgrex.Pgpass do
+  @moduledoc """
+  Parses .pgpass file to obtain database credentials
+  """
+
+  @pgpass_path Application.get_env(:postgrex, :pgpass, Path.join(System.user_home, '.pgpass'))
+  use Bitwise, only_operators: true
+
+  @doc """
+  Obtains the username given a keyword list containing hostname, database & port
+  """
+  def username(opts) do
+    with [_h, _p, _d, username, _pwd] <- pgpass_for(opts[:hostname], opts[:database], opts[:port]) do
+      username
+    else
+      _ -> nil
+    end
+  end
+
+  @doc """
+  Obtains the password given a keyword list containing hostname, database, port and optionally the username.
+  """
+  def password(opts) do
+    with [_h, _p, _d, _u, password] <- pgpass_for(opts[:hostname], opts[:database], opts[:port], opts[:username]) do
+      password
+    else
+      _ -> nil
+    end
+  end
+
+  defp pgpass_for(hostname, database, port, username \\ nil) do
+    with {:ok, stat}     <- File.stat(@pgpass_path),
+         0o0600          <- stat.mode &&& 0o0777, # don't read pgpass without proper permissions
+         {:ok, contents} <- File.read(@pgpass_path) do
+      contents
+      |> parse
+      |> Enum.filter(&(Enum.at(&1, 0) == "*" || (hostname && hostname == Enum.at(&1, 0)))) # matches hostname
+      |> Enum.filter(&(Enum.at(&1, 1) == "*" || (port && to_string(port) == Enum.at(&1, 1)))) # matches port
+      |> Enum.filter(&(Enum.at(&1, 2) == "*" || (database && database == Enum.at(&1, 2)))) # matches database
+      |> Enum.filter(&(Enum.at(&1, 3) == "*" || !username || (username && username == Enum.at(&1, 3)))) # matches username
+      |> List.first
+    end
+  end
+
+  defp parse(contents) do
+    contents
+    |> String.split("\n")
+    |> Enum.reject(&(String.match?(&1,~r/^#/)))
+    |> Enum.map(&(String.split(&1, ":")))
+  end
+end
--- a/lib/postgrex/pgpass.ex
+++ b/lib/postgrex/pgpass.ex
@@ -6,30 +6,19 @@ defmodule Postgrex.Pgpass do
   use Bitwise, only_operators: true
 
   @doc """
-  Obtains the username given a keyword list containing hostname, database & port
-  """
-  def username(opts) do
-    with [_h, _p, _d, username, _pwd] <- pgpass_for(opts[:hostname], opts[:database], opts[:port]) do
-      username
-    else
-      _ -> nil
-    end
-  end
-
-  @doc """
   Obtains the password given a keyword list containing hostname, database, port and optionally the username.
   """
   def password(opts) do
-    with [_h, _p, _d, _u, password] <- pgpass_for(opts[:hostname], opts[:database], opts[:port], opts[:username]) do
+    with [_h, _p, _d, _u, password] <- pgpass_for(opts[:hostname], opts[:database], opts[:port], opts[:username], opts[:passfile]) do
       password
     else
       _ -> nil
     end
   end
 
-  defp pgpass_for(hostname, database, port, username \\ nil) do
-    with true <- readable_pgpass_file?(),
-         {:ok, contents} <- File.read(pgpass_path()),
+  defp pgpass_for(hostname, database, port, username, passfile) do
+    with {:ok, path} <- readable_pgpass_file?(passfile),
+         {:ok, contents} <- File.read(path),
          entries when is_list(entries) <- parse(contents) do
       entries
       |> filter_hostname(hostname)
@@ -47,17 +36,45 @@ defmodule Postgrex.Pgpass do
     |> Enum.map(&(String.split(&1, ":")))
   end
 
-  defp readable_pgpass_file? do
-    with {:ok, stat} <- File.stat(pgpass_path()),
+  defp readable_pgpass_file?(nil) do
+    # when the pgpass is not passed via the :passfile option, simply ignore and move on
+    with default_path <- pgpass_path(),
+         {:ok, stat} <- File.stat(default_path),
          0o0600 <- stat.mode &&& 0o0777,
-    do: true,
+    do: {:ok, default_path},
     else: (_ -> false)
+  end
+  defp readable_pgpass_file?(passfile) do
+    # when using the :passfile options, the file must be readable or an error is raised
+    case File.stat(passfile) do
+      {:ok, stat} ->
+         case stat.mode &&& 0o0777 do
+           0o0600 ->
+             {:ok, passfile}
+           permissions_decimal ->
+             permissions =
+               permissions_decimal
+               |> Integer.digits(8)
+               |> Enum.join
+            raise RuntimeError, "passfile #{passfile} must have permissions 0600; yours was #{permissions}"
+         end
+      {:error,:enoent} ->
+        raise RuntimeError, "passfile '#{passfile}' does not exist"
+      {:error,:eacces} ->
+        raise RuntimeError, "passfile '#{passfile}' cannot be read"
+      {:error,:eisdir} ->
+        raise RuntimeError, "passfile '#{passfile}' is a directory"
+      {:error,:enomem} ->
+        raise RuntimeError, "passfile '#{passfile}' is to big to read into memory"
+      _ ->
+        raise RuntimeError, "passfile '#{passfile}' is invalid"
+    end
   end
 
   defp filter_hostname(entries, hostname), do: Enum.filter(entries, &(Enum.at(&1, 0) == "*" || (hostname && hostname == Enum.at(&1, 0))))
   defp filter_port(entries, port), do: Enum.filter(entries, &(Enum.at(&1, 1) == "*" || (port && to_string(port) == Enum.at(&1, 1))))
   defp filter_database(entries, database), do: Enum.filter(entries, &(Enum.at(&1, 2) == "*" || (database && database == Enum.at(&1, 2))))
-  defp filter_username(entries, username), do: Enum.filter(entries, &(Enum.at(&1, 3) == "*" || !username || (username && username == Enum.at(&1, 3))))
+  defp filter_username(entries, username), do: Enum.filter(entries, &(Enum.at(&1, 3) == "*" || (username && username == Enum.at(&1, 3))))
 
   defp pgpass_path, do: System.get_env("PGPASSFILE") || os_default_location()
 
@@ -68,6 +85,5 @@ defmodule Postgrex.Pgpass do
       _ -> nil
     end
   end
-
 
 end
--- a/lib/postgrex/pgpass.ex
+++ b/lib/postgrex/pgpass.ex
@@ -59,5 +59,15 @@ defmodule Postgrex.Pgpass do
   defp filter_database(entries, database), do: Enum.filter(entries, &(Enum.at(&1, 2) == "*" || (database && database == Enum.at(&1, 2))))
   defp filter_username(entries, username), do: Enum.filter(entries, &(Enum.at(&1, 3) == "*" || !username || (username && username == Enum.at(&1, 3))))
 
-  defp pgpass_path, do: System.get_env("PGPASSFILE") || Path.join(System.user_home, ".pgpass")
+  defp pgpass_path, do: System.get_env("PGPASSFILE") || os_default_location()
+
+  defp os_default_location do
+    case :os.type do
+      {:unix, _} -> Path.join(System.user_home, ".pgpass")
+      {:win32, _} -> System.get_env("APPDATA") && Path.join(System.get_env("APPDATA"), "postgresql/pgpass.conf")
+      _ -> nil
+    end
+  end
+
+
 end
--- a/lib/postgrex/pgpass.ex
+++ b/lib/postgrex/pgpass.ex
@@ -3,10 +3,6 @@ defmodule Postgrex.Pgpass do
   Parses .pgpass file to obtain database credentials
   """
 
-  defmodule PassfileError do
-    defexception message: "passfile error"
-  end
-
   use Bitwise, only_operators: true
 
   @doc """
@@ -47,9 +43,8 @@ defmodule Postgrex.Pgpass do
          0o0600 <- stat.mode &&& 0o0777
     do {:ok, default_path}
     else
-      permissions_decimal when is_integer(permissions_decimal) ->
-        permissions = int_to_octal(permissions_decimal)
-        IO.warn "WARNING: passfile \"#{pgpass_path()}\" has group or world access (#{permissions}); permissions should be u=rw (0600) or less"
+      permissions when is_integer(permissions) ->
+        IO.warn "WARNING: passfile \"#{pgpass_path()}\" has group or world access (#{inspect(permissions, [base: :octal])}); permissions should be u=rw (0600) or less"
         false
       _ ->
         false
@@ -61,9 +56,8 @@ defmodule Postgrex.Pgpass do
          0o0600 <- stat.mode &&& 0o0777 do
       {:ok, passfile}
     else
-      permissions_decimal ->
-        permissions = int_to_octal(permissions_decimal)
-        raise PassfileError, "ERROR: passfile \"#{passfile}\" has group or world access (#{permissions}); permissions should be u=rw (0600) or less"
+      permissions ->
+        raise RuntimeError, "ERROR: passfile \"#{passfile}\" has group or world access (#{inspect(permissions, [base: :octal])}); permissions should be u=rw (0600) or less"
     end
   end
 
@@ -81,7 +75,5 @@ defmodule Postgrex.Pgpass do
       _ -> nil
     end
   end
-
-  defp int_to_octal(x), do: x |> Integer.digits(8) |> Enum.join
 
 end
--- a/lib/postgrex/pgpass.ex
+++ b/lib/postgrex/pgpass.ex
@@ -50,8 +50,7 @@ defmodule Postgrex.Pgpass do
 
   defp readable_pgpass_file? do
     with {:ok, stat}     <- File.stat(@pgpass_path),
-         # 0o0600          <- stat.mode &&& 0o0777, # don't read pgpass without proper permissions
-    do: true,
+    do: (stat.mode &&& 0o0777) == 0o0600 || Mix.env == :test,
     else: (_ -> false)
   end
 
--- a/lib/postgrex/pgpass.ex
+++ b/lib/postgrex/pgpass.ex
@@ -3,7 +3,6 @@ defmodule Postgrex.Pgpass do
   Parses .pgpass file to obtain database credentials
   """
 
-  @pgpass_path Application.get_env(:postgrex, :pgpass, Path.join(System.user_home, '.pgpass'))
   use Bitwise, only_operators: true
 
   @doc """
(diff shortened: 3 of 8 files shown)


capitalist:
The postgres docs state a .pgpass file must be chmod to 0600. However, git does not preserve these permissions for test/support/pgpass. So, I simply added another check for Mix.env == :test

The alternative would be to check if the test pgpass file is in use and chmod it on every test run.

Lastly, you'll probably want to Squash and Merge ;)

config/config.exs
```diff
@@ -2,4 +2,5 @@ use Mix.Config
 
 if Mix.env == :test do
   config :ex_unit, :assert_receive_timeout, 1000
+  config :postgrex, :pgpass, "test/support/pgpass"
```
fishcakez: Application config does not work well when there are multiple postgresql backends. I think we should use System.get_env `PGPASSFILE`. This can be an option `passfile` to `start_link`.

lib/postgrex/pgpass.ex
```diff
@@ -0,0 +1,62 @@
...
+  defp parse(contents) do
+    contents
+    |> String.split("\n")
+    |> Enum.reject(&(String.match?(&1,~r/^#/)))
+    |> Enum.map(&(String.split(&1, ":")))
+  end
+
+  defp readable_pgpass_file? do
+    with {:ok, stat}     <- File.stat(@pgpass_path),
+    do: (stat.mode &&& 0o0777) == 0o0600 || Mix.env == :test,
```
fishcakez: Can we set file permissions so we dont special case testing please.

Is this correct for windows?

lib/postgrex/pgpass.ex
```diff
@@ -0,0 +1,62 @@
+defmodule Postgrex.Pgpass do
+  @moduledoc """
+  Parses .pgpass file to obtain database credentials
+  """
+
+  @pgpass_path Application.get_env(:postgrex, :pgpass, Path.join(System.user_home, '.pgpass'))
```
fishcakez: Is this correct for windows?

lib/postgrex/utils.ex
```diff
@@ -79,18 +79,23 @@ defmodule Postgrex.Utils do
...
     |> Enum.reject(fn {_k, v} -> is_nil(v) end)
   end
 
   defp normalize_port(port) when is_binary(port), do: String.to_integer(port)
   defp normalize_port(port), do: port
 
+  defp set_credentials(opts) do
+    opts
+    |> Keyword.put_new_lazy(:username, fn () -> System.get_env("PGUSER") || Postgrex.Pgpass.username(opts) || System.get_env("USER") end)
+    |> Keyword.put_new_lazy(:password, fn () -> System.get_env("PGPASSWORD") || Postgrex.Pgpass.password(opts) end)
```
fishcakez: Is this logic equivalent to psql?

test/login_test.exs
```diff
@@ -199,6 +199,12 @@ defmodule LoginTest do
     end
   end
 
+  test "obtains credentials via .pgpass" do
+    opts = [hostname: "localhost", database: "somedb", port: "5432"]
+    assert (System.get_env("PGUSER") || "foo") == P.Utils.default_opts(opts)[:username]
```
fishcakez: Please test with real login.

lib/postgrex/pgpass.ex
michalmuskala: Additionally this won't be correct if you compile code in one place and than move over to another machine (or just change the user).

config/config.exs
capitalist: This line is not for selecting the pgpass file at runtime. It's purely for testing purposes based on an article José wrote about testing at boundaries. http://blog.plataformatec.com.br/2015/10/mocks-and-explicit-contracts/

I agree with your recommendation on using PGPASSFILE from the env. Would this require a change to the Travis setup to make the tests pass, or do I just System.put_env on test runs?

lib/postgrex/pgpass.ex
fishcakez: Sorry I meant we would get rid of this, I was rather interested in the default path being correct for windows.

capitalist: Per the postgres docs, no.

The solution to use the PGPASSFILE from the env would obviate the need for this module attribute.

lib/postgrex/pgpass.ex
capitalist: I commented about that below. I can't figure out how to get git(GitHub) to preserve the permissions. Should I just chmod the support file in the unit's setup block?

lib/postgrex/utils.ex
capitalist: Ohhh, good question - I will research and find out. I just thought it made the most sense.

lib/postgrex/pgpass.ex
capitalist: Very good point. Based on @fishcakez comment above, I'm gonna ditch the module attribute.

config/config.exs
fishcakez: You can just use `System.put_env` yes

lib/postgrex/pgpass.ex
fishcakez: yeah perfect to do it in setup block

test/login_test.exs
capitalist: Done

lib/postgrex.ex
```diff
@@ -35,8 +35,8 @@ defmodule Postgrex do
     * `:hostname` - Server hostname (default: PGHOST env variable, then localhost);
     * `:port` - Server port (default: PGPORT env variable, then 5432);
     * `:database` - Database (default: PGDATABASE env variable; otherwise required);
-    * `:username` - Username (default: PGUSER env variable, then USER env var);
-    * `:password` - User password (default: PGPASSWORD env variable);
+    * `:username` - Username (default: PGUSER env variable, then checks ~/.pgpass, then USER env var);
+    * `:password` - User password (default: PGPASSWORD env variable, then checks ~/.pgpass);
```
fishcakez: please add `:passfile` option and document it

lib/postgrex/pgpass.ex
```diff
@@ -0,0 +1,63 @@
...
+  def password(opts) do
+    with [_h, _p, _d, _u, password] <- pgpass_for(opts[:hostname], opts[:database], opts[:port], opts[:username]) do
+      password
+    else
+      _ -> nil
+    end
+  end
+
+  defp pgpass_for(hostname, database, port, username \\ nil) do
+    with true <- readable_pgpass_file?(),
```
fishcakez: We must raise if system env exists and this is false

test/login_test.exs
```diff
@@ -199,6 +199,16 @@ defmodule LoginTest do
     end
   end
 
+  test "obtains credentials via .pgpass" do
+    Process.flag(:trap_exit, true)
+    set_pgpass_file()
+
+    opts = [hostname: "localhost", username: "postgrex_cleartext_pw", database: "postgres", port: "5432"]
+    assert "postgrex_cleartext_pw" == P.Utils.default_opts(opts)[:password]
```
fishcakez: this line not required

test/login_test.exs
```diff
@@ -199,6 +199,16 @@ defmodule LoginTest do
     end
   end
 
+  test "obtains credentials via .pgpass" do
+    Process.flag(:trap_exit, true)
+    set_pgpass_file()
+
+    opts = [hostname: "localhost", username: "postgrex_cleartext_pw", database: "postgres", port: "5432"]
```
fishcakez: please add [ sync_connect: true, backoff_type: :stop ] so we crash on error

lib/postgrex/utils.ex
```diff
@@ -79,18 +79,23 @@ defmodule Postgrex.Utils do
...
     |> Enum.reject(fn {_k, v} -> is_nil(v) end)
   end
 
   defp normalize_port(port) when is_binary(port), do: String.to_integer(port)
   defp normalize_port(port), do: port
 
+  defp set_credentials(opts) do
+    opts
+    |> Keyword.put_new_lazy(:username, fn () -> System.get_env("PGUSER") || System.get_env("USER") || Postgrex.Pgpass.username(opts) end)
+    |> Keyword.put_new_lazy(:password, fn () -> System.get_env("PGPASSWORD") || Postgrex.Pgpass.password(opts) end)
```
fishcakez: pgpass file must only be run once to prevent inconsistency

lib/postgrex/pgpass.ex
capitalist: So, like if PGPASSFILE is set but we can't read it?

lib/postgrex.ex
capitalist: Working on it.

lib/postgrex/utils.ex
capitalist: I don't follow. Like, if the pgpass file was changed between these two function calls in this pipeline?

The username discovery from .pgpass is kinda a nice bonus. psql does not do this, but if you have the credentials in your .pgpass I think it's nicer not to have to go lookup the username.

capitalist: Ok, I dropped support for the username lookups. It creates the complication of reading the pgpass twice and psql doesn't have this function.

lib/postgrex.ex
capitalist: Done

lib/postgrex/pgpass.ex
capitalist: Done

test/login_test.exs
capitalist: Done

test/login_test.exs
capitalist: And, as such, has been removed.

lib/postgrex/pgpass.ex
```diff
@@ -47,17 +36,45 @@ defmodule Postgrex.Pgpass do
...
+    with default_path <- pgpass_path(),
+         {:ok, stat} <- File.stat(default_path),
          0o0600 <- stat.mode &&& 0o0777,
-    do: true,
+    do: {:ok, default_path},
     else: (_ -> false)
   end
+  defp readable_pgpass_file?(passfile) do
+    # when using the :passfile options, the file must be readable or an error is raised
+    case File.stat(passfile) do
```
fishcakez: We could use `File.stat!` so we don't need to raise all these `RuntimeError`s. Note that `:file.format_error` can format errors but `FileError` would be clearer I think.

lib/postgrex/pgpass.ex
```diff
@@ -47,17 +36,45 @@ defmodule Postgrex.Pgpass do
...
 
-  defp readable_pgpass_file? do
-    with {:ok, stat} <- File.stat(pgpass_path()),
+  defp readable_pgpass_file?(nil) do
+    # when the pgpass is not passed via the :passfile option, simply ignore and move on
+    with default_path <- pgpass_path(),
+         {:ok, stat} <- File.stat(default_path),
          0o0600 <- stat.mode &&& 0o0777,
-    do: true,
+    do: {:ok, default_path},
```
fishcakez: If the default file exists but has bad permissions then I think we should error because we must assume the intention was to use it.

test/login_test.exs
```diff
@@ -199,11 +199,23 @@ defmodule LoginTest do
...
 
-    opts = [hostname: "localhost", username: "postgrex_cleartext_pw", database: "postgres", port: "5432"]
+    opts = [hostname: "localhost", username: "postgrex_cleartext_pw", database: "postgres", port: "5432",
+            sync_connect: true, backoff_type: :stop]
+    assert {:ok, pid} = P.start_link(opts)
+    assert {:ok, %Postgrex.Result{}} = P.query(pid, "SELECT 123", [])
+  end
+
+  test "obtains credentials via .pgpass at passfile option" do
+    Process.flag(:trap_exit, true)
```
fishcakez: Only need to trap if the postgrex process will exit, and it doesn't in this (or the test above)

lib/postgrex.ex
```diff
@@ -35,8 +35,11 @@ defmodule Postgrex do
     * `:hostname` - Server hostname (default: PGHOST env variable, then localhost);
     * `:port` - Server port (default: PGPORT env variable, then 5432);
     * `:database` - Database (default: PGDATABASE env variable; otherwise required);
-    * `:username` - Username (default: PGUSER env variable, then checks ~/.pgpass, then USER env var);
-    * `:password` - User password (default: PGPASSWORD env variable, then checks ~/.pgpass);
+    * `:username` - Username (default: PGUSER env variable, then USER env var);
+    * `:password` - User password (default: PGPASSWORD env variable, then checks pgpass);
+    * `:passfile` - A credentials file using the .pgpass file format:
+    https://www.postgresql.org/docs/current/static/libpq-pgpass.html
+    (default: ~/.pgpass on unix %APPDATA%/postgresql/pgpass.conf on win32)
```
fishcakez: "default: PGPASSFILE` env variable then..."

fishcakez:
@jmarca would you like to try this branch as it resolves #296?

lib/postgrex/pgpass.ex
capitalist: psql only warns, then won't use the file. Which means you can fix it later - a raise here would obligate you to fix the problem. Should we warn like psql?

lib/postgrex/pgpass.ex
capitalist: I rolled a custom exception because File.Error only wants posix codes and I want to be clear on what the problem is.

lib/postgrex/pgpass.ex
```diff
@@ -40,34 +44,26 @@ defmodule Postgrex.Pgpass do
...
-        raise RuntimeError, "passfile '#{passfile}' is to big to read into memory"
-      _ ->
-        raise RuntimeError, "passfile '#{passfile}' is invalid"
+    with stat <- File.stat!(passfile),
+         0o0600 <- stat.mode &&& 0o0777 do
+      {:ok, passfile}
+    else
+      permissions_decimal ->
+        permissions = int_to_octal(permissions_decimal)
+        raise PassfileError, "ERROR: passfile \"#{passfile}\" has group or world access (#{permissions}); permissions should be u=rw (0600) or less"
```
fishcakez: `RuntimeError` is fine here I think. it won't be possible to specifically rescue this new exception  because start_link is called by a supervisor.

lib/postgrex/pgpass.ex
```diff
@@ -0,0 +1,87 @@
...
+
+  defp os_default_location do
+    case :os.type do
+      {:unix, _} -> Path.join(System.user_home, ".pgpass")
+      {:win32, _} -> System.get_env("APPDATA") && Path.join(System.get_env("APPDATA"), "postgresql/pgpass.conf")
+      _ -> nil
+    end
+  end
+
+  defp int_to_octal(x), do: x |> Integer.digits(8) |> Enum.join
```
fishcakez: `inspect(x, [base: :octal])`

capitalist: Neat

capitalist:
Whew

jmarca:
@fishcakez I will try it out this morning.

fishcakez:
> @fishcakez I will try it out this morning.

Thank you @jmarca!

jmarca:
Sorry for the delay reporting back.  I was *not* able to get this to run---yet.  That doesn't mean that the code is broken or anything, just that my half hour spent trying to get it to work with my setup failed.  I need to dig further to figure out why it is failing.  Experience so far was that the code ignored my .pgpass file.

lib/postgrex/pgpass.ex
```diff
@@ -0,0 +1,79 @@
...
+
+  @doc """
+  Obtains the password given a keyword list containing hostname, database, port and optionally the username.
+  """
+  def password(opts) do
+    with [_h, _p, _d, _u, password] <- pgpass_for(opts[:hostname], opts[:database], opts[:port], opts[:username], opts[:passfile]) do
+      password
+    else
+      _ -> nil
+    end
```
ericmj: If there's only one clause in `with` you can use `case` instead.

lib/postgrex/pgpass.ex
```diff
@@ -0,0 +1,79 @@
...
+    |> Enum.reject(&(String.match?(&1,~r/^#/)))
+    |> Enum.map(&(String.split(&1, ":")))
+  end
+
+  defp readable_pgpass_file?(nil) do
+    # when the pgpass is not passed via the :passfile option, simply ignore and move on
+    with default_path <- pgpass_path(),
+         {:ok, stat} <- File.stat(default_path),
+         0o0600 <- stat.mode &&& 0o0777
+    do {:ok, default_path}
```
ericmj: Put the body on it's own line and move `do` up on the line above. If you use keyword syntax `, do: ` you can do it like this.

lib/postgrex/utils.ex
```diff
@@ -79,11 +79,11 @@ defmodule Postgrex.Utils do
   @spec default_opts(Keyword.t) :: Keyword.t
   def default_opts(opts) do
     opts
-    |> Keyword.put_new(:username, System.get_env("PGUSER") || System.get_env("USER"))
-    |> Keyword.put_new(:password, System.get_env("PGPASSWORD"))
     |> Keyword.put_new(:database, System.get_env("PGDATABASE"))
     |> Keyword.put_new(:hostname, System.get_env("PGHOST") || "localhost")
     |> Keyword.update(:port, normalize_port(System.get_env("PGPORT")), &normalize_port/1)
+    |> Keyword.put_new(:username, System.get_env("PGUSER") || System.get_env("USER"))
+    |> Keyword.put_new_lazy(:password, fn () -> System.get_env("PGPASSWORD") || Postgrex.Pgpass.password(opts) end)
```
ericmj: Remove the parenthesis in `fn () ->`. They will cause complications on anything but the latest versions of Elixir.

test/login_test.exs
```diff
@@ -214,4 +234,10 @@ defmodule LoginTest do
   defp set_db_name(db_name) when is_binary(db_name) do
     System.put_env("PGDATABASE", db_name)
   end
+
+  defp set_pgpass_file do
+    with path <- Path.join(__DIR__, "support/pgpass"),
+         :ok <- System.put_env("PGPASSFILE", path),
+      do: File.chmod!(path, 0o0600)
```
ericmj: There is no need to use `with` for this.

fishcakez:
> I need to dig further to figure out why it is failing

@jmarca did you get anywhere with this?

jmarca:
If this isn't the right place to post this, please tell me.

Also, new to elixir, and using phoenix to try to learn by doing.

So,  bear with my verbose explanation.  Following along http://www.phoenixframework.org/docs/up-and-running, I created hello_phoenix app.

I hacked around and modified things a bit, so as to link to my version of postgrex that is located at ~/repos/elixir_related/postgrex

```
cd ~/repos/elixir_related/postgrex
git log

   postgrex version: commit c30f52ef01203f33f4af12b6538d03136bd8ea3a
```

Then back in the hello_phoenix app, in mix.exs, I did:

```
...
  # Type `mix help deps` for examples and options.
  defp deps do
    [{:phoenix, "~> 1.2.1"},
     {:phoenix_pubsub, "~> 1.0"},
     {:phoenix_ecto, "~> 3.0"},
     {:postgrex,  path: "/home/james/repos/elixir_related/postgrex", override: true }, # ">= 0.0.0"},
     {:phoenix_html, "~> 2.6"},
```

After running 

```
mix deps.get
mix ecto.create
```

I got this error:

```
21:59:43.219 [error] GenServer #PID<0.1640.0> terminating
** (KeyError) key :password not found in: [types: Postgrex.DefaultTypes, username: "slash", backoff_type: :stop, pool: DBConnection.Connection, database: "template1", otpdefmodule HelloPhoenix.Mixfile do
  use Mix.Project

  def project do
    [app: :hello_phoenix,
     version: "0.0.1",
     elixir: "~> 1.2",
     elixirc_paths: elixirc_paths(Mix.env),
     compilers: [:phoenix, :gettext] ++ Mix.compilers,
     build_embedded: Mix.env == :prod,
     start_permanent: Mix.env == :prod,
     aliases: aliases(),
     deps: deps()]
  end

  # Configuration for the OTP application.
  #
  # Type `mix help compile.app` for more information.
  def application do
    [mod: {HelloPhoenix, []},
     applications: [:phoenix, :phoenix_pubsub, :phoenix_html, :cowboy, :logger, :gettext,
_app: :hello_phoenix, repo: HelloPhoenix.Repo, adapter: Ecto.Adapters.Postgres, hostname: "127.0.0.1", pool_size: 10]
    (elixir) lib/keyword.ex:343: Keyword.fetch!/2
    (postgrex) lib/postgrex/protocol.ex:581: Postgrex.Protocol.auth_md5/4
    (postgrex) lib/postgrex/protocol.ex:475: Postgrex.Protocol.handshake/2
    (db_connection) lib/db_connection/connection.ex:134: DBConnection.Connection.connect/2
    (stdlib) proc_lib.erl:247: :proc_lib.init_p_do_apply/3
Last message: nil
State: Postgrex.Protocol
** (Mix) The database for HelloPhoenix.Repo couldn't be created: an exception was raised:
    ** (KeyError) key :password not found in: [types: Postgrex.DefaultTypes, username: "slash", backoff_type: :stop, pool: DBConnection.Connection, database: "template1", otp_app: :hello_phoenix, repo: HelloPhoenix.Repo, adapter: Ecto.Adapters.Postgres, hostname: "127.0.0.1", pool_size: 10]
        (elixir) lib/keyword.ex:343: Keyword.fetch!/2
        (postgrex) lib/postgrex/protocol.ex:581: Postgrex.Protocol.auth_md5/4
        (postgrex) lib/postgrex/protocol.ex:475: Postgrex.Protocol.handshake/2
        (db_connection) lib/db_connection/connection.ex:134: DBConnection.Connection.connect/2
        (stdlib) proc_lib.erl:247: :proc_lib.init_p_do_apply/3
```

I was suspicious that perhaps the dependency fetch was pulling in a different version of postgrex.
So as to verify that I *really* am using the dev version of postgrex and didn't accidentally pull in the canonical versions, I hacked the mix.exs  in my local version of postgrex as follows:

```
git diff


diff --git a/mix.exs b/mix.exs
index b45992f..4d4c82a 100644
--- a/mix.exs
+++ b/mix.exs
@@ -1,7 +1,7 @@
 defmodule Postgrex.Mixfile do
   use Mix.Project
 
-  @version "0.13.0"
+  @version "0.13.2"
 
   def project do
     [app: :postgrex,
@@ -30,7 +30,7 @@ defmodule Postgrex.Mixfile do
   end
 
   defp description do
-    "PostgreSQL driver for Elixir."
+    "PostgreSQL driver for Elixir, my version."
   end
 
   defp package do
lines 1-22/22 (END)
```

So then, back in my hello_phoenix app, I see:

```
 hello_phoenix$ find . -name "*postgrex*"
./_build/dev/lib/postgrex
./_build/dev/lib/postgrex/ebin/postgrex.app
```

```
hello_phoenix$ grep "PostgreSQL driver" * -R
Binary file _build/dev/lib/postgrex/ebin/Elixir.Postgrex.beam matches
_build/dev/lib/postgrex/ebin/postgrex.app:             [{description,"PostgreSQL driver for Elixir, my version."},
deps/cowlib/erlang.mk:pkg_pgsql_description = Erlang PostgreSQL driver
deps/ranch/erlang.mk:pkg_pgsql_description = Erlang PostgreSQL driver
deps/cowboy/erlang.mk:pkg_pgsql_description = Erlang PostgreSQL driver
```


```
less  _build/dev/lib/postgrex/ebin/postgrex.app 

{application,postgrex,
             [{description,"PostgreSQL driver for Elixir, my version."},
    ....
                        'Elixir.String.Chars.Postgrex.Stream']},
              {registered,[]},
              {vsn,"0.13.2"},
              {applications,[kernel,stdlib,elixir,logger,db_connection,
                             decimal,crypto]},
              {mod,{'Elixir.Postgrex.App',[]}},
              {env,[{type_server_reap_after,180000}]}]}.
```

So those are my "custom" string and version number, so I know I'm getting the right version.

Looking at the test file, I tried explicitly setting the .pgpass env variable:

```
james@emma hello_phoenix$ env | grep -i pg
PGPASSFILE=/home/james/.pgpass
PGDATABASE=spatialvds
PGHOST=127.0.0.1
PGUSER=slash
```

and my ~/.pgpass has an entry:

```
127.0.0.1:5432:*:slash:correct horse battery staple
```

(except that isn't my real password).

I also tried adding another line with `template1` in place of the `*` but it made no difference.  

To prove that I *can* attach to the db from the command line:

```
 psql -U slash -h 127.0.0.1 template1
psql (9.5.3)
Type "help" for help.

template1=#   
```

So that is about the limit of my clue in elixir.  I can't get the new code to work when using it in a hello world phoenix app that I've hacked a little bit.

Hope that helps.

capitalist:
@jmarca Did you ever get it to work? Can you check the permissions on your .pgpass file - they should be 0600

jmarca:
@capitalist no sorry, I have not been able to get this to work for me.  I put up a pretty detailed run through above of what I was trying to do.

Yes, my .pgpass file is 0600

```
james@emma ~$ ls -lrta .pgpass
-rw------- 1 james users 233 Mar 14 22:10 .pgpass
```

As I think I said above, it could be the something in the intro code I am using.

I am *not* currently doing anything with elixir, but I may come back to it in the future.  That's why I posted the detailed run through above of how I'm hitting the failure...I'm not a regular elixir user and I had to redo/reremember everything!  

If someone better at elixir could run through the hello world example I described above and figure out why it doesn't work, that would be great, but I'm not well versed enough in elixir/erlang to do that right now.

capitalist:
@jmarca Thanks for that. I appreciate the time you took to describe the problem. I've got to recreate it to solve it so that helps a lot.

isaacsanders:
@fishcakez @capitalist Any word on this?

lib/postgrex/pgpass.ex
```diff
@@ -0,0 +1,78 @@
...
+  defp parse(contents) do
+    contents
+    |> String.split("\n")
+    |> Enum.reject(&(String.match?(&1,~r/^#/)))
+    |> Enum.map(&(String.split(&1, ":")))
+  end
+
+  defp readable_pgpass_file?(nil) do
+    # when the pgpass is not passed via the :passfile option, simply ignore and move on
+    with default_path <- pgpass_path(),
```
isaacsanders: Instead of putting this here, maybe put it in the `Postgrex.Utils.default_opts/1` function? This way defaults are handled in the same way as the rest of the options.

lib/postgrex/pgpass.ex
```diff
@@ -0,0 +1,78 @@
...
+      |> filter_username(username)
+      |> List.first
+    end
+  end
+
+  defp parse(contents) do
+    contents
+    |> String.split("\n")
+    |> Enum.reject(&(String.match?(&1,~r/^#/)))
+    |> Enum.map(&(String.split(&1, ":")))
```
isaacsanders: `Enum.map(&(String.split(&1, ":")))` is incorrect. The pgpassfile format can have `\`-escaped colons. `String.split/2` with the `parts` option set to 5 might be enough to ensure that the erroneous case isn't hit.

Sources: https://wiki.postgresql.org/wiki/Pgpass https://www.postgresql.org/docs/current/static/libpq-pgpass.html

isaacsanders: I might suggest making the records into tuples, maps, or structs. Their access pattern is more "Elixir" than the random access of lists that is the current style.

lib/postgrex.ex
```diff
@@ -36,7 +36,10 @@ defmodule Postgrex do
     * `:port` - Server port (default: PGPORT env variable, then 5432);
     * `:database` - Database (default: PGDATABASE env variable; otherwise required);
     * `:username` - Username (default: PGUSER env variable, then USER env var);
-    * `:password` - User password (default: PGPASSWORD env variable);
+    * `:password` - User password (default: PGPASSWORD env variable, then checks pgpass);
+    * `:passfile` - A credentials file using the .pgpass file format:
+    https://www.postgresql.org/docs/current/static/libpq-pgpass.html
+    (default: PGPASSFILE env variable, then ~/.pgpass on unix %APPDATA%/postgresql/pgpass.conf on win32)
```
isaacsanders: The path in Windows is `%APPDATA%\postgresql\pgpass.conf`

lib/postgrex/pgpass.ex
isaacsanders: The current use in the `Postgrex.Utils.default_opts/1` function is not right and looks like it is causing @jmarca's issue.

lib/postgrex/utils.ex
```diff
@@ -79,11 +79,11 @@ defmodule Postgrex.Utils do
   @spec default_opts(Keyword.t) :: Keyword.t
   def default_opts(opts) do
     opts
-    |> Keyword.put_new(:username, System.get_env("PGUSER") || System.get_env("USER"))
-    |> Keyword.put_new(:password, System.get_env("PGPASSWORD"))
     |> Keyword.put_new(:database, System.get_env("PGDATABASE"))
     |> Keyword.put_new(:hostname, System.get_env("PGHOST") || "localhost")
     |> Keyword.update(:port, normalize_port(System.get_env("PGPORT")), &normalize_port/1)
+    |> Keyword.put_new(:username, System.get_env("PGUSER") || System.get_env("USER"))
```
isaacsanders: @capitalist Modifying the `default_opts/1` function to something like below should fix the issues that @jmarca is seeing, or at least fix issues that do exist in the current implementation.

```
  def default_opts(opts) do
    password_infering_opts = opts
    |> Keyword.put_new(:database, System.get_env("PGDATABASE"))
    |> Keyword.update(:port, normalize_port(System.get_env("PGPORT")), &normalize_port/1)
    |> Keyword.put_new(:username, System.get_env("PGUSER") || System.get_env("USER"))
    |> Keyword.put_new(:pgpassfile, System.get_env("PGPASSFILE"))
    |> Keyword.put_new_lazy(:pgpassfile, &Postgrex.Pgpass.os_default_location/0)

    password_infering_opts
    |> Keyword.put_new_lazy(:password, fn -> System.get_env("PGPASSWORD") || Postgrex.Pgpass.password(password_infering_opts) end)
    |> Keyword.put_new(:types, Postgrex.DefaultTypes)
  end
```

isaacsanders: @capitalist There may be additional modifications necessary to the code to make everything work. If you would accept a pull request into your branch, I can work on it.

capitalist:
@isaacsanders happy to accept a PR to my fork+branch so they commits are added to this PR.

isaacsanders:
@capitalist Deal.

isaacsanders:
@fishcakez Is there an expectation that the configuration file is the source of truth?

Or are the environment variables meant to override the configuration?

fishcakez:
> Or are the environment variables meant to override the configuration?

Options set in config.exs for Ecto or passed as the options to `Postgrex.start_link` should overrule the environment variables. However I think these options are able to interact with .pgpass. We would like to behave the same as `psql`.

isaacsanders:
@fishcakez Can Postgrex interface through a unix domain socket? There are parts of the pgpass implementation that are related to sockets.

isaacsanders:
I refer to the libpq implementation

fishcakez:
Not at the moment. It would be possible though.

PR #301, opened 2017-03-06
Postgrex Result module inconsistent docs

shdblowers:
The docs are inconsistent between whether `rows` is a list of tuples or a list of lists.

Changing the moduledocs to use list of lists, rather than list of tuples.


josevalim:

PR #304, opened 2017-03-14
Add info about pgpool

mkaszubowski:
When the `prepare: :unnamed` weren't used, all the queries timed out. Adding this option seems to have solved the problem, so it's worth adding the information about this case in the docs.

### DIFF
--- a/README.md
+++ b/README.md
@@ -103,12 +103,13 @@ query("select nextval($1::text::regclass)", ["some_sequence"])
 query("select nextval($1)", [sequence_oid])
 ```
 
-## PgBouncer
+## PgBouncer / pgpool
 
 When using PgBouncer with transaction or statement pooling named prepared
 queries can not be used because the bouncer may route requests from the same
 postgrex connection to different PostgreSQL backend processes and discards named
-queries after the transactions closes. To force unnamed prepared queries:
+queries after the transactions closes. The issue may also appear when using pgpool.
+To force unnamed prepared queries:
 
 ```elixir
 Postgrex.start_link(prepare: :unnamed)


fishcakez:
We should get an postgres error, not a query timeout. Could you open an issue about that? I will merge this though because it is good fix - but we still have another issue somewhere. Its possible its `pgpool` though.

fishcakez:
Thank you :heart:

mkaszubowski:
Of course, I'll open a separate issue soon :)

mkaszubowski:
Actually, it seems the timeout comes from `DBConnection` rather then from Postgrex and the error is returned correctly. 

```elixir
(DBConnection.ConnectionError) tcp recv: closed
    (ecto) lib/ecto/adapters/postgres/connection.ex:86: Ecto.Adapters.Postgres.Connection.prepare_execute/5
    (ecto) lib/ecto/adapters/sql.ex:243: Ecto.Adapters.SQL.sql_call/6
    (ecto) lib/ecto/adapters/sql.ex:431: Ecto.Adapters.SQL.execute_and_cache/7
    (ecto) lib/ecto/repo/queryable.ex:130: Ecto.Repo.Queryable.execute/5
    (ecto) lib/ecto/repo/queryable.ex:35: Ecto.Repo.Queryable.all/4
17:04:47.236 [error] Postgrex.Protocol (#PID<0.1471.0>) disconnected: ** (DBConnection.ConnectionError) client #PID<0.1441.0> timed out because it checked out the connection for longer than 15000ms
```

Should I open an issue in this case? (personally, I bet it's an issue with pgpool, not with postgrex)

fishcakez:
Postgrex is creating the DBConnection.ConnectionError, and its because as you said query takes too long. However if both `postgrex` and `pgpool` used postgres protocol correctly then we would get an postgres error - and not a timeout waiting for a message that we can only assume is not going to come.

PR #317, opened 2017-05-29
Reload types on unknown oid during prepare

### LINKED ISSUE #243: Reload `oid` types when a new type is detected
Since we are already loading types initially, I think it would be amazing if we can trigger a refresh of the stored types when a new type is encountered.

Running some migrations is very painful with Postgrex & Ecto right now, because existing connections become broken.  I currently have to restart my web server after running migrations due to this, which I hope shouldn't be necessary long term.

``` elixir
{:ok, pid} = Postgrex.start_link(hostname: "localhost", username: "postgres", password: "postgres", database: "postgres")
Postgrex.query!(pid, "create type test_type as (name varchar);", [])
Postgrex.query!(pid, "select '(name)'::test_type", [])

** (RuntimeError) oid `319629` was not bootstrapped and lacks type information
[error] Postgrex.Protocol (#PID<0.350.0>) disconnected: ** (RuntimeError) oid `319629` was not bootstrapped and lacks type information
    (db_connection) lib/db_connection.ex:481: DBConnection.prepare_execute!/4
         (postgrex) lib/postgrex.ex:146: Postgrex.query!/4
```

fishcakez:
Closes #243. This is WIP.

- [x] Add more tests
- [x] Log reload

@mspanc can you check if this branch resolves your issue?

### DIFF
--- a/lib/postgrex/protocol.ex
+++ b/lib/postgrex/protocol.ex
@@ -973,6 +973,9 @@ defmodule Postgrex.Protocol do
   end
 
   defp reload_spawn(s, status, %Query{ref: ref} = query, oid, buffer) do
+    Logger.warn(fn() ->
+      [inspect(query) | " uses unknown oid `#{oid}` causing bootstrap"]
+    end)
     {_, mon} = spawn_monitor(fn() -> reload_lock(s, status, ref, buffer) end)
     receive do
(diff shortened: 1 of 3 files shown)


fishcakez:
This is ready to go.

fishcakez:
Oh the tests can't use range because it requires 9.2

PR #320, opened 2017-07-07
Avoid loading information about table-types at start

sumerman:
In one of our analytical databases, we have approx. 200k tables, this way we have more than 400k types in the result when you factor in array types automatically created by postgres. Therefore this bootstrap statement runs long enough to cause all other connection processes that are waiting in `fetch` to exit which, in turn, causes a restart of the pool's supervisor, thus type info is never obtained. While relaxing pool's `max_restarts` helps to mask the problem the bootstrap query still runs for tens of seconds and obstructs any useful work. 

Proposed change avoids loading potentially enormous chunk of data at start, but will load everything if a subsequent bootstrap is triggered (e.g by `prepare`)

In the case of our app it cuts initial bootstrap runtime from tens of seconds down to ~300ms and subsequent bootstrap queries are avoided.

### DIFF
--- a/lib/postgrex/types.ex
+++ b/lib/postgrex/types.ex
@@ -57,7 +57,14 @@ defmodule Postgrex.Types do
     filter_oids =
       case oids do
         [] ->
-          ""
+          # avoid loading information about table-types
+          # since there might be a lot them and most likely
+          # they won't be used; subsequent bootstrap will
+          # fetch them along with any other "new" types
+          """
+          WHERE (t.typrelid=0)
+          AND (t.typelem = 0 OR t.typelem NOT IN (SELECT oid FROM pg_catalog.pg_type WHERE typrelid!=0))
+          """
         _  ->
           # equiv to `WHERE t.oid NOT IN (SELECT unnest(ARRAY[#{Enum.join(oids, ",")}]))`
           # `unnest` is not supported in redshift or postgres version prior to 8.4


sumerman:
The failed build is due to failed dependency fetch.

fishcakez:
Hi @sumerman. That's a lot of tables! AFAIK we had resolved the bootstrapping issue for people with 10s of thousands, rather than 100s of thousands. I can see this being a real problem for you! I am concerned that this approach may delay bootstrapping to more time sensitive parts in other peoples projects. Perhaps we could offer a switch to make this more manageable for people so people could opt into lazy bootstrapping that might limit to the built-in oids on initial bootstrap and then fetch oids as required when a prepare fails to find those required (I think we call it reload). What do you think?

sumerman:
It's indeed not a common workload 😄 Current approach only postpones loading of table-types and their array counterparts, those are unlikely to ever hit a wire. (The only case I can think of is `SELECT foo()` where `foo` returns a table-type and that's not common at all). 

One way to make it strictly better is to always skip table-types unless an OID that caused a reload is of table type. It would retain current behavior, for the most part, yet touch table types, the main source of bloat, only when absolutely necessary.

fishcakez:
@sumerman I think that sounds like a good. I think this also highlights that we could end up with an expensive bootstrap during a query: https://github.com/elixir-ecto/postgrex/blob/2c026cb29c4f8e62ce1b2fec46f444fac440ff4d/lib/postgrex/protocol.ex#L890. In that situation we could improve the situation by just bootstrapping the missing oids for that query (currenty we just find the first missing oid). If we avoid bootstrapping tables at connect time (`bootstrap`) and bootstrap just the oids at query time (`reload`) i think that would be a very neat improvement as we minimise handshaking and add the smallest possible overhead for a query that misses types. Would you like to handle it?

fishcakez:
@sumerman actually we could resolve #310 by only doing the bootstrap (minus tables) on first connect and then only doing bootstrap in the lazy fashion of `reload` for required types.

sumerman:
@fishcakez I glanced thru `protocol.ex`, it seems quite intricate, esp. flow-switching between query execution and bootstrap via `reload`. Given that the change you are proposing seems to imply some refactoring of what was done in #317 I believe you are better equipped to tackle this. I can try to do this on my own applying the least invasive change to the existing flow and changing bootstrap to load only certain OIDs if provided.

PR #321, opened 2017-07-11
Update to latest DBConnection

fishcakez:
WIP: awaiting DBConnection release

* Adds support for non-stream cursors
* Adds support for advance transaction handling

### DIFF
--- a/lib/postgrex/protocol.ex
+++ b/lib/postgrex/protocol.ex
@@ -241,7 +241,7 @@ defmodule Postgrex.Protocol do
       %{postgres: {_, _}} ->
         lock_error(s, :execute, copy)
       _ ->
-        handle_close_portal(copy, opts, s)
+        handle_close_portal(copy, mode(opts), opts, s)
     end
   end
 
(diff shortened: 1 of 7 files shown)


lib/postgrex/protocol.ex
```diff
@@ -340,65 +330,75 @@ defmodule Postgrex.Protocol do
 
   @spec handle_begin(Keyword.t, state) ::
     {:ok, Postgrex.Result.t, state} |
-    {:error, Postgrex.Error.t, state} |
-    {:error | :disconnect, %RuntimeError{}, state} |
-    {:disconnect, %DBConnection.ConnectionError{}, state}
+    {DBConnecation.status, state} |
```
samphilipd: Seems like a typo?

fishcakez: Thanks fixed.

fishcakez: Thanks fixed.

fishcakez:
`max_rows` should be a `fetch` only option and not stored in cursor, the behaviour in this branch is confusing.

The `mode` should be stored in cursor so that when the cursor is deallocated the savepoint is cleaned up if option is not passed. This problem is a particular issue when copying from the database.

josevalim:
@fishcakez I believe I have addressed the two concerns you had left. :)

fishcakez:
@josevalim yes the changes are good 👍

Issue #324, opened 2017-08-22
Can not establish SSL connection when :ssl app is not started


AndrewDryga:
While running migrations ([here is a code](https://gist.github.com/AndrewDryga/a9172fbd9351d0b89092ff35d5912b40/b5ba6ee5390f263cf855fe852322c3a42675f084)) we have faced a weird issue:
```
Loading app..
Starting dependencies..
Starting repos..
Running migrations for myapp
{"init terminating in do_boot",{#{'__exception__'=>true,'__struct__'=>'Elixir.DBConnection.ConnectionError',message=><<"connection not available because of disconnection">>},[{'Elixir.DBConnection',checkout,2,[{file,"lib/db_connection.ex"},{line,926}]},{'Elixir.DBConnection',run,3,[{file,"lib/db_connection.ex"},{line,742}]},{'Elixir.DBConnection',run_meter,3,[{file,"lib/db_connection.ex"},{line,1133}]},{'Elixir.DBConnection',prepare_execute,4,[{file,"lib/db_connection.ex"},{line,584}]},{'Elixir.Ecto.Adapters.Postgres.Connection',execute,4,[{file,"lib/ecto/adapters/postgres/connection.ex"},{line,93}]},{'Elixir.Ecto.Adapters.SQL',sql_call,6,[{file,"lib/ecto/adapters/sql.ex"},{line,243}]},{'Elixir.Ecto.Adapters.SQL','query!',5,[{file,"lib/ecto/adapters/sql.ex"},{line,193}]},{'Elixir.Ecto.Adapters.Postgres','-execute_ddl/3-fun-0-',4,[{file,"lib/ecto/adapters/postgres.ex"},{line,86}]}]}}
init terminating in do_boot ()
Crash dump is being written to: erl_crash.dump...done
```

PosgtrSQL log has an error entry:
```
2017-08-22T10:26:28+00:00 app[postgres.15761]: [DATABASE] [5-1]  sql_error_code = 08P01 LOG:  could not accept SSL connection: EOF detected
```

This were fixed by starting :ssl app before Postgrex, but the error is not clear enough and I don't think that we should try to establish an SSL connection when `:ssl` is not started. Maybe let's add `Application.ensure_started/1` before establishing an SSL connection?

josevalim:
Doing application:ensure_started will hide bugs, because it will work in development but not if you build a release, as ssl may not actually be there.

AndrewDryga:
[...]

PR #325, opened 2017-08-29
Raise error when :ssl application is required and not started

AndrewDryga:
Related to #324

### DIFF
--- a/lib/postgrex.ex
+++ b/lib/postgrex.ex
@@ -83,6 +83,7 @@ defmodule Postgrex do
   """
   @spec start_link(Keyword.t) :: {:ok, pid} | {:error, Postgrex.Error.t | term}
   def start_link(opts) do
+    ensure_deps_started!(opts)
     opts = Postgrex.Utils.default_opts(opts)
     DBConnection.start_link(Postgrex.Protocol, opts)
   end
@@ -453,5 +454,18 @@ defmodule Postgrex do
   ## Helpers
   defp defaults(opts) do
     Keyword.put_new(opts, :timeout, @timeout)
+  end
+
+  defp ensure_deps_started!(opts) do
+    if Keyword.get(opts, :ssl) and Application.start(:ssl) != {:error, {:already_started, :ssl}} do
+      raise """
+      SSL connection can not be established because `:ssl` application is not started,
+      you can add it to `extra_application` in your `mix.exs`:
+
+        def application do
+          [extra_applications: [:ssl]]
+        end
+      """
+    end
   end
 end
--- a/lib/postgrex.ex
+++ b/lib/postgrex.ex
@@ -457,7 +457,8 @@ defmodule Postgrex do
   end
 
   defp ensure_deps_started!(opts) do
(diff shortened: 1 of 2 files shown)


lib/postgrex.ex
```diff
@@ -454,4 +455,17 @@ defmodule Postgrex do
   defp defaults(opts) do
     Keyword.put_new(opts, :timeout, @timeout)
   end
+
+  defp ensure_deps_started!(opts) do
+    if Keyword.get(opts, :ssl, false) and Application.start(:ssl) != {:error, {:already_started, :ssl}} do
```
fishcakez: `Application.start` has side effects, and depending how Postgrex is used it may still work fine in dev and not in a release. For example if Postgrex is placed under a pool it may succeed on second start attempt in dev because ssl will be started by the first attempt. However in release it would fail.

AndrewDryga: @fishcakez can you suggest a better way to check is application loaded?

fishcakez: For example `Application.loaded_applications` returns a list of loaded applications. Check `Application` and `:application` modules.

AndrewDryga: @fishcakez That was my first idea how to check them, but loaded doesn't mean started. It is possible that you have :ssl application, but because it's not started you still would not be able to establish an SSL connection..

lib/postgrex.ex
```diff
@@ -454,4 +455,18 @@ defmodule Postgrex do
   defp defaults(opts) do
     Keyword.put_new(opts, :timeout, @timeout)
   end
+
+  defp ensure_deps_started!(opts) do
+    ssl_required? = Keyword.get(opts, :ssl, false)
+    if ssl_required? and Enum.any?(:application.which_applications(), fn {app, _desc, _vsn} -> app == :ssl end) do
```
AndrewDryga: `:application.which_applications/0` is a GenServer call which can result in a timeout. I don't think that we need to handle it, but decided to place this comment FYI.

josevalim: The boolean logic is inverted, you want `not any?`. Here is another way to write it though:

    if ssl_required? and List.keymember?(:application.which_applications, :ssl, 0)

AndrewDryga: Sorry for that. I had a bad time trying to run postgrex tests on my development env. Getting socket/psql errors all the time because the database is running within a container. So I'm relying on a Travis in a PR and debug via `iex` calls.

Thanks for `List.keymember?/3`, never used that module 👍.

fishcakez:
Thank you :heart:

wojtekmach:
Blast from the past!

I just realised that `ensure_deps_started!` function is only called on [start_link/1](https://github.com/elixir-ecto/postgrex/blob/v0.14.3/lib/postgrex.ex#L153), not [child_spec/1](https://github.com/elixir-ecto/postgrex/blob/v0.14.3/lib/postgrex.ex#L477), and so for a lot of people we don't actually run this check.

I'm happy to fix this however I'm wondering if Postgrex should just start ssl anyway? How often people don't have it?

josevalim:
Yes, please!

PR #327, opened 2017-09-10
Add credo configuration and fix linter errors

jlarigakis:
Added a credo config file with mostly default settings, just disabled the check on `cond` statements since it seems to be broken (https://github.com/rrrene/credo/issues/440).

Also made some minor style changes following credo's suggestions, and added `credo:disable` directives where appropriate.


jlarigakis:
Went from 430 issues to 153... all the ones left are low-priority. I guess ebert runs credo in strict mode regardless of what the config file says.

josevalim:
Thank you! This is the credo file that ebert uses: https://github.com/plataformatec/linters/blob/master/.credo.exs

Can you please remove the lines that disable credo? It is not a style that we would like to introduce and rely on ebert mostly to guarantee that future contributions will stay consistent. Thanks!

josevalim:
Ping. :)

jlarigakis:
Yeah, that makes sense about wanting to enforce consistency on future contributions. 

My goal with this PR was to be able to run `credo` locally to figure what my violations were before pushing, but there's too many for that to be realistic. Not a particularly valuable PR in the end, oh well 😛

josevalim:
No, this is very helpful, especially the `@moduledoc false` ones. Thank you! :heart: :green_heart: :blue_heart: :yellow_heart: :purple_heart:

PR #328, opened 2017-09-11
fixes custom range domains not supported natively

evadne:
For example, `CREATE DOMAIN public.validity_range AS tsrange CONSTRAINT …` will create a domain based on `tsrange`, and without the change, the end user must create a custom Postgrex extension to make use of such a type.

In this change, the relevant bootstrap query used by the Type Server to determine base types within a range type is updated to take such domain types into account, so the native Range extension can be used. In addition, the comment within Type Info was updated to reflect the true nature of the underlying value.


fishcakez:
Thank you :heart:

PR #330, opened 2017-09-15
rise ArgumentError when encoding infinite Decimal

lukaszsamson:
Fixes https://github.com/elixir-ecto/postgrex/issues/329

### DIFF
--- a/lib/postgrex/extensions/numeric.ex
+++ b/lib/postgrex/extensions/numeric.ex
@@ -27,6 +27,10 @@ defmodule Postgrex.Extensions.Numeric do
     <<0 :: int16, 0 :: int16, 0xC000 :: uint16, 0 :: int16>>
   end
 
+  def encode_numeric(%Decimal{coef: :inf} = decimal) do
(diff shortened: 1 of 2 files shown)


lib/postgrex/extensions/numeric.ex
```diff
@@ -27,6 +27,10 @@ defmodule Postgrex.Extensions.Numeric do
     <<0 :: int16, 0 :: int16, 0xC000 :: uint16, 0 :: int16>>
   end
 
+  def encode_numeric(%Decimal{coef: :inf, sign: sign}) do
+    raise ArgumentError, "Decimal value #{if sign == 1, do: "+", else: "-"}Infinity cannot be represented as numeric type"
```
josevalim: Would you get a better report if you wrote it as:

```elixir
def encode_numeric(%Decimal{coef: :inf} = decimal) do
  raise ArgumentError, "cannot represent #{inspect decimal} as numeric type"
end
```

?

lukaszsamson: Yep, looks better

josevalim:

PR #332, opened 2017-10-05
Fix empty range encoding

mememori:
As of yesterday, me and @fsabe found a potential bug, that whenever postgrex tries to encode an empty range (ie: `'(,)'` or in this case `%Postgrex.Range{lower: nil, upper: nil}`), it would fail with a "Protocol violation" error with message "insufficient data left in message".

This was due to the `encode` clause that would return `<<1>>` when both lower and upper were nil, after removing it, the range passes through the default encoding clause, succeeding.

This happened on PostgreSQL 9.6, i can't right now test on other pg envs, so i'll leave that to Travis and fix it shall it fail on any other pg version :)

With the new tests, leaving the old encoding clause, the error i would get is
```
1) test encode range (QueryTest)
     test/query_test.exs:527
     Assertion with == failed
     code:  [[empty_range]] == query("SELECT $1::daterange", [empty_range])
     left:  [[%Postgrex.Range{lower: nil, lower_inclusive: false, upper: nil,
               upper_inclusive: false}]]
     right: %Postgrex.Error{connection_id: 8524, message: nil,
             postgres: %{code: :protocol_violation, file: "pqformat.c",
               line: "557", message: "insufficient data left in message",
               pg_code: "08P01", routine: "pq_getmsgbytes", severity: "ERROR",
               unknown: "ERROR"}}
     stacktrace:
       test/query_test.exs:551: (test)
```

Now it adds a few more tests to encoding and decoding (the decoding are a bit redundant but i think it might be a good idea to include them as a safeguard) to ensure that empty ranges work as expected


josevalim:

Issue #335, opened 2017-10-18
Postgres 10 support


enilsen16:
Opening this because I couldn't find any information on the current status of this is. 

Does postgrex plan to support postgres 10 now that it is officially released?

josevalim:
@enilsen16 hopefully it just works? :)

enilsen16:
I couldn't get it to. :(

```
     ** (Postgrex.Error) ERROR XX000 (internal_error): could not load library "/usr/local/lib/postgresql/plpgsql.so": dlopen(/usr/local/lib/postgresql/plpgsql.so, 10): Symbol not found: _SPI_register_trigger_data
       Referenced from: /usr/local/lib/postgresql/plpgsql.so
       Expected in: /usr/local/opt/postgresql/bin/postgres
      in /usr/local/lib/postgresql/plpgsql.so
     stacktrace:
       (ecto) lib/ecto/adapters/sql.ex:571: Ecto.Adapters.SQL.struct/7
       (ecto) lib/ecto/repo/schema.ex:469: Ecto.Repo.Schema.apply/4
       (ecto) lib/ecto/repo/schema.ex:205: anonymous fn/13 in Ecto.Repo.Schema.do_insert/4
       (ecto) lib/ecto/repo/schema.ex:691: anonymous fn/3 in Ecto.Repo.Schema.wrap_in_transaction/6
       (ecto) lib/ecto/adapters/sql.ex:620: anonymous fn/3 in Ecto.Adapters.SQL.do_transaction/3

```

AndrewDryga:
@enilsen16 error looks like it's an internal PostgreSQL error, see the status code..

enilsen16:
@AndrewDryga Completely removing postgres and removing the old postgres 9 db seemed to fix it.

enilsen16:
Since it sounds like postgres 10 support is intended, I'll update the readme and travis. Thanks everyone!

PR #336, opened 2017-10-18
Update readme to specify postgresql 10 support

enilsen16:
ref #335 

Doesn't look like travis supports Postgres 10 yet.


josevalim:
Let's hold this until travis has postgres 10 (or we can add it manually) because otherwise we can't really guarantee that everything works on 10.

fishcakez:
Thank you :heart:

PR #337, opened 2017-10-18
fixes array domains not supported natively

evadne:
Array domains do not have `typelem` set in `pg_types` and are not supported out-of-the-box.

For example, this creates an array with a constraint ensuring that it has at least one element.

    CREATE DOMAIN public.page_quads AS polygon[]
       CONSTRAINT is_populated
       CHECK (COALESCE(array_length(VALUE, 1), 0) >= 1);

In this changeset, the bootstrap query used by the Type Server is updated again, so to support this use case without requiring a custom Postgrex extension.

Within the change, I have limited this to PostgreSQL 9.0 and upwards but the version check may actually be unnecessary.

### DIFF
--- a/test/query_test.exs
+++ b/test/query_test.exs
@@ -65,6 +65,24 @@ defmodule QueryTest do
     assert [[[[0]]]] = query("SELECT ARRAY[ARRAY[0]]", [])
   end
 
+  test "decode array domain", context do
+    assert [[[1.0, 2.0, 3.0]]] =
+           query("SELECT ARRAY[1, 2, 3]::floats_domain", [])
+
+    assert [[[%Postgrex.Point{x: 1.0, y: 1.0}, %Postgrex.Point{x: 2.0, y: 2.0}, %Postgrex.Point{x: 3.0, y: 3.0}]]] =
+           query("SELECT ARRAY[point '1,1', point '2,2', point '3,3']::points_domain", [])
+  end
+
+  test "encode array domain", context do
+    floats = [1.0, 2.0, 3.0]
+    floats_string = "{1,2,3}"
+    assert [[^floats_string]] = query("SELECT $1::floats_domain::text", [floats])
+
+    points = [%Postgrex.Point{x: 1.0, y: 1.0}, %Postgrex.Point{x: 2.0, y: 2.0}, %Postgrex.Point{x: 3.0, y: 3.0}]
+    points_string = "{\"(1,1)\",\"(2,2)\",\"(3,3)\"}"
+    assert [[^points_string]] = query("SELECT $1::points_domain::text", [points])
+  end
+
   test "decode interval", context do
     assert [[%Postgrex.Interval{months: 0, days: 0, secs: 0}]] =
            query("SELECT interval '0'", [])
--- a/lib/postgrex/types.ex
+++ b/lib/postgrex/types.ex
@@ -46,6 +46,14 @@ defmodule Postgrex.Types do
   def bootstrap_query(version, {_, table}) do
     oids = :ets.select(table, [{{:"$1", :_, :_}, [], [:"$1"]}])
 
+    {typelem, join_domain} =
+      if version >= {9, 0, 0} do
+        {"coalesce(d.typelem, t.typelem)",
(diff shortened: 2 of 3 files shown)


fishcakez:
@evadne please can we get a test for this to prevent regression :)

evadne:
@fishcakez will do, sorry I did not notice.

ps. The 9.0 test failure in Travis seems to be related to something else

fishcakez:
> ps. The 9.0 test failure in Travis seems to be related to something else

Yeah, you can ignore that.

evadne:
@fishcakez would you please help reviewing the test case as I think the Type Server should be reloaded after a new type has been added, otherwise it raises a warning. I have not been able to figure out how to trigger that within the test and also am not sure whether doing that would be correct or if all types used in the tests should be created centrally from somewhere else.

fishcakez:
@evadne create the domain in https://github.com/elixir-ecto/postgrex/blob/3a5bfa139509875b44d28a795af4d0c48c34e77b/test/test_helper.exs#L80. If possible avoid creating a table like https://github.com/elixir-ecto/postgrex/blob/285d32aacc0d00af0127fd507d62e53602d25986/test/calendar_test.exs#L13. There we use text casting so we only ever do serialization one way. This technique avoids us doing the encoding and decoding which could possibly hide bugs if mistake on both sides.

evadne:
🌈

test/query_test.exs
```diff
@@ -65,6 +65,26 @@ defmodule QueryTest do
...
+    assert :ok = query("INSERT INTO points VALUES (ARRAY[point '1,1', point '2,2', point '3,3']::points_domain)", [])
+    assert [[^points]] = query("SELECT * FROM points", [])
+  end
+
+  test "encode array domain", context do
+    floats = [1.0, 2.0, 3.0]
+    assert :ok = query("INSERT INTO floats VALUES ($1::floats_domain)", [floats])
+
+    points = [%Postgrex.Point{x: 1.0, y: 1.0}, %Postgrex.Point{x: 2.0, y: 2.0}, %Postgrex.Point{x: 3.0, y: 3.0}]
+    assert :ok = query("INSERT INTO points VALUES ($1::points_domain)", [points])
```
fishcakez: Needs to read these values casted to text to ensure we inserted them correctly.

test/query_test.exs
```diff
@@ -65,6 +65,26 @@ defmodule QueryTest do
     assert [[[[0]]]] = query("SELECT ARRAY[ARRAY[0]]", [])
   end
 
+  test "decode array domain", context do
+    floats = [1.0, 2.0, 3.0]
+    assert :ok = query("DELETE FROM floats", [])
```
fishcakez: Is it possible to make these tests pure, i.e. without a table: `SELECT ARRAY[1,2,3]::floats_domain` and `SELECT $1::floats_domain::text`?

evadne:
@fishcakez does it look better now?

test/query_test.exs
```diff
@@ -65,6 +65,22 @@ defmodule QueryTest do
...
+    assert [[[%Postgrex.Point{x: 1.0, y: 1.0}, %Postgrex.Point{x: 2.0, y: 2.0}, %Postgrex.Point{x: 3.0, y: 3.0}]]] =
+           query("SELECT ARRAY[point '1,1', point '2,2', point '3,3']::points_domain", [])
+  end
+
+  test "encode array domain", context do
+    floats = [1.0, 2.0, 3.0]
+    assert [[^floats]] = query("SELECT $1::floats_domain", [floats])
+
+    points = [%Postgrex.Point{x: 1.0, y: 1.0}, %Postgrex.Point{x: 2.0, y: 2.0}, %Postgrex.Point{x: 3.0, y: 3.0}]
+    assert [[^points]] = query("SELECT $1::points_domain", [points])
```
fishcakez: Can we cast the result to text so that we encode the parameter to binary but decode the result as text in this test, i.e. `SELECT $1::points_domain::text`.

fishcakez: The reason to do this is so that we have encode tests that don't depend on our decoding. It is okay to have an identity test too but not required.

fishcakez:
Thank you :heart:

PR #339, opened 2017-10-21
Support application name

shribe:
It's very useful to set the application name so that pg_stat_activity will give us that clue as to what connections are created by what clients. This can currently be set by adding application_name to the parameters option in start_link. I propose to: 1) set fallback_application_name to "postgrex" so that if the user doesn't specify application_name there's at least that much of a clue as to where leaked connections are coming from, and, 2) support use of the "PGAPPNAME" env variable for consistency with other current PG tools.

I don't really know how you'd want to structure this, if you'd really want this logic split over two locations as I show here, or somehow consolidated. So I'm not really thinking of this as a literal pull request so much as a feature request illustrated with code ;-)

### DIFF
--- a/lib/postgrex/protocol.ex
+++ b/lib/postgrex/protocol.ex
@@ -567,7 +567,6 @@ defmodule Postgrex.Protocol do
 
   defp startup(s, %{opts: opts} = status) do
     params = opts[:parameters] || []
-    params = Keyword.put_new(params, :fallback_application_name, "postgrex")
     user = Keyword.fetch!(opts, :username)
     database = Keyword.fetch!(opts, :database)
     msg = msg_startup(params: [user: user, database: database] ++ params)
--- a/lib/postgrex/protocol.ex
+++ b/lib/postgrex/protocol.ex
@@ -567,6 +567,7 @@ defmodule Postgrex.Protocol do
 
   defp startup(s, %{opts: opts} = status) do
     params = opts[:parameters] || []
+    params = Keyword.put_new(params, :fallback_application_name, "postgrex")
     user = Keyword.fetch!(opts, :username)
     database = Keyword.fetch!(opts, :database)
     msg = msg_startup(params: [user: user, database: database] ++ params)
--- a/lib/postgrex/utils.ex
+++ b/lib/postgrex/utils.ex
@@ -90,12 +90,10 @@ defmodule Postgrex.Utils do
 
   @spec default_app_name(Keyword.t) :: Keyword.t
   defp default_app_name(opts) do
-    if app_name = System.get_env("PGAPPNAME") do
-      params = Keyword.get(opts, :parameters, []) |> Keyword.put_new(:application_name, app_name)
-      Keyword.put(opts, :parameters, params)
-    else
-      opts
-    end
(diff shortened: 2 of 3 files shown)


lib/postgrex/protocol.ex
```diff
@@ -567,6 +567,7 @@ defmodule Postgrex.Protocol do
 
   defp startup(s, %{opts: opts} = status) do
     params = opts[:parameters] || []
+    params = Keyword.put_new(params, :fallback_application_name, "postgrex")
```
josevalim: Where is `:fallback_application_name` used?

josevalim: I think instead we can always set the application_name name to postgrex unless the user overrides it. No need for the fallback and the application name.

shribe: well, `:fallback_application_name` is actually a connection parameter recognized by PG--certainly you could structure postgrex to not need it, just using `:application_name` and defaulting as needed--it really comes down to how deeply you want that default embedded, the fallback is more useful if a very low-level lib is likely to be pulled out & re-used elsewhere

josevalim: I see, thanks! It still looks kind of unnecessary given that we can handle this logic ourselves without sending it to the DB, no?

shribe: yes, I'd be perfectly happy to not use it--setting the default in utils.ex has the added advantage of not spreading the name-assignment logic across multiple locations

shribe: so do you think it's good for the default to simply be postgrex, or should it pull from application settings somewhere, attempting to be like "postgrex:myapp"???

lib/postgrex/utils.ex
```diff
@@ -84,9 +84,18 @@ defmodule Postgrex.Utils do
     |> Keyword.put_new(:hostname, System.get_env("PGHOST") || "localhost")
     |> Keyword.update(:port, normalize_port(System.get_env("PGPORT")), &normalize_port/1)
     |> Keyword.put_new(:types, Postgrex.DefaultTypes)
+    |> default_app_name()
     |> Enum.reject(fn {_k, v} -> is_nil(v) end)
   end
 
+  @spec default_app_name(Keyword.t) :: Keyword.t
+  defp default_app_name(opts) do
+    app_name = System.get_env("PGAPPNAME") || "postgrex"
```
fishcakez: We can get the OTP application name using `:application.get_application()` but it might be `:undefined` in some non-prod cases (i.e. spawning from shell). However I would prefer not to set a default application_name because it is not obvious how to disable it. Would an empty string be equivalent it to not being set? That would be an usual pattern.

Also if we want to be able to have default parameters I think we should have a single way to do it and should be be opt in. I think the place for that would be in application env for postgrex.

fishcakez:
@shribe ping, would you like add a Application env to postgrex that contains all the default parameters and not just application name?

lib/postgrex/protocol.ex
bobkat5: Ok

josevalim:
Given we don't want to inject a parameter name as default, let's keep the mechanism as is where any parameter can be given via the `:parameters` option. Thank you!

PR #348, opened 2017-11-20
Try to cast params if param types change

fishcakez:
Casting params when preparing a query with known types will prevent errors when types have changed as we can use the encode terms for the params, the new types to decode the result and get a refreshed query struct back.

The only case when we can't handle it gracefully (no error) is if we can't cast the param types from old to new. This will produce a native Postgrex.Error so we no longer have to create a custom error and relies on the database as source. However in that situation a user should be using a new query. This shouldn't occur for Ecto as it should require a schema code reload to get into such a state. Fortunately Ecto will be able to handle the error and reset the query cache in a similar way to when the types change and the query has already been prepared.

Depends on #347.

### DIFF
--- a/lib/postgrex.ex
+++ b/lib/postgrex.ex
@@ -258,6 +258,8 @@ defmodule Postgrex do
     case DBConnection.execute(conn, query, params, defaults(opts)) do
       {:ok, _} = ok ->
         ok
+      {:ok, _query, result} ->
+        {:ok, result}
       {:error, %Postgrex.Error{}} = error ->
         error
       {:error, err} ->
@@ -269,9 +271,15 @@ defmodule Postgrex do
   Runs an (extended) prepared query and returns the result or raises
   `Postgrex.Error` if there was an error. See `execute/4`.
   """
-  @spec execute!(conn, Postgrex.Query.t, list, Keyword.t) :: Postgrex.Result.t
+  @spec execute!(conn, Postgrex.Query.t, list, Keyword.t) ::
+    Postgrex.Result.t
   def execute!(conn, query, params, opts \\ []) do
-    DBConnection.execute!(conn, query, params, defaults(opts))
+    case DBConnection.execute!(conn, query, params, defaults(opts)) do
+      {_query, result} ->
+        result
+      result ->
+        result
+    end
   end
 
   @doc """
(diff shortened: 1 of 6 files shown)


lib/postgrex.ex
```diff
@@ -269,7 +272,8 @@ defmodule Postgrex do
   Runs an (extended) prepared query and returns the result or raises
   `Postgrex.Error` if there was an error. See `execute/4`.
   """
-  @spec execute!(conn, Postgrex.Query.t, list, Keyword.t) :: Postgrex.Result.t
+  @spec execute!(conn, Postgrex.Query.t, list, Keyword.t) ::
+    Postgrex.Result.t | {Postgrex.Query.t, Postgrex.Result.t}
```
michalmuskala: I'm a little worried about this interface change. It can go unnoticed for a long time and lead to subtle and hard to detect bugs.

fishcakez: Do you want the query to only be returned for non-bang version?

michalmuskala: I'm not sure, but I would say so. With the non-bang version, the user most probably already pattern matches on the result, so the change will be noticed quickly. With `execute!` it's likely the result is not pattern matched and the error could appear somewhere completely unrelated to the call.

fishcakez: The trouble here is that if someone is using `execute!` (or preparing queries in anyway that last for longer than a single transaction) then they can't handle type changes unless they abuse internals like Ecto does. Of course Ecto can only really handle this because development is synced by `elixir-ecto` ;). Returning a query is an alternative to raising an error here. If we ignore the query then all executes take 2 roundtrips from this point forward.

fishcakez: Shall we add an `execute` option to enable this feature such as `redescribe: boolean` that defaults to false.

michalmuskala: Alternatively, we could have the `execute!` call start to always return a tuple. This should also protect from those "subtle bugs". I'm not sure, though, what we would return as the second element in that case.

fishcakez: We are able to use the `ref` value to quickly compare difference between query structs so a fast comparison can be done. It seems preferable than sometimes returning the query as testing that is harder. However it is breaking the API.

We could also ignore the query struct return, like we do for `query`, and continue to abuse internals in Ecto. Then we delay this change until we have a complete solution: https://github.com/elixir-ecto/db_connection/issues/102.

fishcakez: We'll ignore the query struct for now but can change this later.

PR #349, opened 2017-11-23
Only test decode binary on OTP <= 19

fishcakez:
This is an awkward situation as we are liberally copying (against documented advice) and in OTP 20 the binary GC is much improved. Therefore we add performance penalty in OTP 20 that shouldn't be required(?). I am unsure if we can deprecate the feature once we are OTP 20+ because I don't know if the leaks could still occur.

### DIFF
--- a/test/query_test.exs
+++ b/test/query_test.exs
@@ -331,12 +331,13 @@ defmodule QueryTest do
 
   @tag min_pg_version: "9.0"
   test "hstore copies binaries by default", context do
-    text = "hello world"
+    # For OTP 20+ refc binaries up to 64 bytes might be copied during a GC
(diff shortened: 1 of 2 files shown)


michalmuskala:
The mechanism introduced in OTP 20 works in a way that a reference into a big binary, that would otherwise fit into a heap binary is converted into a heap binary during a GC.

With the current implementation parameters, this means references smaller than 64 bytes will be copied during GC. Bigger references can still cause the same problems as before.

fishcakez:
Shall we use 65byte binaries then and always run?

michalmuskala:
Yeah, I think that would work. It's relying on internals, but then again this feature is build exactly to work around those internals.

Issue #365, opened 2018-02-05
Cannot connect to Google Cloud SQL instance using :socket_dir option


sausman:
I was trying to connect to a Cloud SQL instance using the `:socket_dir` option, but couldn't get it to work with [cloudsql-proxy](https://github.com/GoogleCloudPlatform/cloudsql-proxy).

cloudsql-proxy expects the filename to be the "Instance connection name" of the Cloud SQL instance you are trying to connect to. For example:

```
# For -fuse you do not need to specify instance names ahead of time:
./cloud_sql_proxy -dir=/cloudsql -fuse &
mysql -u root -S /cloudsql/my-project:us-central1:sql-inst
```

Postgrex [currently](https://github.com/elixir-ecto/postgrex/blob/5e787635097b393aa169fc0270d3f973a76b08c8/lib/postgrex/protocol.ex#L49) doesn't allow a filename to be configured for the socket.

```elixir
{{:local, "#{socket}/.s.PGSQL.#{port}"}, 0}
```

I was able to get it working with this [small change](https://github.com/sausman/postgrex/commit/dbf41a99f2e9e24364a893bdaa211166f7a41474) that allows the full path to be specified. I can work on a pull request, but wanted to get some feedback on the desired behavior.

Thanks!

josevalim:
@sausman can you please send a pull request where we also accept a `:socket_file` configuration too? So the plan is:

  1. Try socket_file
  2. Try socket_dir
  3. Fallback to hostname

We should document socket_file *after* socket_dir and make it clear that socket_dir is the preferred mechanism to configure sockets.

Thanks!

PR #366, opened 2018-02-07
Add :socket_file option.

sausman:
Fixes [365](#365), let me know if any changes are needed.

### DIFF
--- a/lib/postgrex.ex
+++ b/lib/postgrex.ex
@@ -34,7 +34,9 @@ defmodule Postgrex do
 
     * `:hostname` - Server hostname (default: PGHOST env variable, then localhost);
     * `:socket_dir` - Connect to Postgres via UNIX sockets in the given directory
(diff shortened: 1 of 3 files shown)


fishcakez:
From Ecto perspective it would be nice to get some consistency around configuration and `:socket_file` gets us out of sync with Mariaex, which uses `:socket` for this case in latest release. There is no `:socket_dir` convention with a port in the path name for MySQL. If `:socket` does not work in the Postgrex case, and I'm not sure it does, then we need to send a PR to Mariaex to maintain consistency.

There is the technicality that the socket may start with `<<0>>` and be abstract namespace, rather than a path. However I don't think anyone using that feature would be confused by the option naming.

josevalim:
@fishcakez Let's go with `:socket` then. I will merge this and update it.

josevalim:

PR #367, opened 2018-02-08
fix parsing of version string from latest pgdg repository build

danj3:
Version string for an official postgres repo build looks like this:

10.2 (Ubuntu 10.2-1.pgdg16.04+1)

which the version parser chokes on since it's only expecting something like "10.2".

### DIFF
--- a/lib/postgrex/utils.ex
+++ b/lib/postgrex/utils.ex
@@ -63,6 +63,8 @@ defmodule Postgrex.Utils do
   def parse_version(version) do
     list =
       version
+      |> String.split(" ")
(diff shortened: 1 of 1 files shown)


Nicd:
Just tested and this PR fixes postgrex exploding in my GitLab CI build where I use docker `postgres:latest`:

Breaking: https://gitlab.com/code-stats/code-stats/-/jobs/51800143

Fixed with PR version: https://gitlab.com/code-stats/code-stats/-/jobs/51800864

josevalim:

PR #371, opened 2018-02-23
Add test for Utils.parse_version/1

davich:
Test for this change: https://github.com/elixir-ecto/postgrex/pull/369


ericmj:
Thank you!

PR #381, opened 2018-04-12
Replace type manager by registry

### LINKED ISSUE #362: weird state in TypeManager
Postgrex version 0.13.3

I've encountered a weird state in `Postgrex.TypeManager` which I can't currently explain. I'm going to dump what I've gathered so far, so maybe someone else has an idea.

The manifestation of the problem was that a process trying to establish a connection with `sync_connect: true` would hang forever. Basically, `Postgrex.start_link` would never return. We had some processes stuck in this state for a couple of days.

After a bit of tracing, I've discovered the following stack trace in the connection process:

```elixir
{:gen, :do_call, 4, [file: 'gen.erl', line: 169]},
  {GenServer, :call, 3, [file: 'lib/gen_server.ex', line: 771]},
  {Postgrex.Protocol, :bootstrap, 3,
   [file: 'lib/postgrex/protocol.ex', line: 621]},
  {Postgrex.Protocol, :handshake, 2,
   [file: 'lib/postgrex/protocol.ex', line: 475]},
  {DBConnection.Connection, :connect, 2,
   [file: 'lib/db_connection/connection.ex', line: 134]},
  {Connection, :init_it, 6, [file: 'lib/connection.ex', line: 424]},
  {:proc_lib, :init_p_do_apply, 3, [file: 'proc_lib.erl', line: 247]}
```

Poking around the code a bit, I concluded that the process ends up running an infinite loop, due to [this condition](https://github.com/elixir-ecto/postgrex/blob/ad83bd78092a2ed2bc2e1190df22e79409638e40/lib/postgrex/protocol.ex#L628-L629). I've quickly traced the process and proved that it was indeed the case.

Digging deeper, I observed that [Postgrex.TypeManager.get](https://github.com/elixir-ecto/postgrex/blob/ad83bd78092a2ed2bc2e1190df22e79409638e40/lib/postgrex/protocol.ex#L621) returned a pid of the dead process. It was always returning the same pid.

Finally, I invoked `:sys.get_state` on `Postgrex.TypeManager` process, and saw that the pid of the dead process is indeed cached in the process state.

Looking at the code, I couldn't figure out how that happened. I tried reproducing it locally by inserting a code which asynchronously kills the type server after a brief delay, but type manager correctly handled that case.

The final thing I did was to manually send the monitor message to the type server, to indicate that the process has indeed died. This was properly detected by the type manager, and at this point, all the queries were unblocked.

Looking at the code, I currently don't have any explanation for this. Unfortunately, I also can't reproduce it anymore. Once I've manually simulated the monitor message, everything came back to normal.

Does anyone else have a clue here?

EDIT: I just tried the following on the same BEAM instance, and it appears that all cached type servers are dead, except for the one which I restarted by manually sending the monitor message:

```elixir

:sys.get_state(Postgrex.TypeManager) |> elem(0) |> Enum.map(fn {_, pid} -> pid end) |> Enum.reject(&Process.alive?/1)

[#PID<0.7534.164>, #PID<0.18393.164>, #PID<0.6709.162>, #PID<0.6742.162>,
 #PID<0.7630.164>, #PID<0.8526.163>, #PID<0.29855.163>, #PID<0.8600.163>,
 #PID<0.12130.162>, #PID<0.6669.162>, #PID<0.12062.162>, #PID<0.29869.163>,
 #PID<0.29820.163>, #PID<0.18391.164>, #PID<0.32756.170>, #PID<0.6731.162>,
 #PID<0.30772.162>, #PID<0.30767.162>]
```

josevalim:
The Registry is capable of handling requests currently,
takes less lines of code and also handles dead processes.

Note though we do a preemptive lookup on the registry
before starting the process to avoid spawning a process
that will die immediately after.

Closes #362


lib/postgrex/type_supervisor.ex
```diff
@@ -3,29 +3,57 @@ defmodule Postgrex.TypeSupervisor do
...
+  end
+
+  @doc """
+  Locates a type server for the given module-key pair.
+  """
+  def locate(module, key) do
+    pair = {module, key}
+
+    case Registry.lookup(@manager, pair) do
+      [{pid, _}] -> if Process.alive?(pid), do: pid, else: start_server(module, pair)
```
michalmuskala: Do we really gain anything by this check? It still does not give any guarantees, since the process might die right after this check.

josevalim: I am reproducing the logic of `whereis_name` which does check if the process is alive. The registry is not atomic regarding DOWN messages, so this is necessary.

michalmuskala: Since we already trap exits in the type server process, I wonder if it would make sense to try to "mark dying" the server in the registry from the terminate callback.

josevalim:
I would prefer to not couple the type server implementation withthe
registry. :)
-- 


lib/postgrex/type_supervisor.ex
```diff
@@ -3,29 +3,57 @@ defmodule Postgrex.TypeSupervisor do
...
+      {:ok, pid} -> pid
+      {:error, {:already_started, pid}} -> pid
+    end
   end
 
+  # Callbacks
+
   def init(:manager) do
-    manager    = worker(Postgrex.TypeManager, [])
+    manager = worker(Registry, [:unique, @manager])
```
sasa1977: Shouldn't this be `supervisor`?

lib/postgrex/type_supervisor.ex
sasa1977: I also think that `Process.alive?/1` is unnecessary here.

lib/postgrex/type_supervisor.ex
josevalim: Great catch! This is why I live Elixir's child_spec, haha.

josevalim:

Issue #383, opened 2018-05-06
Postgrex.CIDR/Postgrex.INET unification is a breaking change


halostatue:
The change in a948626 to remove `Postgrex.CIDR` and replace with `Postgrex.INET` is a breaking change for anyone who uses [adam12/ecto_network](https://github.com/adam12/ecto_network). While I have put a PR (adam12/ecto_network#11) in which makes `ecto_network` work with the Postgrex master branch, it is a breaking change for anyone who is using `Postgrex.CIDR` and many extant Postgrex version specifiers (the default version specifier in Phoenix is `{:postgrex, ">= 0.0.0"}`).

I don’t know the Postgrex internals, but I have no reason to believe that an approach similar to the one that I took with EctoNetwork would be inappropriate (keep `Postgrex.CIDR`, but have it essentially alias `Postgrex.INET` via `defdelegate` where possible).

I believe, based on the information that I’m seeing, that this is a *good* change, but I only found it after needing to switch to Postgrex master in order to properly test Ecto master for an issue.

halostatue:
I will also note that there is a further breaking change, because it used to be that `{127, 0, 0, 1}` would be returned as `%Postgrex.INET{address: {127, 0, 0, 1}}` (that is, `%Postgrex.INET{address: {127, 0, 0, 1}, netmask: nil}`), but is now returned as `%Postgrex.INET{address: {127, 0, 0, 1}, netmask: 32}`.

fishcakez:
Please see https://github.com/elixir-ecto/postgrex/issues/292. Postgrex master has other breaking changes compared to the previous release. It is still pre 1.0 unfortunately.

halostatue:
I think you misunderstand what I’m after. I’m not after reverting, because I think that the unification is a good thing for all of the reasons in #292, and I’ve already got a fix that will make it work with EctoNetwork as a PR.

There are two things that I’d like to see changed/fixed about this, both of which I can provide as PRs if you wish:

1. This change should be called out explicitly as something that will break previous use of Postgrex.CIDR or Postgrex.INET. When I looked over the changelog, this did not jump out as a potentially breaking change to me until I experienced it.

[...]

PR #384, opened 2018-05-06
Automatically encode/decode /32 or /128 netmasks

halostatue:
Pull request to fix an issue mentioned in #383.

### DIFF
--- a/lib/postgrex/extensions/inet.ex
+++ b/lib/postgrex/extensions/inet.ex
@@ -6,10 +6,15 @@ defmodule Postgrex.Extensions.INET do
 
   def encode(_) do
     quote location: :keep do
+      %Postgrex.INET{address: {a, b, c, d}, netmask: nil} ->
+        <<8 :: int32, 2, 32, 0, 4, a, b, c, d>>
       %Postgrex.INET{address: {a, b, c, d}, netmask: n} ->
-        <<8 :: int32, 2, n || 32, 0, 4, a, b, c, d>>
+        <<8 :: int32, 2, n, 1, 4, a, b, c, d>>
+      %Postgrex.INET{address: {a, b, c, d, e, f, g, h}, netmask: nil} ->
+        <<20 :: int32, 3, 128, 0, 16,
+          a::16, b::16, c::16, d::16, e::16, f::16, g::16, h::16>>
       %Postgrex.INET{address: {a, b, c, d, e, f, g, h}, netmask: n} ->
-        <<20 :: int32, 3, n || 128, 0, 16,
+        <<20 :: int32, 3, n, 1, 16,
           a::16, b::16, c::16, d::16, e::16, f::16, g::16, h::16>>
       other ->
         raise ArgumentError, Postgrex.Utils.encode_msg(other, Postgrex.INET)
@@ -18,12 +23,12 @@ defmodule Postgrex.Extensions.INET do
 
   def decode(_) do
     quote location: :keep do
-      <<8 :: int32, 2, n, _cidr?, 4, a, b, c, d>> ->
-        n = if(n == 32, do: nil, else: n)
+      <<8 :: int32, 2, n, cidr?, 4, a, b, c, d>> ->
+        n = if(cidr? == 1 or n != 32, do: n, else: nil)
         %Postgrex.INET{address: {a, b, c, d}, netmask: n}
-      <<20 :: int32, 3, n, _cidr?, 16,
+      <<20 :: int32, 3, n, cidr?, 16,
         a::16, b::16, c::16, d::16, e::16, f::16, g::16, h::16>> ->
-          n = if(n == 128, do: nil, else: n)
+        n = if(cidr? == 1 or n != 128, do: n, else: nil)
         %Postgrex.INET{address: {a, b, c, d, e, f, g, h}, netmask: n}
     end
   end
--- a/lib/postgrex/extensions/inet.ex
+++ b/lib/postgrex/extensions/inet.ex
@@ -7,9 +7,9 @@ defmodule Postgrex.Extensions.INET do
   def encode(_) do
     quote location: :keep do
       %Postgrex.INET{address: {a, b, c, d}, netmask: n} ->
-        <<8 :: int32, 2, n, 0, 4, a, b, c, d>>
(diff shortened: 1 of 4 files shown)


lib/postgrex/extensions/inet.ex
```diff
@@ -19,9 +19,11 @@ defmodule Postgrex.Extensions.INET do
   def decode(_) do
     quote location: :keep do
       <<8 :: int32, 2, n, _cidr?, 4, a, b, c, d>> ->
+        n = if(n == 32, do: nil, else: n)
         %Postgrex.INET{address: {a, b, c, d}, netmask: n}
       <<20 :: int32, 3, n, _cidr?, 16,
         a::16, b::16, c::16, d::16, e::16, f::16, g::16, h::16>> ->
+          n = if(n == 128, do: nil, else: n)
```
fishcakez: Can we always set the netmask as an integer on decode but handle `nil` on the encoding side (as is done already). Otherwise we lose the netmask information. Following the generous with what we allow as input but strict with what we have as output.

test/query_test.exs
```diff
@@ -610,16 +610,16 @@ defmodule QueryTest do
 
   @tag min_pg_version: "9.0"
   test "encode network types", context do
-    assert [[%Postgrex.INET{address: {127, 0, 0, 1}, netmask: 32}]] =
-           query("SELECT $1::inet", [%Postgrex.INET{address: {127, 0, 0, 1}, netmask: 32}])
-    assert [[%Postgrex.INET{address: {8193, 43981, 0, 0, 0, 0, 0, 0}, netmask: 128}]] =
-           query("SELECT $1::inet", [%Postgrex.INET{address: {8193, 43981, 0, 0, 0, 0, 0, 0}, netmask: 128}])
+    assert [[%Postgrex.INET{address: {127, 0, 0, 1}, netmask: nil}]] =
```
fishcakez: The tests in this function should output text instead of the structs - the reverse of the decode test. This ensure we only test encode or decode at one time to avoid hiding bugs on both sides. If want to do identity tests have that in a separate test. It is an existing issue with this test.
It should be like:
```elixir
assert query("SELECT $1::inet::text", [%Postgrex.INET{address: {127, 0, 0, 1}, netmask: nil}]) == [["127.0.0.1/32"]]
```

lib/postgrex/extensions/inet.ex
halostatue: I believe that the correct behaviour is what I have written here, and matches what is displayed from `psql`:

```sql
CREATE TABLE IF NOT EXISTS inet_example (id bigserial, address inet);
INSERT INTO inet_example (address) VALUES
('127.0.0.1'::inet), ('127.0.0.1/32'::inet), ('127.0.0.1/16'::inet),
('::1'::inet), ('::1/128'::inet), ('::1/16'::inet), ('::1/32'::inet);
SELECT address FROM inet_example;
```

The output is:
```
┌──────────────┐
│   address    │
├──────────────┤
│ 127.0.0.1    │
│ 127.0.0.1    │
│ 127.0.0.1/16 │
│ ::1          │
│ ::1          │
│ ::1/16       │
│ ::1/32       │
└──────────────┘
```

This is exactly the conversion that I am doing on decode, because it’s what most people would expect to be the case for an IP address. When you’re asked for your IP address, you typically answer `10.10.2.1`, not `10.10.2.1/32` (which a network admin will say is correct). From 8.9.1 inet in the PostgreSQL manual:

> The `inet` type holds an IPv4 or IPv6 host address, and optionally its subnet, all in one field. The subnet is represented by the number of network address bits present in the host address (the “netmask”). If the netmask is 32 and the address is IPv4, then the value does not indicate a subnet, only a single host. In IPv6, the address length is 128 bits, so 128 bits specify a unique host address. Note that if you want to accept only networks, you should use the `cidr` type rather than `inet`.
> The input format for this type is *`address/y`* where address is an IPv4 or IPv6 address and y is the number of bits in the netmask. If the *`/y`* portion is missing, the netmask is 32 for IPv4 and 128 for IPv6, so the value represents just a single host. On display, the *`/y`* portion is suppressed if the netmask specifies a single host.

This might be the distinction that I want to put for adam12/ecto_network#11 so that there’s a different interpretation in the human representation for Ecto (that is, it would always show `127.0.0.1/32` if the type is CIDR).

There’s an important *logical* distinction between INET and CIDR. `SELECT cidr '127.0.0.1/32';` is legal, but `SELECT cidr '127.0.0.1/24'` is not because CIDR does not allow nonzero bits to the right of the netmask, which can result in errors. There does appear to be automatic safe conversion because `SELECT cidr (inet '127.0.0.1/24')` returns the expected `127.0.0.0/24`, so I’m ultimately not that concerned about that logical distinction (and I tested by providing an `inet` to a column expecting a `cidr`).

test/query_test.exs
halostatue: I can rewrite these tests, but the only difference for these particular tests is to clear the netmask.

lib/postgrex/extensions/inet.ex
halostatue: I should point out, after reading your second comment, @fishcakez, that when I `SELECT address::text FROM inet_example`, this is the result:

```
┌──────────────┐
│   address    │
├──────────────┤
│ 127.0.0.1/32 │
│ 127.0.0.1/32 │
│ 127.0.0.1/16 │
│ ::1/128      │
│ ::1/128      │
│ ::1/16       │
│ ::1/32       │
└──────────────┘
```

The netmask is displayed. But this is not the default display from psql.

fishcakez:
> Should I address the ebert comments

Don't worry about those

lib/postgrex/extensions/inet.ex
halostatue: That sounds good. It will probably take a day or two to do this, including the test changes requested.

fishcakez: No rush, it's better we get this right.

josevalim:
@halostatue ping!

halostatue:
I haven’t forgotten; I have just gotten busy. I recall hitting a snag with this, but I don’t recall what it is offhand.

halostatue:
This should bring this up to speed. I have left my corrections as individual commits for now, but am happy to squash them prior to merge.

josevalim:

Issue #386, opened 2018-05-30
Postgrex.Notification connection silently dies


mmmries:
I ran into an issue similar to #359 in my production setup (`postgrex 0.13.5`, postgres 9.6). I had a node listening for notifications for 7 days. I noticed that I wasn't seeing updates elsewhere and tracked it back to a `Postgrex.Notification` process that was alive with a state of:

```
iex(spiff-ex@10.12.0.5)6> :sys.get_state(p2)
%Connection{
  backoff: nil,
  mod: Postgrex.Notifications,
  mod_state: %Postgrex.Notifications{
    listener_channels: %{
      "background_tasks" => %{
        #Reference<0.2314467320.2725511169.157557> => #PID<0.477.0>
      }
    },
    listeners: %{
      #Reference<0.2314467320.2725511169.157557> => {"background_tasks",
       #PID<0.477.0>}
    },
    parameters: nil,
    protocol: %Postgrex.Protocol{
      buffer: :active_once,
      connection_id: 19762,
      connection_key: 290296840,
      null: nil,
      parameters: #Reference<0.2314467320.2725511169.157556>,
      peer: {{10, 15, 241, 191}, 5432},
      postgres: :idle,
      queries: nil,
      sock: {:gen_tcp, #Port<0.12742>},
      timeout: 15000,
      transactions: :naive,
      types: nil
    }
  },
  raise: nil
}
```

I tested the process by trying to add an additional listener like this:

```
iex(spiff-ex@10.12.0.5)11> Postgrex.Notifications.listen(p2, "virtual")
** (exit) exited in: :gen_server.call(#PID<0.478.0>, {:listen, "virtual"}, 5000)
    ** (EXIT) %DBConnection.ConnectionError{message: "tcp async_recv: closed"}
    (stdlib) gen_server.erl:214: :gen_server.call/3
```

So it looks like the connection had died, but the `Postgrex.Notification` process had not been notified that the TCP connection was down, until it tried to use it again.

## Proposed Solution

I think it might be a good idea to do some sort of a PING message periodically. This could be managed by the `Postgrex.Notification` directly, or perhaps I could add a method that allows for this to be triggered elsewhere. Maybe a `Postgrex.Notification.check_alive(pid)` function?

I have some time to look into this and provide a PR or find a way of replicating if that is more helpful.

josevalim:
[...]

PR #388, opened 2018-05-30
implement an idle_timeout for Postgrex.Notification to ping the connection

mmmries:
This is a first attempt at addressing #386 

@josevalim mentioned that this is already implemented for normal postgrex connections and it looks to me like this is [handled as part of the DbConnection library](https://github.com/elixir-ecto/db_connection/blob/7588c321be4be4e740c7121dd1885569468228e9/lib/db_connection/connection.ex#L371-L379). So I tried to follow the same pattern here and use the [GenServer timeout feature](https://hexdocs.pm/elixir/1.6.5/GenServer.html#c:handle_call/3) to make sure that if our connection is sitting idle for too long we send a ping to the server.

In this case I've just matched on `{:ok, updated_protocol}` because if the ping fails it means something has gone wrong with our connection and I had no idea how to recover from that, so it seemed best to just crash.

This is my first PR to `postgrex`, so please don't hesitate to suggest changes. I'm very open to learning. And thank you for all of the hard work by the maintainers here ❤️ 💛 💚 💙 💜 ❤️ 💛 💚 💙 💜

### DIFF
--- a/lib/postgrex/notifications.ex
+++ b/lib/postgrex/notifications.ex
@@ -10,7 +10,11 @@ defmodule Postgrex.Notifications do
 
   @timeout 5000
 
-  defstruct protocol: nil, parameters: nil, listeners: Map.new(), listener_channels: Map.new()
+  defstruct idle_timeout: 5000,
+            protocol: nil,
+            parameters: nil,
+            listeners: Map.new(),
+            listener_channels: Map.new()
 
   ## PUBLIC API ##
 
@@ -88,7 +92,7 @@ defmodule Postgrex.Notifications do
   def init(opts) do
     if opts[:sync_connect] do
       case connect(:init, opts) do
-        {:ok, _} = ok -> ok
+        {:ok, _, _} = ok -> ok
         {:stop, reason, _} -> {:stop, reason}
       end
     else
@@ -99,7 +103,8 @@ defmodule Postgrex.Notifications do
   def connect(_, opts) do
     case Protocol.connect([types: nil] ++ opts) do
       {:ok, protocol} ->
-        {:ok, %__MODULE__{protocol: protocol}}
+        idle_timeout = Keyword.get(opts, :idle_timeout, 5000)
+        {:ok, %__MODULE__{idle_timeout: idle_timeout, protocol: protocol}, idle_timeout}
 
       {:error, reason} ->
         {:stop, reason, opts}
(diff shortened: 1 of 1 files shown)


mmmries:
The existing tests for `Postgrex.Notification` are making sure that my changes didn't break the existing functionality, but I didn't find a good way to write a test to make sure that we are actually pinging the connection. [DbConnection tests this](https://github.com/elixir-ecto/db_connection/blob/d0cbabaa0eacd67d3b4cdaa84e2c3082c500f8ac/integration_test/cases/idle_test.exs) by implementing a `TestAgent` protocol that captures the the ping attempt so they can check it, but I didn't see a good way to verify the behavior here. Any suggestions are totally welcome.

lib/postgrex/notifications.ex
```diff
@@ -99,7 +103,8 @@ defmodule Postgrex.Notifications do
   def connect(_, opts) do
     case Protocol.connect([types: nil] ++ opts) do
       {:ok, protocol} ->
-        {:ok, %__MODULE__{protocol: protocol}}
+        idle_timeout = Keyword.get(opts, :idle_timeout, 5000)
+        {:ok, %__MODULE__{idle_timeout: idle_timeout, protocol: protocol}} |> include_idle_timeout
```
josevalim: Since we are always including the idle timeout, I would then include it directly in the tuples and remove the indirection. So this would be `{:ok, ..., idle_timeout}`.

lib/postgrex/notifications.ex
```diff
@@ -115,60 +120,72 @@ defmodule Postgrex.Notifications do
     # If this is the first listener for the given channel,
     # we need to actually issue the LISTEN query.
     if Map.size(s.listener_channels[channel]) == 1 do
-      listener_query("LISTEN \"#{channel}\"", {:ok, ref}, from, s)
+      listener_query("LISTEN \"#{channel}\"", {:ok, ref}, from, s) |> include_idle_timeout
     else
-      {:reply, {:ok, ref}, s}
+      {:reply, {:ok, ref}, s} |> include_idle_timeout
```
josevalim: And this would be `{:reply, {:ok, ref}, s, s.idle_timeout}`... and so on. :)

lib/postgrex/notifications.ex
```diff
@@ -115,60 +120,72 @@ defmodule Postgrex.Notifications do
...
+          listener_query("UNLISTEN \"#{channel}\"", :ok, nil, s) |> include_idle_timeout
         else
-          {:noreply, s}
+          {:noreply, s} |> include_idle_timeout
         end
     end
   end
 
+  def handle_info(:timeout, %{protocol: protocol} = state) do
+    {:ok, protocol} = Protocol.ping(protocol)
```
josevalim: Regarding the ok/error, you should do exactly like in this case here: https://github.com/elixir-ecto/postgrex/pull/388/files#diff-121fce3140827f25d3c6efc58a82dac7R173

mmmries:
Thanks for the feedback, especially handling the protocol error to do an orderly shutdown with the right reason :+1:. That will make error messages a lot better for the user.

The build is failing now on something that looks like a flakey error since it didn't fail on any of the other versions or previous runs?

mmmries:
Thanks for all the help reviewing the issue and reviewing the PR. Keep up the great work!

PR #400, opened 2018-09-02
Return DBConnection.EncodeError and support repreparing queries

josevalim:
A test still needs to be added before merging.

### DIFF
--- a/lib/postgrex/protocol.ex
+++ b/lib/postgrex/protocol.ex
@@ -181,7 +181,15 @@ defmodule Postgrex.Protocol do
     lock_error(s, :prepare, query)
   end
 
-  def handle_prepare(%Query{types: nil, name: ""} = query, opts, s) do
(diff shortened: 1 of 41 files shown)


lib/postgrex/protocol.ex
```diff
@@ -181,7 +181,15 @@ defmodule Postgrex.Protocol do
     lock_error(s, :prepare, query)
   end
 
-  def handle_prepare(%Query{types: nil, name: ""} = query, opts, s) do
+  def handle_prepare(%Query{ref: ref} = query, opts, s) when is_reference(ref) do
+    # If the query already has a reference, then it means DBConnection rescued
+    # a DBConnection.EncodeError and wants us to reprepare a query
+    opts = Keyword.put(opts, :function, :prepare_execute)
```
fishcakez: This can also happen on declare so we need to check the `:function`, i.e. if its `:execute` do `:prepare_execute`, and if its `:declare` do `:prepare_declare`. This code likely works but in future it might not.

PR #403, opened 2018-09-13
Reduce the amount of metadata and structs

josevalim:
This is a series of commits to simplify a bit the implementation:

  1. Instead of passing :function metadata, we just track what is important, which is the :prepare flag
  2. Do not pass the stream as query, instead pass the query itself and use the stream option
  3. Remove the copy structs in favor of tuples

### DIFF
--- a/lib/postgrex.ex
+++ b/lib/postgrex.ex
@@ -121,8 +121,6 @@ defmodule Postgrex do
     * `:timeout` - Query request timeout (default: `#{@timeout}`);
     * `:decode_mapper` - Fun to map each row in the result to a term after
     decoding, (default: `fn x -> x end`);
-    * `:pool` - The pool module to use, must match that set on
-    `start_link/1`, see `DBConnection`
     * `:mode` - set to `:savepoint` to use a savepoint to rollback to before the
     query on error, otherwise set to `:transaction` (default: `:transaction`);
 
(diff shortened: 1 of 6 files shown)


lib/postgrex.ex
```diff
@@ -216,7 +208,7 @@ defmodule Postgrex do
     opts =
       opts
       |> defaults()
-      |> Keyword.put(:function, :prepare)
+      |> Keyword.put(:prepare, true)
```
josevalim: We only care if function is `:prepare`, so that's the information I am passing.

lib/postgrex/stream.ex
```diff
@@ -58,14 +49,15 @@ defimpl Collectable, for: Postgrex.Stream do
 
   def into(%Stream{conn: %DBConnection{}} = stream) do
     %Stream{conn: conn, query: query, params: params, options: opts} = stream
+    opts = Keyword.put(opts, :copy, true)
```
josevalim: Instead of passing the stream, we will pass the query and the copy flag above. This means less protocols to implement and less work around streams.

lib/postgrex/stream.ex
```diff
@@ -77,35 +69,15 @@ defimpl Collectable, for: Postgrex.Stream do
   defp make_into(conn, stream, %Postgrex.Copy{ref: ref} = copy, opts) do
     fn
       :ok, {:cont, data} ->
-        copy_data = %Postgrex.CopyData{ref: ref, data: data}
```
josevalim: I gave up on the structs, we can pass tuples instead.

josevalim:

PR #407, opened 2018-09-24
Add cache_statement to query

josevalim:


josevalim:
Need to get a test from `alter_test` and make sure the cache gets cleared.

josevalim:
@fishcakez I have added a commit that cleans up the cache and retries the query on feature_not_supported. You can see that I am doing this clean up on all queries, even on the prepared ones. If this is not advised, I can update the guard.

josevalim:

Issue #374, opened 2018-03-25
Add scram-sha-256 auth method


seriyps:
New authorization method was added to PostgreSQL 10.

Details and examples could be found here https://github.com/epgsql/epgsql/pull/156 and here https://github.com/epgsql/epgsql/issues/142

seriyps:
@josevalim sory, I don't know Elixir. I just implemented this auth method for epgsql and started to think that my implementation might be a good reference for postgrex. And to remind postgrex devs about such a feature.

josevalim:
Thanks @seriyps! Your reference implementation will definitely be useful. Let's wait until somebody needs this feature and then they can tackle it. :)

Neustradamus:
Linked to:
- https://github.com/scram-xmpp/info/issues/1

PR #410, opened 2018-09-29
Support scram-sha-256 authentication #374

nakagami:
Support scram-sha-256 authentication method new in PostgreSQL 10.

### DIFF
--- a/lib/postgrex/protocol.ex
+++ b/lib/postgrex/protocol.ex
@@ -11,6 +11,7 @@ defmodule Postgrex.Protocol do
   @max_packet 64 * 1024 * 1024
   @nonposix_errors [:closed, :timeout]
   @max_rows 500
+  @nonce_length 24
 
   defstruct sock: nil,
             connection_id: nil,
@@ -683,6 +684,15 @@ defmodule Postgrex.Protocol do
       {:ok, msg_auth(type: :md5, data: salt), buffer} ->
         auth_md5(s, status, salt, buffer)
 
+      {:ok, msg_auth(type: :sasl, data: _), buffer} ->
+        auth_sasl(s, status, buffer)
+
+      {:ok, msg_auth(type: :sasl_cont, data: data), buffer} ->
+        auth_cont(s, status, data, buffer)
+
+      {:ok, msg_auth(type: :sasl_fin, data: _), buffer} ->
+        auth_recv(s, status, buffer)
+
       {:ok, msg_error(fields: fields), buffer} ->
         disconnect(s, Postgrex.Error.exception(postgres: fields), buffer)
 
@@ -693,7 +703,7 @@ defmodule Postgrex.Protocol do
 
   defp auth_cleartext(s, %{opts: opts} = status, buffer) do
     pass = Keyword.fetch!(opts, :password)
-    auth_send(s, msg_password(pass: pass), status, buffer)
+    auth_send(s, msg_password(pass: [pass, 0]), status, buffer)
   end
 
   defp auth_md5(s, %{opts: opts} = status, salt, buffer) do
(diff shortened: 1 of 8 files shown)


mix.exs
```diff
@@ -32,7 +32,8 @@ defmodule Postgrex.Mixfile do
       {:jason, "~> 1.0", only: :test},
       {:decimal, "~> 1.0"},
       {:db_connection, "~> 2.0.0-dev", github: "elixir-ecto/db_connection", ref: "fbe3775"},
-      {:connection, "~> 1.0"}
+      {:connection, "~> 1.0"},
+      {:pbkdf2_elixir, "~> 0.12"}
```
josevalim: It should be an optional dependency.

lib/postgrex/protocol.ex
```diff
@@ -702,7 +712,43 @@ defmodule Postgrex.Protocol do
...
+        "0123456789abcdefghijklmnopqrstuvwxyzABCDEFGHIJKLMNOPQRSTUVWXYZ+/"
+        |> String.codepoints
+        |> Enum.random
+      end
+    )
+    message = "n,,n=,r=" <> nonce
+    auth_send(s, msg_password(pass: ["SCRAM-SHA-256", 0, << byte_size(message) :: signed-size(32) >>, message]), status, buffer)
+  end
+
+  defp auth_cont(s, %{opts: opts} = status, data, buffer) do
```
josevalim: Can you please rewrite this function with a couple things in mind?

1. Do not use the string functions but instead use pattern matching. For example, it is fine to do `String.split/2` but given all data has the same shape, it is preferred to do this:

```
for kv <- String.split(data, ",") do
  value_size = byte_size(kv) - 2
  <<key, ?=, value::size(value_size)>> = value_size
  {key, value}
end
```

2. Avoid conversion to atoms. Let's keep the codepoint as keys, for example: `server[?s]`

3. Do not use `Enum.join` or string concatenation. Instead, we can just work with IO.data. This means that `"c=biws,r=" <> server[:r]` becomes `["c=biws,r=", server[:r]]` and that `Enum.join(["r=", server[:r], ",s=", server[:s], ",i=", server[:i]], "")` becomes simply `["r=", server[:r], ",s=", server[:s], ",i=", server[:i]]`. If for some reason you need a string, you can call `IO.iodata_to_binary`.

4. Use the `Base` module instead of `:base64`

lib/postgrex/protocol.ex
```diff
@@ -702,7 +712,43 @@ defmodule Postgrex.Protocol do
...
+
+  defp auth_sasl(s, status = _, buffer) do
+    nonce = Enum.join(
+      for _ <- 1..@nonce_length do
+        "0123456789abcdefghijklmnopqrstuvwxyzABCDEFGHIJKLMNOPQRSTUVWXYZ+/"
+        |> String.codepoints
+        |> Enum.random
+      end
+    )
+    message = "n,,n=,r=" <> nonce
```
josevalim: We don't need to call `Enum.join` nor use binary concatenation, let's work with IO lists here.

lib/postgrex/protocol.ex
```diff
@@ -702,7 +712,43 @@ defmodule Postgrex.Protocol do
 
     digest = :crypto.hash(:md5, [pass, user]) |> Base.encode16(case: :lower)
     digest = :crypto.hash(:md5, [digest, salt]) |> Base.encode16(case: :lower)
-    auth_send(s, msg_password(pass: ["md5", digest]), status, buffer)
+    auth_send(s, msg_password(pass: ["md5", digest, 0]), status, buffer)
+  end
+
+  defp auth_sasl(s, status = _, buffer) do
+    nonce = Enum.join(
```
josevalim: The nonce can be computed more safely as `:crypto.strong_rand_bytes(18) |> Base.encode64`.

josevalim:
Thanks @nakagami for the PR! I have added some comments! :)

nakagami:
Thanks,
I'm Elixir newbie, so your advice are  very helpful for me.

Have you something another comments ?

josevalim:

lib/postgrex/protocol.ex
```diff
@@ -702,7 +712,41 @@ defmodule Postgrex.Protocol do
...
+        << k :: utf8, "=", v :: binary >> = kv
+        {k, v}
+      end
+    server = for {k, v} <- kv_list, into: %{}, do: {k, v}
+    client_nonce = String.slice(server[?r], 0, @nonce_length)
+
+    pass = Keyword.fetch!(opts, :password)
+    {server_i, ""} = Integer.parse(server[?i])
+    {:ok, server_s} = Base.decode64(server[?s])
+    salted_pass_hex = Pbkdf2.Base.hash_password(pass, server_s, [rounds: server_i, digest: :sha256, length: 32, format: :hex])
```
seriyps: I'm not sure how important is it, but PostgreSQL docs https://www.postgresql.org/docs/current/static/sasl-authentication.html#SASL-SCRAM-SHA-256 say that password should be "normalized" before hashing

> The SCRAM specification dictates that the password is also in UTF-8, and is processed with the SASLprep algorithm. PostgreSQL, however, does not require UTF-8 to be used for the password. When a user's password is set, it is processed with SASLprep as if it was in UTF-8, regardless of the actual encoding used. However, if it is not a legal UTF-8 byte sequence, or it contains UTF-8 byte sequences that are prohibited by the SASLprep algorithm, the raw password will be used without SASLprep processing, instead of throwing an error. This allows the password to be normalized when it is in UTF-8, but still allows a non-UTF-8 password to be used, and doesn't require the system to know which encoding the password is in.

In epgsql I decided to not implement all this complex normalization and only explicitly disallow non-ASCII in passwords: https://github.com/epgsql/epgsql/pull/156/files#diff-4a054884b454458eb3b45acded1fee2bR102

Neustradamus:
@nakagami and others: Thanks a lot for all about SCRAM.

Linked to:
- https://github.com/scram-xmpp/info/issues/1

PR #414, opened 2018-10-09
Disconnect on read-only transaction error

wojtekmach:
Similar to https://github.com/xerions/mariaex/issues/180

Example: script below tries to constantly write and at some point we cause manual failover and expect the system to self-heal:

```elixir
# lib/aurora.ex
defmodule Aurora.Application do
  use Application

  def start(_, _) do
    opts = [
      # ...
      name: :aurora,
      disconnect_on_read_only_transaction_error: true
    ]

    children = [
      {Postgrex, opts}
    ]

    Supervisor.start_link(children, strategy: :one_for_one, name: Aurora.Supervisor)
  end
end
```

```elixir
# aurora.exs
Enum.each(1..1000, fn i ->
  :timer.sleep(1000)
  IO.write(to_string(DateTime.utc_now()))

  statement = "INSERT INTO integers (x) VALUES ($1)"


  case Postgrex.query(:aurora, statement, [i]) do
    {:ok, _result} ->
      IO.puts(" (ok)")


    {:error, %Postgrex.Error{postgres: %{code: :read_only_sql_transaction}} = exception} ->
      IO.puts(" (error)")
      IO.inspect(exception)

    {:error, exception} ->
      IO.puts(" (error)")
      IO.inspect(exception)
  end
end)
```

```
2018-10-09 15:32:39.785912Z (ok)
2018-10-09 15:32:41.017704Z (ok)
2018-10-09 15:32:42.235790Z (ok)
2018-10-09 15:32:43.443853Z (ok)
2018-10-09 15:32:44.657805Z (error)

17:32:46.545 [error] Postgrex.Protocol (#PID<0.198.0>) disconnected: ** (DBConnection.ConnectionError) tcp recv: closed
%DBConnection.ConnectionError{message: "tcp recv: closed"}

17:32:46.656 [error] Postgrex.Protocol #PID<0.198.0> could not cancel backend: tcp connect: connection refused - :econnrefused

17:32:47.157 [error] Postgrex.Protocol (#PID<0.198.0>) failed to connect: ** (DBConnection.ConnectionError) tcp connect (aurorabug-cluster.cluster-c6tmwgqa951y.eu-central-1.rds.amazonaws.com:5432): connection refused - :econnrefused
2018-10-09 15:32:47.546801Z (error)
%DBConnection.ConnectionError{
  message: "connection not available and request was dropped from queue after 2775ms"
}
2018-10-09 15:32:51.323003Z (error)
%DBConnection.ConnectionError{
  message: "connection not available and request was dropped from queue after 1999ms"
}
2018-10-09 15:32:54.322926Z (error)
%DBConnection.ConnectionError{
  message: "connection not available and request was dropped from queue after 1999ms"
}
2018-10-09 15:32:57.322953Z (error)
%DBConnection.ConnectionError{
  message: "connection not available and request was dropped from queue after 1999ms"
}

17:32:59.954 [error] Postgrex.Protocol (#PID<0.198.0>) failed to connect: ** (Postgrex.Error) FATAL 57P03 (cannot_connect_now) the database system is starting up
2018-10-09 15:33:00.322945Z (error)
%DBConnection.ConnectionError{
  message: "connection not available and request was dropped from queue after 1999ms"
}
2018-10-09 15:33:03.322770Z (ok)
2018-10-09 15:33:06.088702Z (ok)
2018-10-09 15:33:07.317767Z (ok)
2018-10-09 15:33:08.535858Z (ok)
```

### DIFF
--- a/lib/postgrex/protocol.ex
+++ b/lib/postgrex/protocol.ex
@@ -1694,6 +1694,8 @@ defmodule Postgrex.Protocol do
     end
   end
 
+  defp maybe_disconnect({:error, _, %{disconnect_on_error_codes: []}} = result), do: result
+
   defp maybe_disconnect({:error,
          %Postgrex.Error{postgres: %{code: code}} = error,
          %{disconnect_on_error_codes: codes} = state
--- a/lib/postgrex/protocol.ex
+++ b/lib/postgrex/protocol.ex
@@ -1610,6 +1610,7 @@ defmodule Postgrex.Protocol do
     else
       {:error, %Postgrex.Error{} = err, s, buffer} ->
         error_ready(s, status, err, buffer)
+        |> maybe_disconnect()
 
       {:disconnect, _err, _s} = disconnect ->
         disconnect
@@ -1646,6 +1647,7 @@ defmodule Postgrex.Protocol do
       {:error, %Postgrex.Error{} = err, s, buffer} ->
         query_delete(s, err, query)
         error_ready(s, status, err, buffer)
+        |> maybe_disconnect()
 
       {:disconnect, _err, _s} = disconnect ->
         disconnect
@@ -1680,6 +1682,14 @@ defmodule Postgrex.Protocol do
 
       {:disconnect, _err, _s} = disconnect ->
         disconnect
+    end
+  end
+
+  defp maybe_disconnect({:error, %Postgrex.Error{postgres: %{code: code}} = error, state}) do
+    if code == :read_only_sql_transaction do
+      {:disconnect, error, state}
+    else
+      {:error, error, state}
     end
   end
 
--- a/lib/postgrex/protocol.ex
+++ b/lib/postgrex/protocol.ex
@@ -24,7 +24,7 @@ defmodule Postgrex.Protocol do
             postgres: :idle,
             transactions: :strict,
             buffer: nil,
-            disconnect_on_read_only_transaction_error: false
+            disconnect_on_errors: []
 
   @type state :: %__MODULE__{
           sock: {module, any},
@@ -39,7 +39,7 @@ defmodule Postgrex.Protocol do
           postgres: DBConnection.status() | {DBConnection.status(), reference},
           transactions: :strict | :naive,
           buffer: nil | binary | :active_once,
-          disconnect_on_read_only_transaction_error: boolean
+          disconnect_on_errors: [atom()]
         }
 
   @type notify :: (binary, binary -> any)
@@ -66,7 +66,7 @@ defmodule Postgrex.Protocol do
     sock_opts = [send_timeout: timeout] ++ (opts[:socket_options] || [])
     ssl? = opts[:ssl] || false
     types_mod = Keyword.fetch!(opts, :types)
-    disconnect_on_read_only_transaction_error = opts[:disconnect_on_read_only_transaction_error] || false
+    disconnect_on_errors = opts[:disconnect_on_errors] || []
 
     transactions =
       case opts[:transactions] || :naive do
@@ -84,7 +84,7 @@ defmodule Postgrex.Protocol do
       timeout: timeout,
       postgres: :idle,
       transactions: transactions,
-      disconnect_on_read_only_transaction_error: disconnect_on_read_only_transaction_error
+      disconnect_on_errors: disconnect_on_errors
     }
 
     types_key = if types_mod, do: {host, port, Keyword.fetch!(opts, :database)}
@@ -1694,12 +1694,15 @@ defmodule Postgrex.Protocol do
     end
   end
 
-  defp maybe_disconnect({
-         :error,
-         %Postgrex.Error{postgres: %{code: :read_only_sql_transaction}} = error,
-         %{disconnect_on_read_only_transaction_error: true} = state
-       }) do
-    {:disconnect, error, state}
+  defp maybe_disconnect({:error,
+         %Postgrex.Error{postgres: %{code: code}} = error,
+         %{disconnect_on_errors: codes} = state
+       } = result) do
+    if code in codes do
+      {:disconnect, error, state}
+    else
+      result
+    end
   end
 
   defp maybe_disconnect(other), do: other
--- a/lib/postgrex/protocol.ex
+++ b/lib/postgrex/protocol.ex
@@ -1694,13 +1694,15 @@ defmodule Postgrex.Protocol do
     end
   end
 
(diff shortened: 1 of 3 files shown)


lib/postgrex/protocol.ex
```diff
@@ -23,7 +23,8 @@ defmodule Postgrex.Protocol do
             queries: nil,
             postgres: :idle,
             transactions: :strict,
-            buffer: nil
+            buffer: nil,
+            disconnect_on_read_only_transaction_error: false
```
wojtekmach: I'm not a fan either of this option name or putting it on the state but can't think of anything better at the moment.

lib/postgrex/protocol.ex
```diff
@@ -1683,6 +1694,14 @@ defmodule Postgrex.Protocol do
     end
   end
 
+  defp maybe_disconnect({:error, %Postgrex.Error{postgres: %{code: code}} = error, state}) do
+    if code == :read_only_sql_transaction and state.disconnect_on_read_only_transaction_error do
```
josevalim: Maybe we should fully pattern match here?

```elixir
defp maybe_disconnect({
       :error,
       %Postgrex.Error{postgres: %{code: :read_only_sql_transaction}} = error,
       %{disconnect_on_read_only_transaction_error: true} = state
     }) do
  {:disconnect, error, state}
end

defp maybe_disconnect(other), do: other
```

test/transaction_test.exs
```diff
@@ -98,6 +99,44 @@ defmodule TransactionTest do
...
+      P.query!(conn, "SET TRANSACTION READ ONLY", []).connection_id
+
+      {:ok, query} = P.prepare(conn, "query_1", "insert into uniques values (1);", [])
+
+      assert capture_log(fn ->
+        {:error, %Postgrex.Error{postgres: %{message: "cannot execute INSERT in a read-only transaction"}}} =
+          P.execute(conn, query, [])
+
+        pid = context[:pid]
+        assert_receive {:EXIT, ^pid, :killed}
```
josevalim: @fishcakez why killed? Are those exists going to count towards the supervisor max_restarts? If so, I can see this being a problem.

josevalim: To clarify, I thought that only stops counted as exits, and not disconnects.

lib/postgrex/protocol.ex
josevalim: I think both of these are fine! :+1:

test/transaction_test.exs
```diff
@@ -98,6 +99,44 @@ defmodule TransactionTest do
...
+  test "transaction read-only only error disconnects with prepare and execute", context do
+    Process.flag(:trap_exit, true)
+
+    assert transaction(fn conn  ->
+      P.query!(conn, "SET TRANSACTION READ ONLY", []).connection_id
+
+      {:ok, query} = P.prepare(conn, "query_1", "insert into uniques values (1);", [])
+
+      assert capture_log(fn ->
+        {:error, %Postgrex.Error{postgres: %{message: "cannot execute INSERT in a read-only transaction"}}} =
```
josevalim: @wojtekmach we should match on the code rather than on the message because the message can be internationalized. This will also fix the build on 8.4.

test/transaction_test.exs
```diff
@@ -98,6 +99,44 @@ defmodule TransactionTest do
...
+          P.execute(conn, query, [])
+
+        pid = context[:pid]
+        assert_receive {:EXIT, ^pid, :killed}
+      end) =~ "disconnected: ** (Postgrex.Error) ERROR 25006 (read_only_sql_transaction) cannot execute INSERT"
```
josevalim: We should assert only up to "disconnected: ** (Postgrex.Error) ERROR 25006 (read_only_sql_transaction)" as the rest can be I18ned.

lib/postgrex/protocol.ex
josevalim: @fishcakez I like making it a bit more general. :+1: it should also address @wojtekmach's concern with naming. :)

wojtekmach: sounds great, how about `:disconnect_on_errors`?

josevalim: @wojtekmach excellent.

lib/postgrex/protocol.ex
```diff
@@ -1683,6 +1694,19 @@ defmodule Postgrex.Protocol do
     end
   end
 
+  defp maybe_disconnect({:error,
+         %Postgrex.Error{postgres: %{code: code}} = error,
+         %{disconnect_on_error_codes: codes} = state
```
josevalim: For roflscale reasons, let's have a clause before this one like this:

```elixir
    defp maybe_disconnect({:error, _, %{disconnect_on_error_codes: []}} = result), do: result
```

because that's the most common scenario after all.

josevalim:
One last comment and this looks good to me!

lib/postgrex/protocol.ex
wojtekmach: > For roflscale reasons

techempower here we go! oh wait, they don't do fault tolerance now do they?

josevalim:

lib/postgrex/protocol.ex
```diff
@@ -1610,6 +1619,7 @@ defmodule Postgrex.Protocol do
     else
       {:error, %Postgrex.Error{} = err, s, buffer} ->
         error_ready(s, status, err, buffer)
+        |> maybe_disconnect()
```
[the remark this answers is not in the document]
wojtekmach: I tried writing a similar test using mode: :savepoint and I couldn't see a failure so maybe it's fine to keep it as is.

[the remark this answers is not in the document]
josevalim: Ah, sounds good then!

archSeer:
Would it be possible to port a similar fix to mariaex? We're still getting bit by xerions/mariaex#180

fishcakez:
@archSeer sure you could do exactly the same, though I think it would have to be integer error codes there.

wojtekmach:
Spoken with @studzien about this yesterday; one problem with this solution is that it's retroactive, we need to run into a failure in order to disconnect as opposed to be proactive. Depending on DB adapter we may know whether we're in read-only transaction (e.g. MySQL server can return a SERVER_STATUS_IN_TRANS_READONLY status flag)

I think the current solution is probably good enough for most cases, but just mentioning this for posterity.

PR #429, opened 2018-11-23
Fix runtime warnings when running against decimal >= 1.6

lukaszsamson:
Fixes
```
warning: passing float to Decimal.new/1 is deprecated, use Decimal.from_float/1 instead
  (decimal) lib/decimal.ex:1063: Decimal.new/1
  (postgrex) lib/postgrex/type_module.ex:713: Postgrex.DefaultTypes.encode_params/3
  (postgrex) lib/postgrex/query.ex:62: DBConnection.Query.Postgrex.Query.encode/3
  (db_connection) lib/db_connection.ex:1074: DBConnection.encode/5
  (db_connection) lib/db_connection.ex:1172: DBConnection.run_prepare_execute/5
  (db_connection) lib/db_connection.ex:1268: DBConnection.run/6
  (db_connection) lib/db_connection.ex:480: DBConnection.parsed_prepare_execute/5
  (db_connection) lib/db_connection.ex:473: DBConnection.prepare_execute/4
  (ecto_sql) lib/ecto/adapters/sql.ex:536: Ecto.Adapters.SQL.execute!/4
  (ecto_sql) lib/ecto/adapters/sql.ex:528: Ecto.Adapters.SQL.execute/5
  (ecto) lib/ecto/repo/queryable.ex:147: Ecto.Repo.Queryable.execute/4
  (ecto) lib/ecto/repo/queryable.ex:18: Ecto.Repo.Queryable.all/3
```

### DIFF
--- a/lib/postgrex/extensions/numeric.ex
+++ b/lib/postgrex/extensions/numeric.ex
@@ -11,7 +11,7 @@ defmodule Postgrex.Extensions.Numeric do
       n when is_float(n) ->
         data = unquote(__MODULE__).encode_numeric(Decimal.from_float(n))
         [<<IO.iodata_length(data)::int32>> | data]
-      n when is_number(n) ->
+      n when is_integer(n) ->
         data = unquote(__MODULE__).encode_numeric(Decimal.new(n))
         [<<IO.iodata_length(data)::int32>> | data]
     end


lukaszsamson:
In postgrex IMO it's not necessary as it depends on > 1. In ecto - yes as it uses `Decimal.to_float/1` which is broken in < 1.6
https://github.com/ericmj/decimal/pull/83

wojtekmach:
We need to bump the dependency on postgrex because your patch won't compile on Decimal versions before 1.5. Bumping dependency on Ecto isn't strictly necessary (code will compile) but due to that bug it's a good idea to force a good version. Thanks!

lib/postgrex/extensions/numeric.ex
```diff
@@ -8,6 +8,9 @@ defmodule Postgrex.Extensions.Numeric do
       %Decimal{} = decimal ->
         data = unquote(__MODULE__).encode_numeric(decimal)
         [<<IO.iodata_length(data)::int32>> | data]
+      n when is_float(n) ->
+        data = unquote(__MODULE__).encode_numeric(Decimal.from_float(n))
+        [<<IO.iodata_length(data)::int32>> | data]
       n when is_number(n) ->
```
michalmuskala: Let's change this check now to `is_integer`. `is_number` is basically `is_integer + is_float` and we covered the first part already.

PR #430, opened 2018-11-27
DBConnection replaced :idle_timeout with :idle_interval

wojtekmach:
Note we still have `:idle_timeout` option in the codebase for `Postgrex.Notifications` which is unrelated to the DBConnection one: https://github.com/elixir-ecto/postgrex/blob/master/lib/postgrex/notifications.ex#L13


josevalim:

PR #432, opened 2018-12-28
Strip "devel" from Postgres version before processing numbers

jflatow:


lib/postgrex/utils.ex
```diff
@@ -63,6 +63,7 @@ defmodule Postgrex.Utils do
     [version_string | _] = String.split(version, " ", parts: 2)
     segments =
       version_string
+      |> String.trim("devel")
```
josevalim: Should this be `trim_trailing`?

jflatow: Yes that makes sense, updated.

josevalim:

PR #444, opened 2019-03-07
Use application inference

wojtekmach:
We depend on Elixir ~> 1.4 so it's safe


josevalim:

PR #450, opened 2019-03-14
Replace :crypto.hash/2 with :erlang.md5/1 for authentication

### LINKED ISSUE #449: Use :erlang.md5/1 instead of :crypto.hash/2
Hi, we are using erlang and host OS in FIPS mode, talking to PG on aws. Problem we are facing with is that PG on aws requests authentication encrypted with md5 using given salt, and that can't be changed. That procedure is FIPS 140-2 compliant since communication is TLS encrypted. Problem is that in `Postgrex.Protocol.auth_md5/4` :crypto.hash(:md5, iodata) is used and openssl in fips mode refuses to do md5 encryption:

```
[error] GenServer #PID<0.269.0> terminating
** (RuntimeError) Connect raised a ErlangError error. The exception details are hidden, as
they may contain sensitive data such as database credentials.

    (crypto) crypto.erl:949: :crypto.notsup_to_error/1
    (postgrex) lib/postgrex/protocol.ex:612: Postgrex.Protocol.auth_md5/4
    (postgrex) lib/postgrex/protocol.ex:504: Postgrex.Protocol.handshake/2
    (db_connection) lib/db_connection/connection.ex:135: DBConnection.Connection.connect/2
    (stdlib) proc_lib.erl:247: :proc_lib.init_p_do_apply/3
Last message: nil
```

Erlang docs for FIPS mode state that "For exceptional, non-cryptographic use cases one may consider switching to erlang:md5/1 as well." and indeed it works as expected. Can you consider this change?

Thanks

burmajam:
When used in FIPS mode, :crypto.hash(:md5, iodata) refuses to do encryption. Authentication using md5 hashed password [is FIPS 140-2 compliant if connection is TLS encrypted](https://docs.oracle.com/cd/E36784_01/html/E54953/fips-notok-1.html). This PR changes use of :crypto.hash/2 function with :erlang.md5/1 as suggested by [erlang documentation](http://erlang.org/doc/apps/crypto/fips.html).

Closes #449

### DIFF
--- a/lib/postgrex/protocol.ex
+++ b/lib/postgrex/protocol.ex
@@ -718,8 +718,8 @@ defmodule Postgrex.Protocol do
     user = Keyword.fetch!(opts, :username)
     pass = Keyword.fetch!(opts, :password)
 
-    digest = :crypto.hash(:md5, [pass, user]) |> Base.encode16(case: :lower)
-    digest = :crypto.hash(:md5, [digest, salt]) |> Base.encode16(case: :lower)
+    digest = :erlang.md5([pass, user]) |> Base.encode16(case: :lower)
+    digest = :erlang.md5([digest, salt]) |> Base.encode16(case: :lower)
     auth_send(s, msg_password(pass: ["md5", digest, 0]), status, buffer)
   end
 


josevalim:

PR #459, opened 2019-05-02
Update close!/3 to match spec and close/3

### LINKED ISSUE #458: Postgrex.close! returns Result struct though spec and docs declare :ok
Just that ^

wojtekmach:
Closes #458

### DIFF
--- a/lib/postgrex.ex
+++ b/lib/postgrex.ex
@@ -382,11 +382,8 @@ defmodule Postgrex do
   """
   @spec close(conn, Postgrex.Query.t, [option]) :: :ok | {:error, Exception.t}
   def close(conn, query, opts \\ []) do
-    case DBConnection.close(conn, query, opts) do
-      {:ok, _} ->
-        :ok
-      {:error, _} = error ->
-        error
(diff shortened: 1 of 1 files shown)


wojtekmach:
This patch is mostly for consistency (with spec and close/3), but can't help but wonder if we should return result struct?

it doesn't have much useful information but it's still more than just ok atom, this is what we had before:

```elixir
{:ok, pid} = Postgrex.start_link(database: "wojtek"); query = Postgrex.prepare!(pid, "SELECT now()", []); Postgrex.close!(pid, query)
%Postgrex.Result{
  columns: nil,
  command: :close,
  connection_id: 19913,
  messages: nil,
  num_rows: nil,
  rows: nil
}
```

fwiw, I guess we could always grab the result struct from LogEntry.

josevalim:
I think the consistency here would be to actually return the result. The bang variants most often return the `value` part from `{:ok, value}`.

wojtekmach:
Closing in favour of #461

josevalim:

PR #460, opened 2019-05-03
Add explicit cast for CockroachDB v19.1

tlvenn:
The latest version of CockroachDB comes with full support for correlated subqueries which was an issue before. Now the only stoper is one sql fragment where an explicity cast is currently needed.

See: https://github.com/cockroachdb/cockroach/issues/14554


tlvenn:
Hi @fishcakez, would that tiny change be acceptable ? I don't think the explicit cast should hinder performance but I might be wrong.

josevalim:

tlvenn:
Thanks guys, is there any chance to release a v0.14.3 with this PR so that people can start using the official driver from hex to talk to CDB ?

Thanks in advance.

PR #461, opened 2019-05-03
Return result from Postgrex.close/3

### LINKED ISSUE #458: Postgrex.close! returns Result struct though spec and docs declare :ok

wojtekmach:
Closes #458

### DIFF
--- a/lib/postgrex.ex
+++ b/lib/postgrex.ex
@@ -344,9 +344,7 @@ defmodule Postgrex do
   """
   @spec execute(conn, Postgrex.Query.t, list, [execute_option]) ::
     {:ok, Postgrex.Query.t, Postgrex.Result.t} | {:error, Postgrex.Error.t}
-  def execute(conn, query, params, opts \\ []) do
-    DBConnection.execute(conn, query, params, opts)
-  end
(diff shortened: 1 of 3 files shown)


josevalim:
Doh, just now I saw that the previous version was returning `:ok` for close and not `{:ok, result}`. The previous PR is correct then? Which one do you prefer? Which approach did you take in MyXQL?

wojtekmach:
@josevalim yeah that's what I meant by trying to make it consistent :) In MyXQL I followed Postgrex (so does Mariaex), so `close` returns `:ok`. There's no `close!/3` in Mariaex/MyXQL.

I opened the previous PR because it seemed like a smaller change (e.g. people would use non-bang version more often so we wouldn't break that contract). I don't have a strong opinion which way it should be. I guess I'm leaning a little bit towards returning result struct as it returns _more_ data (even though there's not a lot of useful data there, I guess connection id could be useful?).

Again, not a strong opinion either way, in either case we need to make a breaking change so if there're no other arguments perhaps we pick the one which is less disruptive? FWIW ecto_sql doesn't use this function.

PR #465, opened 2019-06-12
Fixes #464.

### LINKED ISSUE #464: Unable to use beta version of postgres
This issue stems from this [SO](https://stackoverflow.com/questions/56543947/can-connect-to-postgres-container-through-shell-but-not-through-phoenix-app) question (not mine). It seems that the current beta version of postgres reports it's version as `12beta1`. The `Postgrex.Utils.parse_version/1` function crashes with this version.

am-kantox:
The implementation considers `"alpha" < "beta" < 0`. To achieve this, we convert the value to charlist, take the head and make it negative :)

### DIFF
--- a/lib/postgrex/utils.ex
+++ b/lib/postgrex/utils.ex
@@ -63,9 +63,16 @@ defmodule Postgrex.Utils do
     [version_string | _] = String.split(version, " ", parts: 2)
     segments =
       version_string
-      |> String.trim_trailing("devel")
-      |> String.split(".", parts: 4)
-      |> Enum.map(&String.to_integer/1)
+      |> String.split(" ")
+      |> hd()
+      |> String.split(~r/\.|(?<=\d)(?=[^\d.])|(?<=[^\d.])(?=\d)/, parts: 4)
+      |> Enum.map(fn number_or_name ->
+        # we want "alpha" < beta" < 0
+        case Integer.parse(number_or_name) do
+          :error -> number_or_name |> String.to_charlist() |> hd() |> Kernel.-(255)
+          {i, ""} -> i
+        end
+      end)
 
     case segments do
       [major, minor, patch, _] -> {major, minor, patch}
--- a/lib/postgrex/utils.ex
+++ b/lib/postgrex/utils.ex
@@ -63,16 +63,8 @@ defmodule Postgrex.Utils do
     [version_string | _] = String.split(version, " ", parts: 2)
     segments =
       version_string
-      |> String.split(" ")
-      |> hd()
-      |> String.split(~r/\.|(?<=\d)(?=[^\d.])|(?<=[^\d.])(?=\d)/, parts: 4)
-      |> Enum.map(fn number_or_name ->
-        # we want "alpha" < "beta" < 0
-        case Integer.parse(number_or_name) do
-          :error -> number_or_name |> String.to_charlist() |> hd() |> Kernel.-(255)
-          {i, ""} -> i
-        end
-      end)
+      |> do_parse()
+      |> Enum.reverse()
 
     case segments do
       [major, minor, patch, _] -> {major, minor, patch}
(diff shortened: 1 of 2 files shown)


lib/postgrex/utils.ex
```diff
@@ -63,9 +63,16 @@ defmodule Postgrex.Utils do
     [version_string | _] = String.split(version, " ", parts: 2)
     segments =
       version_string
-      |> String.trim_trailing("devel")
-      |> String.split(".", parts: 4)
-      |> Enum.map(&String.to_integer/1)
+      |> String.split(" ")
+      |> hd()
+      |> String.split(~r/\.|(?<=\d)(?=[^\d.])|(?<=[^\d.])(?=\d)/, parts: 4)
```
am-kantox: FWIW, this regex matches:

- dots
- boundaries between digits and neither digits nor dots
- boundaries between neither digits nor dots and digits

It preserves the existing behaviour but also splits on boundaries.

josevalim: Thanks! I would prefer to not have to rely on a regex though. It will be quite hard to maintain and update it in the future. Maybe we can achieve this with "regular" code?

am-kantox: You brought `Regex` to _Elixir_, not me :)

Pick one, please:

```elixir
  defp do_parse(s, acc) do
    case Integer.parse(s) do
      {i, ""} ->
        [i | acc]

      {i, " " <> _} ->
        [i | acc]

      {i, "." <> rest} ->
        do_parse(rest, [i | acc])

      {i, <<c::binary-size(1), rest::binary()>>} ->
        [ver] = String.to_charlist(c)
        do_parse(rest, [ver - 255, i | acc])

      :error ->
        with <<_::binary-size(1), rest::binary()>> <- s, do: do_parse(rest, acc)
    end
  end
```
or
```elixir
  def do_parse(s) do
    s
    |> String.codepoints()
    |> Enum.reduce_while({false, []}, fn
      d, {false, acc} when d in ~w|0 1 2 3 4 5 6 7 8 9 0| ->
        {:cont, {true, [d | acc]}}

      d, {true, [h | acc]} when d in ~w|0 1 2 3 4 5 6 7 8 9 0| ->
        {:cont, {true, [h <> d | acc]}}

      ".", {true, acc} ->
        {:cont, {false, acc}}

      " ", acc ->
        {:halt, acc}

      any, {true, acc} ->
        {:cont,
         {false, [any |> String.to_charlist() |> hd() |> Kernel.-(255) |> to_string() | acc]}}

      _, acc ->
        {:cont, acc}
    end)
    |> elem(1)
    |> Enum.map(&String.to_integer/1)
  end
```
called as 
```elixir
    segments =
      version_string
      |> do_parse()
      |> Enum.reverse()
```

Am I missing some obvious easy solution?

josevalim:
My take:

```elixir
def parse_version(version) do
    segments =
      version
      |> String.split(" ", parts: 2)
      |> hd()
      |> String.split(".", parts: 4)
      |> Enum.map(&parse_version_bit/1)

    case segments do
      [major, minor, patch, _] -> {major, minor, patch}
      [major, minor, patch] -> {major, minor, patch}
      [major, minor] -> {major, minor, 0}
      [major] -> {major, 0, 0}
    end
  end

  defp parse_version_bit(bit) do
    {int, _} = Integer.parse(bit)
    int
  end
```

I don't think we should return `{12, -negative, ...}`. This is not meant to be an accurate parsing of the version. We mostly use it for feature capability checks.

am-kantox:
Unfortunately `"beta" < 0 #⇒ false`. If the version is supposed to be compared later for the feature capability, `"12beta1"` would be treated as the topmost in `"12"` major branch.
 `

josevalim:
We don't expose this feature in any way and this library doesn't care about this information, so there is no need for this complexity. :) If we eventually need to distinguish between betas and regular, then we can add it, but right now we don't.

am-kantox:
Got it. For the sake of easy merge, I pushed your code, so if you want you might just merge this.

josevalim:

PR #466, opened 2019-06-16
Raise exception if encoding a date that is not in Calendar.ISO

kipcole9:
Postgrex currently assumes that a date to be encoded is in
the Calendar.ISO calendar (the default). However Elixir
supports different calendar types in the %Date{} struct.

Postgres does not support dates outside the Calendar.ISO
calendar for the Postgres standard date type.

This commit raises the standard DBConnection.EncodeError
error if a date in a different calendar is provided.


josevalim:

PR #471, opened 2019-07-17
Fix doc: PostgreSQL, not Postgres / typo

chulkilee:
I found that postgrex uses both PostgreSQL and Postgres.

Although Postgres is "widely accepted as nickname or alias" [manual](https://www.postgresql.org/docs/current/history.html), it would be better to stick with the official name here.

... or just change all PostgreSQL to Postgres :)

Also included small typo fix.

### DIFF
--- a/lib/postgrex/builtins.ex
+++ b/lib/postgrex/builtins.ex
@@ -1,6 +1,6 @@
 defmodule Postgrex.Interval do
   @moduledoc """
-  Struct for PostgreSQL interval.
+  Struct for PostgreSQL `interval`.
 
   ## Fields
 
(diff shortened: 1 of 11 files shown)


lib/postgrex/builtins.ex
```diff
@@ -190,7 +190,7 @@ end
 
 defmodule Postgrex.Lexeme do
   @moduledoc """
-  Struct for Postgres Lexeme (A Tsvector type is composed of multiple lexemes)
+  Struct for PostgreSQL lexeme (A tsvector type is composed of multiple lexemes)
```
chulkilee: `lexeme` and `tsvector` are actually type name in lower case. 

Maybe we can wrap them with backticks (e.g.  `` `lexeme` ``)?

josevalim: Backticks sounds like a good call. ALso, please rebase!

chulkilee: Added backticks, and rebased!

josevalim:

PR #472, opened 2019-07-17
mix format

chulkilee:
Ran `mix format --check-equivalent` :)


chulkilee:
One thing we can add: `mix format --check-formatted` to `.travis.yml` - but seems like the default is Elixir 1.6, which does not have mix format.

Probably it can be done with adding some steps, but I don't know much about travis so I'll leave it for others..

josevalim:

PR #481, opened 2019-10-05
Enable PG 11 on travis

thiamsantos:
### Motivation

Recently I got in a discussion at work if we should use pg 11 or a older version because of no explicit support of `postgrex`. So would be nice to enable PG 11 on travis to guarantee that everything is working.

### Proposed solution

Include `PGVERSION=11` on `.travis.yml`


josevalim:
Thanks. The build fails but due to other PG versions, most likely to some changes in Travis.

josevalim:

Issue #433, opened 2019-01-08
Notifications system does not automatically retry connections


bernardd:
For "normal" Postgrex connections (those started with `Postgrex.start_link()`), if the database connection drops for whatever reason, the system will continually retry the connection with a capped backoff, just as one might expect/hope.

Notification connections, however, (those started with `Postgrex.Notifications.start_link()`) just terminate the process and require the user to have set up a supervision tree (or some other method) to handle any retries. Digging into it, the reason is that, in `Postgrex.Notifications.handle_info/2`, the `:error` and `:disconnect` responses from `Protocol.handle_info/2` are explicitly mapped to a `{:stop ...}` tuple. Is there a good reason for this or should I put together a PR that makes notification connections retry the same way other connections do?

josevalim:
I am not sure there is a good reason. @ericmj / @fishcakez ?

We could implement backoff too... but I have to ask, what is the issue with putting it under  a supervisor? :)

fishcakez:
Notification is stateful, and state is lost in such a way it can't be recovered on disconnect.

josevalim:
That answers it, thanks @fishcakez !

fishcakez:
A better description is that the notification process has a pointer to a location in the stream of events on the back end and we lose that point on disconnect. If the error isn't to crash we need to send messages.

I think for those looking to guarantee events and sync with data in a table it's not possible to do unless listen first then read table.

Certainly we could support opt in reconnect etc if someone writes it.

bernardd:
Ah, okay, that makes more sense to me, thanks a lot :) If you're happy to have reconnection as an option, I might look into writing it.

PR #484, opened 2019-10-22
Add auto reconnection option for notifications

bernardd:
I finally got around to implementing the notification auto-reconnect that was discussed over in #433. The basic functionality is enabled by adding the `auto_reconnect: true` to the startup opts.

I haven't written any docs yet because I wanted to run the implementation past you before spending time on that. If you're happy with it, I'll update the docs before it's merged.

### DIFF
--- a/lib/postgrex/notifications.ex
+++ b/lib/postgrex/notifications.ex
@@ -59,11 +59,14 @@ defmodule Postgrex.Notifications do
 
   @timeout 5000
 
-  defstruct idle_interval: 5000,
+  defstruct opts: [],
+            idle_interval: 5000,
             protocol: nil,
             parameters: nil,
             listeners: Map.new(),
-            listener_channels: Map.new()
+            listener_channels: Map.new(),
+            auto_reconnect: false,
+            reconnect_backoff: 500
 
   ## PUBLIC API ##
 
@@ -141,8 +144,31 @@ defmodule Postgrex.Notifications do
   ## CALLBACKS ##
 
   def init(opts) do
+    auto_reconnect = Keyword.get(opts, :auto_reconnect, false)
+    reconnect_backoff = Keyword.get(opts, :reconnect_backoff, 500)
+
+    idle_timeout = opts[:idle_timeout]
+
+    if idle_timeout do
+      require Logger
+
+      Logger.warn(
+        ":idle_timeout in Postgrex.Notifications is deprecated, " <>
+          "please use :idle_interval instead"
+      )
+    end
+
+    idle_interval = Keyword.get(opts, :idle_interval, idle_timeout || 5000)
+
+    state = %__MODULE__{
+      opts: opts,
+      idle_interval: idle_interval,
+      auto_reconnect: auto_reconnect,
+      reconnect_backoff: reconnect_backoff
+    }
+
     if opts[:sync_connect] do
-      case connect(:init, opts) do
+      case connect(:init, state) do
         {:ok, _, _} = ok -> ok
         {:stop, reason, _} -> {:stop, reason}
       end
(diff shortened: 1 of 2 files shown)


josevalim:
It looks great, no comments from me. :)

bernardd:
Docs written and rebased against `master`. Should be all set.

lib/postgrex/notifications.ex
```diff
@@ -72,10 +83,21 @@ defmodule Postgrex.Notifications do
...
+
+  ## Options
+
+    * `:sync_connect` - controls if the connection should be established on boot
+      or asynchronously right after boot. Defaults to `true`.
+
+    * `:auto_reconnect` - automatically attempt to reconnect to the database
+      in event of a disconnection. See the
+      [note about consistency](#module-important-note-about-consistency)
+      above. Defaults to `false`.
```
josevalim:
```suggestion
      above. Defaults to `false`.

```

josevalim:

jscheid:
Thanks for building and adding this - I was just looking for this very feature.

I can't quite seem to get it to work though, I might be missing something?

I'd like my app to be able to start even when a connection can't be established, and I don't want it to crash when an already established connection goes down.

To this end I've tried with `auto_reconnect: true` and `sync_connect: false`, but that gives me an argument error:

```
** (ArgumentError) argument error
    :erlang.apply([hostname: "localhost", sync_connect: false, adapter: Ecto.Adapters.Postgres, types: Myapp.PostgresTypes, database: "postgres", username: "myapp", port: 5432], :opts, [])
    (postgrex) lib/postgrex/notifications.ex:203: Postgrex.Notifications.connect/2
    (stdlib) proc_lib.erl:249: :proc_lib.init_p_do_apply/3
```

With `sync_connect: true` (which I think is not what I want, as it doesn't allow the app to start when psql is unavailable) I'm getting a different error: 

```
** (CaseClauseError) no case clause matching: {:backoff, 500, %Postgrex.Notifications{auto_reconnect: true, idle_interval: 5000, listener_channels: %{}, listeners: %{}, opts: [hostname: "localhost", sync_connect: true, auto_reconnect: true, adapter: Ecto.Adapters.Postgres, types: Myapp.PostgresTypes, database: "postgres", username: "myapp", port: 5432], parameters: nil, protocol: nil, reconnect_backoff: 500}}
    (postgrex) lib/postgrex/notifications.ex:193: Postgrex.Notifications.init/1
```

My `mix.exs` contains this:

```elixir
{:postgrex,  git: "https://github.com/elixir-ecto/postgrex.git", tag: "b39611e027c8e2c2a9a56f7145421eabda7499b4", override: true},
```

My child_spec looks like this:

```elixir
%{
  id: Postgrex.Notifications,
  start:
    {Postgrex.Notifications, :start_link,
     [
       Application.fetch_env!(:myapp, Myapp.Repo)
       |> Keyword.put(:name, Myapp.PostgresNotifications)
       |> Keyword.put(:auto_reconnect, true)
       # |> Keyword.put(:sync_connect, false)
     ]},
},
```

(where `Application.fetch_env!(:myapp, Myapp.Repo)` returns the Postgres connection config.)

I'm fairly new to Elixir - apologies if I'm just doing this wrong.

josevalim:
Good job @jscheid! I have fixed both bugs in master.

jscheid:
Thanks @josevalim for the fast turnaround, I've tested and it works perfectly now.

There's one thing I can't quite figure out yet, apologies for asking this question here even though it's not strictly related:

I have another GenServer that calls `Postgrex.Notifications.listen` when it starts and then responds to incoming messages. This works well, however it breaks with `sync_connect: false` when psql isn't available initially. Is there a way for my GenServer to be notified when the connection first comes up?

josevalim:
@jscheid I have just added a feature that if you try to listen while the connection is down, we will schedule it and try again later!

jscheid:
Amazing... that's even better, thanks!

I can't test it right away but will take it for a spin tomorrow.

jscheid:
All works great now, thanks again!

PR #485, opened 2019-10-22
Fix `start_options` `:ssl_opts`

hauleth:
`:ssl.ssl_option/0` is no more and `:ssl.tls_option/0` is the think that should be used instead otherwise [it can be a little bit confusing to the newcomers](https://elixirforum.com/t/how-to-connect-to-postgrex-with-ssl-certificate/26255?u=hauleth).


josevalim:

PR #486, opened 2019-10-23
added link to docs

Fl4m3Ph03n1x:
Adding more information on where to go for the ssl_options flag. 
Follow up from this forum discussion:

https://elixirforum.com/t/how-to-connect-to-postgrex-with-ssl-certificate/26255

### DIFF
--- a/lib/postgrex.ex
+++ b/lib/postgrex.ex
@@ -80,7 +80,8 @@ defmodule Postgrex do
       (defaults to `:timeout` value);
     * `:ssl` - Set to `true` if ssl should be used (default: `false`);
     * `:ssl_opts` - A list of ssl options, see the
-    [ssl docs](http://erlang.org/doc/man/ssl.html);
+    [`tls_client_option`](http://erlang.org/doc/man/ssl.html#type-tls_client_option)
+    from the ssl docs;
     * `:socket_options` - Options to be given to the underlying socket
       (applies to both TCP and UNIX sockets);
     * `:prepare` - How to prepare queries, either `:named` to use named queries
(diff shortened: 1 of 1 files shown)


lib/postgrex.ex
```diff
@@ -79,7 +79,8 @@ defmodule Postgrex do
     * `:handshake_timeout` - Connection handshake timeout in milliseconds
       (defaults to `:timeout` value);
     * `:ssl` - Set to `true` if ssl should be used (default: `false`);
-    * `:ssl_opts` - A list of ssl options, see ssl docs;
+    * `:ssl_opts` - A list of ssl options, see the
+    [ssl docs](http://erlang.org/doc/man/ssl.html);
```
wojtekmach: good call. How about linking to particular function:

```markdown
* `:ssl_opts` - A list of ssl options, see `:ssl.connect/2`;
```

?

Fl4m3Ph03n1x: I agree that more precise documentation would be more helpful, it's just that as this is my first PR into this project I was unsure if my suggestions was too specific.

From my understanding, the connect options the `ssl_opts` expects is actually [`tls_client_option`](http://erlang.org/doc/man/ssl.html#type-tls_client_option). 

Would this be accurate?

wojtekmach: Yup, let's go with that :+1:.

hauleth:
Partially this was done in the #485, but I agree, that some more information in docs would be helpful for newcomers.

👍

wojtekmach:
@hauleth oh, nice, I missed that pr. @Fl4m3Ph03n1x could use also change that type to use tls_client_option as its more specific?

Fl4m3Ph03n1x:
I did some modifications, please let me know if I missed something !

wojtekmach:
Thank you!

PR #489, opened 2019-11-23
Allow dynamic connection configuration for notifications

jscheid:
When using Ecto, which uses `db_connection`, it's possible to provide dynamic connection configuration for every connection attempt via the `configure` option:

> A function to run before every connect attempt to dynamically configure the options, either a 1-arity fun, `{module, function, args}` with options prepended to `args` or `nil` where only returned options are passed to connect callback (default: `nil`) 

This is useful, for example, because it allows for using [temporary database credentials](https://www.vaultproject.io/docs/secrets/databases/postgresql.html#usage).

The same doesn't work for notifications however, apparently because they don't use the connection pool. This change would allow this.

I've omitted tests so far because I'm not sure this is the right place to fix it. What do you think?

### DIFF
--- a/test/notification_test.exs
+++ b/test/notification_test.exs
@@ -156,4 +156,29 @@ defmodule NotificationTest do
       assert Task.await(async)
     end
   end
+
+  test "dynamic configuration with named function" do
+    {:ok, _} =
+      PN.start_link(
+        configure: {NotificationTest, :configure, baz: :frobnicate},
+        foo: :bar
+      )
+  end
+
+  test "dynamic configuration with anonymous function" do
+    {:ok, _} =
+      PN.start_link(
+        configure: fn opts ->
+          assert :bar = Keyword.get(opts, :foo)
+          Keyword.merge(opts, @opts)
+        end,
+        foo: :bar
+      )
+  end
+
+  def configure(opts) do
+    assert :bar = Keyword.get(opts, :foo)
+    assert :frobnicate = Keyword.get(opts, :baz)
+    Keyword.merge(opts, @opts)
+  end
 end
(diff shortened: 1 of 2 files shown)


josevalim:
Looks good to me. Please add test and docs and we can gladly merge it, thank you!

jscheid:
@josevalim I've pushed a change. I would have liked to use the exact [same code as in db_connection](https://github.com/elixir-ecto/db_connection/blob/9286c8e0bf3db70440f913e1eea8dc08527cfd7a/lib/db_connection/connection.ex#L300) but I couldn't get it to work with `[a | b]` instead of `Keyword.merge(a, b)`. Could you have a look and let me know which of the two is correct and what, if anything, needs changing?

josevalim:
`[a | b]` is correct. The idea is that you want to prepend the `opts` to the arguments given on the configuration. So `{Mod, :fun, [:bar, :baz]}` will be called as `Mod.fun(opts, :bar, :baz)`.

jscheid:
Ah, of course. Done.

test/notification_test.exs
```diff
@@ -160,7 +160,7 @@ defmodule NotificationTest do
   test "dynamic configuration with named function" do
     {:ok, _} =
       PN.start_link(
-        configure: {NotificationTest, :configure, baz: :frobnicate},
+        configure: {NotificationTest, :configure, [:bar, :baz]},
         foo: :bar
       )
```
josevalim: We need to assert configure was called. So you can do something like this:

    {NotificationTest, :configure, [self()]},

And in configure you do:

    send(parent, :configured)

And at the end of this test you do:

    assert_received :configured

jscheid:
This was already implicitly tested because without it being called, the opts would result in a connection failure (I've made this clearer now by setting the default database name to an invalid one.)

Anyway, I've also added explicit checks.

josevalim:

jscheid:
Thanks for your help!

PR #492, opened 2019-12-04
More compatible empty list encoding

### LINKED ISSUE #491: Non-strict encoding of empty arrays
I'm not sure if this is a bug in Postgrex or CockroachDB, but empty lists are currently encoded with the last 2 bytes (length + start index) missing.

For example, an empty int4 array is encoded as `<<0, 0, 0, 12, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 11, 134>>
`, whereas cockroachdb expects: `<<0, 0, 0, 12, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 11, 134, 0, 0, 0, 0, 0, 0, 0, 1>>`

From what I could understand, libpq DOES include the extra two bytes when encoding (if ndims == 0 then ndims = 1 type thing), however it also tolerates the missing lengths when decoding. So how this should be encoded seems ambiguous to me (FWIW, I once contributed some array encoding/decoding logic to go's pgx library, and we included those two bytes also).

karlseguin:
libpq supports decoding 0 dimensional arrays, but it encodes them as
empty 1 dimensional arrays. It's therefore unclear how they should be
encoded. Unfortunately, CockroachDB only supports decoding empty 1
dimensional arrays.

Fixes #491

### DIFF
--- a/lib/postgrex/extensions/array.ex
+++ b/lib/postgrex/extensions/array.ex
@@ -41,6 +41,14 @@ defmodule Postgrex.Extensions.Array do
 
   ## Helpers
 
+  # Special case for empty lists. This treats an empty list as an empty 1 d
+  # imensional array. While libpq will decode an payload encoded for a 0-dim
+  # array, CockroachDB will not. Also, this is how libpq actually encodes 0-dim
+  # arrays.
+  def encode([], elem_oid, _encoder) do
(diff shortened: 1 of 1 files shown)


josevalim:
Thanks! FWIW, it is probably worth reporting a bug on CockroachDB side too.

josevalim:

PR #495, opened 2020-02-04
[fix](deps) bump db_connection to 2.2

areski:
`ecto_sql` depends of `db_connection` ~> 2.2 as it requires the idle_time support [https://github.com/elixir-ecto/ecto_sql/commit/66f21fd7f687be5e76674ee56a74e8e01e626868].
A new release of `postgrex` might be necessary for inclusion into projects.

### DIFF
--- a/mix.exs
+++ b/mix.exs
@@ -32,7 +32,7 @@ defmodule Postgrex.Mixfile do
       {:ex_doc, "~> 0.20", only: :docs},
       {:jason, "~> 1.0", optional: true},
       {:decimal, "~> 1.5"},
-      {:db_connection, "~> 2.1"},
+      {:db_connection, "~> 2.2"},
       {:connection, "~> 1.0"}
     ]
   end


wojtekmach:
Thank you for the PR. I believe we don't need to bump the dependency here but it probably doesn't hurt either. It's ok for ecto_sql to have `db_connection ~> 2.2` and postgrex to have `db_connection ~> 2.1`, when you use both in your project the dependency resolution will pick higher version. Did you run into any bugs?

areski:
Thanks for the clarification, I thought it would conflict, so far I haven't run into any issues.
Feel free to close the PR if this is not needed.

josevalim:
Closing as the current requirement is OK. Thanks!

foggy1:
For the record, this did happen to me on a fresh install today

```
Dependencies have diverged:
* db_connection (Hex package)
  the dependency db_connection 2.1.1

  > In deps/postgrex/mix.exs:
    {:db_connection, "~> 2.1", [env: :prod, hex: "db_connection", repo: "hexpm", optional: false]}

  does not match the requirement specified

  > In deps/ecto_sql/mix.exs:
    {:db_connection, "~> 2.2", [env: :prod, hex: "db_connection", repo: "hexpm", optional: false]}

  Ensure they match or specify one of the above in your deps and set "override: true"
** (Mix) Can't continue due to errors on dependencies
```

josevalim:
Just update them: `mix deps.update db_connection postgrex ecto_sql`.

Issue #515, opened 2020-08-26
Type mismatch in Postgrex.child_spec/1


matehat:
Current implementation of `Postgrex.child_spec` is:

```elixir
@doc """
  Returns a supervisor child specification for a DBConnection pool.
  """
  @spec child_spec(Keyword.t) :: Supervisor.Spec.spec
  def child_spec(opts) do
    opts = [types: true] ++ Postgrex.Utils.default_opts(opts)
    DBConnection.child_spec(Postgrex.Protocol, opts)
  end
```

returning `DBConnection.child_spec`, whose spec is:

```elixir
@spec child_spec(module, opts :: Keyword.t) :: :supervisor.child_spec()
```

But `:supervisor.child_spec()` is not the same as `Supervisor.Spec.spec`

The former is:

```elixir
@type child_spec() ::
  %{
    :id => child_id(),
    :start => mfargs(),
    optional(:restart) => restart(),
    optional(:shutdown) => shutdown(),
    optional(:type) => worker(),
    optional(:modules) => modules()
  }
  | {id :: child_id(), startFunc :: mfargs(), restart :: restart(),
     shutdown :: shutdown(), type :: worker(), modules :: modules()}
```

while the latter being:
```
@type spec() ::
  {child_id(), start_fun :: {module(), atom(), [term()]}, restart(),
   shutdown(), worker(), modules()}
```

All that to say that it's now impossible for me to satisfy dialyzer on this call:

```elixir
  def child_spec(_opts) do
    spec = Postgrex.child_spec(
      pool: DBConnection.ConnectionPool,
      hostname: some_host,
      username: some_username,
      password: some_password,
      database: some_database,
      name: __MODULE__
    )

    case spec do
      {_, start_mfa, restart, shutdown, worker, modules} ->
        {__MODULE__, start_mfa, restart, shutdown, worker, modules}

      spec when is_map(spec) ->
        Map.put(spec, :id, __MODULE__)
    end
  end
```

which complains that the second clause `can never match, because previous clauses completely cover the type`.

It seems the `Postgrex.child_spec` should be

```elixir
@spec child_spec(Keyword.t) :: :supervisor.child_spec()
```

like the spec of the function that returns last.

PR #516, opened 2020-08-26
Use :supervisor as child_spec/1 return type

matehat:
Simple fix for #515

### DIFF
--- a/lib/postgrex.ex
+++ b/lib/postgrex.ex
@@ -491,7 +491,7 @@ defmodule Postgrex do
   @doc """
   Returns a supervisor child specification for a DBConnection pool.
   """
-  @spec child_spec([start_option]) :: Supervisor.Spec.spec()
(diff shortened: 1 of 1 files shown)


wojtekmach:
Thank you!

lib/postgrex.ex
```diff
@@ -491,7 +491,7 @@ defmodule Postgrex do
   @doc """
   Returns a supervisor child specification for a DBConnection pool.
   """
-  @spec child_spec([start_option]) :: Supervisor.Spec.spec()
+  @spec child_spec([start_option]) :: :supervisor.child_spec()
```
wojtekmach, line 494: @josevalim should we use `Supervisor.child_spec()` here instead? If so, we should do that in DBConnection too. No need to change MyXQL as it already uses that.

It's probably fine either way but I'm happy to make the change here and in other places if that'd useful.

PR #533, opened 2021-01-21
WIP: Fix crash following mishandled result.

### LINKED ISSUE #498: Cannot connect to server with `client_min_messages` >= `debug[1-5]`
PostgreSQL 12.2
Postgrex 0.15.3

(Similar to #451 ?)

Cannot connect to a 12.2 server when it has config: `client_min_messages = debug1`, and `debug{2,3,4,5}`

I have added whitespace to the backtrace to make is easier to see the four parameters to `msg_recv()`.

```elixir
iex(2)> {:ok, pid} = Postgrex.start_link(hostname: "localhost", username: "postgres", password: "postgres", database: "test", show_sensitive_data_on_connection_error: true)
{:ok, #PID<0.2853.0>}
iex(3)> [error] GenServer #PID<0.2855.0> terminating

** (FunctionClauseError) no function clause matching in Postgrex.Protocol.msg_recv/4

    (postgrex 0.15.3) lib/postgrex/protocol.ex:2836: Postgrex.Protocol.msg_recv(
  {
    {%Postgrex.Protocol{buffer: nil, connection_id: 37704, connection_key: -1269269119, disconnect_on_error_codes: [], null: nil, parameters: %
    {"DateStyle" => "ISO, MDY", "IntervalStyle" => "postgres", "TimeZone" => "America/Toronto", "application_name" => "", "client_encoding" => "UTF8", "integer_datetimes" => "on", "is_superuser" => "on", "server_encoding" =>
     "UTF8", "server_version" => "12.2", "session_authorization" => "postgres", "standard_conforming_strings" => "on"}, peer: {{127, 0, 0, 1}, 5432}, postgres: :idle, queries: nil, sock: {:gen_tcp, #Port<0.1231>}, timeout: 1
    5000, transactions: :naive, types: {Postgrex.DefaultTypes, #Reference<0.3052583275.3171549186.195678>}}, %{messages: [%{code: "00000", file: "catcache.c", line: "867", message: "rehashing catalog cache id 43 for pg_proc;
     257 tups, 128 buckets", routine: "RehashCatCache", severity: "DEBUG", unknown: "DEBUG"}], opts: [pool_index: 1, types: Postgrex.DefaultTypes, hostname: "localhost", username: "postgres", password: "postgres", database:
    "test", show_sensitive_data_on_connection_error: true], prepare: :named, ssl: false, types_key: {'localhost', 5432, "test"}, types_lock: {#PID<0.2799.0>, #Reference<0.3052583275.3171418117.194991>}, types_mod:
    Postgrex.DefaultTypes}},

    %{messages: [%{code: "00000", file: "catcache.c", line: "867", message: "rehashing catalog cache id 42 for pg_proc; 257 tups, 128 buckets", routine: "RehashCatCache", severity: "DEBUG", unknown:
    "DEBUG"}], opts: [pool_index: 1, types: Postgrex.DefaultTypes, hostname: "localhost", username: "postgres", password: "postgres", database: "test", show_sensitive_data_on_connection_error: true], prepare: :named, ss
    l: false, types_key: {'localhost', 5432, "test"}, types_lock: {#PID<0.2799.0>, #Reference<0.3052583275.3171418117.194991>}, types_mod: Postgrex.DefaultTypes}
  },
  :infinity,
  "",
  0)

    (postgrex 0.15.3) lib/postgrex/protocol.ex:816: Postgrex.Protocol.bootstrap_recv/4
    (postgrex 0.15.3) lib/postgrex/protocol.ex:579: Postgrex.Protocol.handshake/2
    (db_connection 2.2.1) lib/db_connection/connection.ex:69: DBConnection.Connection.connect/2
    (connection 1.0.4) lib/connection.ex:622: Connection.enter_connect/5
    (stdlib 3.11.2) proc_lib.erl:249: :proc_lib.init_p_do_apply/3
Last message: nil
State: Postgrex.Protocol
```
```

### PR DESCRIPTION
**Pull request by richardkmichael**
`bootstrap_recv/4` and in turn `msg_recv/4` expect their first argument
to be a Map (usually a `%Postgrex.Protocol{}` struct), but this
case-clause assigns the entire two element tuple return from
`handle_msg/3` to `s` and then passes it onward.

```
** (FunctionClauseError) no function clause matching in Postgrex.Protocol.msg_recv/4
    (postgrex 0.15.8) lib/postgrex/protocol.ex:2837: Postgrex.Protocol.msg_recv({%Postgrex.Protocol{...
                                                       should not be a tuple --^^^
```

Although a bug has not yet occurred, a similar problem seems likely due
to same assignment and call-chain in `reload_recv/4` so it has also been
changed.

Closes #498

### DIFF
--- a/lib/postgrex/protocol.ex
+++ b/lib/postgrex/protocol.ex
@@ -829,7 +829,7 @@ defmodule Postgrex.Protocol do
         bootstrap_fail(s, err, status, buffer)
 
       {:ok, msg, buffer} ->
-        s = handle_msg(s, status, msg)
+        {s, status} = handle_msg(s, status, msg)
         bootstrap_recv(s, status, type_infos, buffer)
 
       {:disconnect, err, s} ->
@@ -1578,7 +1578,7 @@ defmodule Postgrex.Protocol do
         bootstrap_fail(s, err, status, buffer)
 
       {:ok, msg, buffer} ->
-        s = handle_msg(s, status, msg)
+        {s, status} = handle_msg(s, status, msg)
         reload_recv(s, status, acc, buffer)
 
       {:disconnect, err, s} ->


richardkmichael:
I kept the commit message short, but to help a reviewer, here is a longer explanation of how I encountered this problem and what I believe is the fix and why.

(This was my first time through this code base, and I am relatively new to Elixir, certainly given it's use in this project.  I read a fair bit of the protocol and message code.  Hopefully this is the right fix, but I am not sure given fd94dd8c was an intentional optimization.)

----

When a message which is not a row description, data row, command
completion or error message (these are correctly handled by preceding
case-clauses) occurs and there is buffer remaining to be processed,
Postgrex will crash.

Configuring Postgres with `client_min_messages = "DEBUG1"` will cause
trigger this bug.

This occurs because the required variant `handle_msg/3` returns a two
element tuple (specifically, `{%Postgrex.Protocol{}, status}`), and this
case-clause assigns the entire tuple result to `s` (notably different
from other call-sites to `handle_msg/3`).  When `s` is subsequently
passed onward to `bootstrap_recv/4` to continue processing the buffer,
it is expected to be only the %Postgrex.Protocol struct.  (Aside, all
variants of `handle_msg/3` return a two element tuple.)

The bug is not specific to `DEBUG*` severity notice messages; indeed
such messages are decoded and processed correctly in some circumstances
(at the end of a buffer, for example).  However, the bug has been
revealed by `DEBUG1` notice messages because the Postgres server emits
such messages in the middle of a stream of data row messages, which
means a notice message must be handle and then the remaining buffer
processed.  (This difference in processing is hinted at by examining the
Postgres protocol message stream using Wireshark; the notice messages
are interspersed in data-row messages when the bug is triggered.)

Reading the code, I believe the bug would emerge if a notify or
parameter message (these are each handled by a variant of
`handle_msg/3`) were also emitted with subsequent buffer to process.

This assignment appears intentional in fd94dd8c but it seems like a
mistake as the first argument to many of the message and buffer handling
`_send|_recv` functions has remained a `%Postgrex.Protocol` struct.

Also, there is a similar assignment in `reload_recv/4` which has not
triggered a bug, but has the same problem: it passes a tuple to the
continued buffer processing, and so may be a problem.  It has been
changed as well.

josevalim:

richardkmichael:
@josevalim You are very fast!  But I am sorry, I just adjusted my comment regarding `status`, which contains the message queue (`status` is `%{ messages: [...]}`.  There is one additional message after `handle_msg/3`.  I am not yet sure which is correct, or if I introduced a new problem!  Do you know the patch was correct?

josevalim:
This patch is definitely correct! Maybe there is another bug, but, if so, it would be in addition to this. :)

richardkmichael:
Great, thank you! 😌

PR #539, opened 2021-04-08
Upgrade jason dependency

oleks:
Allow jason `>= 1.0.0` instead of `~> 1.0`.

For an overview of what has happened to jason since 1.0.0, see https://github.com/michalmuskala/jason/compare/v1.0.0...v1.2.2


wojtekmach:
`~> 1.0` is equivalent to `>= 1.0.0 and < 2.0.0` so this change is unnecessary. Thanks for the PR!

oleks:
> `~> 1.0` is equivalent to `>= 1.0.0 and < 2.0.0` so this change is unnecessary. Thanks for the PR!

Thanks! That was enlightening :smiley:

PR #544, opened 2021-05-19
Improve error handling when using multiple endpoints

chaodhib:
Closes https://github.com/elixir-ecto/postgrex/issues/542

If Postgrex tries to connect to more than one database instance and if the connection fails for all of them, one global exception is raised and its message is a concatenation of all the exceptions message (one exception per connection failed). 

This PR doesn't alter the behavior of the library if connection is attempted to only one instance (using either the `hostname` & `port` config options, the UNIX socket option or the `endpoints` option with only one {hostname,port} tuple).

### DIFF
--- a/lib/postgrex/protocol.ex
+++ b/lib/postgrex/protocol.ex
@@ -178,7 +178,7 @@ defmodule Postgrex.Protocol do
       end)
       |> Enum.join(", ")
 
-    {:error, %Postgrex.Error{message: "[#{concat_messages}]"}}
+    {:error, %Postgrex.Error{message: "failed to establish connection to multiple endpoints: #{concat_messages}"}}
   end
 
   defp connect_and_handshake(host, port, sock_opts, timeout, s, status) do
--- a/lib/postgrex/protocol.ex
+++ b/lib/postgrex/protocol.ex
@@ -104,7 +104,7 @@ defmodule Postgrex.Protocol do
       target_server_type: target_server_type
     }
 
-    connect_endpoints(endpoints, sock_opts ++ @sock_opts, connect_timeout, s, status)
+    connect_endpoints(endpoints, sock_opts ++ @sock_opts, connect_timeout, s, status, [])
   end
 
   defp endpoints(opts) do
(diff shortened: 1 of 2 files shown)


lib/postgrex/protocol.ex
```diff
@@ -165,12 +156,21 @@ defmodule Postgrex.Protocol do
       {:ok, _} = ret ->
         ret
 
-      {:error, _} = err ->
-        connect_endpoints(remaining_endpoints, sock_opts, timeout, s, status, err)
+      {:error, err} ->
+        connect_endpoints(remaining_endpoints, sock_opts, timeout, s, status, previous_errors ++ [err])
```
josevalim, line 160: It is better to prepend and then reverse at the end:

```suggestion
        connect_endpoints(remaining_endpoints, sock_opts, timeout, s, status, [err | previous_errors])
```

It won't matter here, as those lists are likely small, but it is best to adhere to best practices. :)

lib/postgrex/protocol.ex
```diff
@@ -165,12 +156,21 @@ defmodule Postgrex.Protocol do
...
     end
   end
 
-  defp connect_endpoints([], _, _, _, _, err), do: err
+  defp connect_endpoints([], _, _, _, _, [error]), do: {:error, error}
+
+  defp connect_endpoints([], _, _, _, _, errors) when is_list(errors) do
+    concat_messages =
+      errors
+      |> Enum.map(fn %error_module{} = error -> "#{error_module}: \"#{Exception.message(error)}\"" end)
```
josevalim:
```suggestion
      |> Enum.map(fn %error_module{} = error -> "#{inspect(error_module)}: \"#{Exception.message(error)}\"" end)
```

lib/postgrex/protocol.ex
```diff
@@ -165,12 +156,30 @@ defmodule Postgrex.Protocol do
...
+
+  defp connect_endpoints([], _, _, _, _, errors) when is_list(errors) do
+    concat_messages =
+      errors
+      |> Enum.map(fn %error_module{} = error ->
+        "#{error_module}: \"#{Exception.message(error)}\""
+      end)
+      |> Enum.join(", ")
+
+    {:error, %Postgrex.Error{message: "[#{concat_messages}]"}}
```
josevalim, line 181:
```suggestion
    {:error, %Postgrex.Error{message: "failed to establish connection to multiple endpoints: #{concat_messages}"}}
```

lib/postgrex/protocol.ex
```diff
@@ -165,12 +156,30 @@ defmodule Postgrex.Protocol do
...
-      {:error, _} = err ->
-        connect_endpoints(remaining_endpoints, sock_opts, timeout, s, status, err)
+      {:error, err} ->
+        connect_endpoints(
+          remaining_endpoints,
+          sock_opts,
+          timeout,
+          s,
+          status,
+          previous_errors ++ [err]
```
josevalim, line 166: It is better to prepend and then reverse at the end:

```suggestion
        connect_endpoints(remaining_endpoints, sock_opts, timeout, s, status, [err | previous_errors])
```

It won't matter here, as those lists are likely small, but it is best to adhere to best practices. :)

josevalim:

chaodhib:
Thanks for the review!

PR #547, opened 2021-05-31
Support negative years for date, timestamp and timestampz types

kipcole9:
Supports negative years for the types date, timestamp,
timestampz. A side effect is that the `:calendar` module
is no longer used.

**This implementation support only Elixir 1.11 and 1.12** because
of its use of `NaiveDateTime.to_gregorian_seconds/1`. I looked
at using the private `rata die` functions but that would require
reimplementation of quite a lot of code - or making some private
functions public.

* Support years down to -4713 in date, timestamp and
timestampz type since that is the minimum year supported
by Postgresql

* Support years up to 9999 in timestampz (which is less
than that supported by [Postgresql](https://www.postgresql.org/docs/9.1/datatype-datetime.html))
but is the maximum supported by `Calendar.ISO`.

* Add tests for negative dates, timestamps and timestampz

### DIFF
--- a/lib/postgrex/extensions/date.ex
+++ b/lib/postgrex/extensions/date.ex
@@ -11,7 +11,7 @@ defmodule Postgrex.Extensions.Date do
   @max_days Date.to_gregorian_days(~D[9999-12-31])
 
   # Elixir supports earlier dates but this is the
-  # earliest supported in Postfresql
+  # earliest supported in Postgresql.
   @min_days Date.to_gregorian_days(~D[-4713-01-01])
 
   def encode(_) do
--- a/lib/postgrex/extensions/date.ex
+++ b/lib/postgrex/extensions/date.ex
@@ -3,9 +3,16 @@ defmodule Postgrex.Extensions.Date do
   import Postgrex.BinaryUtils, warn: false
   use Postgrex.BinaryExtension, send: "date_send"
 
-  @gd_epoch :calendar.date_to_gregorian_days({2000, 1, 1})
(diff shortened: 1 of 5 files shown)


test/calendar_test.exs
```diff
@@ -60,14 +60,19 @@ defmodule CalendarTest do
...
 
+  # All Postgres BC dates are decodable to Elixir
+  # Dates
+
+  # test "decode date lower bound error", context do
+  #   assert_raise ArgumentError, fn ->
+  #     query("SELECT date 'January 1, 4713 BC'", [])
+  #   end
+  # end
+
```
josevalim, line 75:
```suggestion
```

josevalim:
@kipcole9 looks great! Can you please update the CI file so the entry:

    PG 13, 1.6.6, 19.3.6.13

Becomes:

    PG 13, 1.11.4, 21.3

?

Thanks!

lib/postgrex/extensions/date.ex
```diff
@@ -3,9 +3,16 @@ defmodule Postgrex.Extensions.Date do
...
+  @gd_epoch Date.to_gregorian_days(~D[2000-01-01])
   @max_year 9999
-  @max_days 3_652_424
+
+  # Latest date supported by Elixir. Postgresql
+  # supports later dates.
+  @max_days Date.to_gregorian_days(~D[9999-12-31])
+
+  # Elixir supports earlier dates but this is the
+  # earliest supported in Postfresql
```
josevalim, line 14:
```suggestion
  # earliest supported in Postgresql.
```

josevalim:

PR #555, opened 2021-06-23
Fix invalid type error after failover

chaodhib:
This PR fixes https://github.com/elixir-ecto/postgrex/issues/550

The change to `recv_parse_describe` ensures that in case the Protocol & Query's `types` differ, the Query gets a new ref (to trigger an update to the Ecto cache) and gets the Protocol's `types` (both of these are done in `describe_result/4`).

Todo:
- [x] Maybe the same change should be done for the streaming of a prepared query?

### DIFF
--- a/lib/postgrex/protocol.ex
+++ b/lib/postgrex/protocol.ex
@@ -1345,12 +1345,12 @@ defmodule Postgrex.Protocol do
   end
 
   defp recv_parse_describe(
-         %{types: proto_types} = s,
+         %{types: protocol_types} = s,
          status,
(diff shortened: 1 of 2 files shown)


josevalim:
> Maybe the same change should be done for the streaming of a prepared query?

I will review this soon but I just want to say the answer to this question is yes. :)

chaodhib:
> 
> 
> > Maybe the same change should be done for the streaming of a prepared query?
> 
> I will review this soon but I just want to say the answer to this question is yes. :)

ok will do :)

josevalim:

Issue #549, opened 2021-06-15
dialyzer issues with elixir 1.12.0


bitboxer:
I just tried to update a project to elixir 1.12.0 and dialyzer spits out these messages:

```
deps/postgrex/lib/postgrex/type_module.ex:897:improper_list_constr
List construction (cons) will produce an improper list, because its second argument is nonempty_improper_list(binary(), bitstring()).
________________________________________________________________________________
deps/postgrex/lib/postgrex/type_module.ex:897:improper_list_constr
List construction (cons) will produce an improper list, because its second argument is bitstring().
________________________________________________________________________________
deps/postgrex/lib/postgrex/type_module.ex:897:improper_list_constr
List construction (cons) will produce an improper list, because its second argument is binary().
```

Has anyone here seen those?

J3RN:
I get this error as well

josevalim:
I don’t run Dialyzer as part of my workflow, so I would appreciate if someone could investigate this and send PRs.

FWIW, the reason why we have Dialyzer warnings on TypeModule is because it does code generation, so it likely generates clauses that can never match for some types but it matches for others, and Dialyzer can see through those. Sometimes the proper fix is to add the generated: true to the quoted code.

bitboxer:
For everyone here: the hot fix is to add a file called `.dialyzer_ignore.exs ` and add this into it:

```elixir
[
   {"deps/postgrex/lib/postgrex/type_module.ex"},
   {"lib/postgrex/type_module.ex"}
]
```

thbar:
For some reason the previous hot fix did not work for me, I had to use a slightly different version. I wonder if this may be because the app I'm referring to is an umbrella app.

Current fix:

```elixir
[
   # temporary fix for https://github.com/elixir-ecto/postgrex/issues/549
   ~r/deps\/postgrex\/lib\/postgrex\/type_module.ex/,
   ~r/lib\/postgrex\/type_module.ex/
 ]
```

al2o3cr:
The warning is accurate, in the sense that it says "an improper list is being constructed here" and one is - the only report I got with a sensible line number pointed to line 358:
```
      defp decode_rows(<<?D, size::int32, rest::binary>>, rem, _, rows) do
        more = size + 1 - rem
        {:more, [?D, <<size::int32>> | rest], rows, more}   # <= MESSAGE POINTS HERE
      end
```
[...]

PR #556, opened 2021-07-06
Mark code that uses improper lists as generated

al2o3cr:
Addresses [#549](https://github.com/elixir-ecto/postgrex/issues/549) by marking code that creates improper lists as `generated: true`.

If it's preferable to explicitly opt-out of the specific Dialyzer warnings (using `@dialyzer`), that's an easy change.

### DIFF
--- a/lib/postgrex/extensions/bit_string.ex
+++ b/lib/postgrex/extensions/bit_string.ex
@@ -6,7 +6,7 @@ defmodule Postgrex.Extensions.BitString do
   def init(opts), do: Keyword.fetch!(opts, :decode_binary)
 
   def encode(_) do
-    quote location: :keep do
+    quote location: :keep, generated: true do
       val when is_binary(val) ->
         [<<byte_size(val) + 4::int32, bit_size(val)::uint32>> | val]
 
--- a/lib/postgrex/type_module.ex
+++ b/lib/postgrex/type_module.ex
@@ -333,7 +333,7 @@ defmodule Postgrex.TypeModule do
   end
 
   defp decode_rows(dispatch, rest, acc, rem, full, rows) do
-    quote location: :keep do
+    quote location: :keep, generated: true do
       def decode_rows(binary, types, rows) do
         decode_rows(binary, byte_size(binary), types, rows)
       end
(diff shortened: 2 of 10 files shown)


lib/postgrex/extensions/bit_string.ex
```diff
@@ -6,7 +6,7 @@ defmodule Postgrex.Extensions.BitString do
   def init(opts), do: Keyword.fetch!(opts, :decode_binary)
 
   def encode(_) do
-    quote location: :keep do
+    quote location: :keep, generated: true do
```
al2o3cr, line 9: This is a common pattern: the code explicitly makes a _definitely_ improper list; this is fine for iodata and saves a cons cell.

lib/postgrex/type_module.ex
```diff
@@ -333,7 +333,7 @@ defmodule Postgrex.TypeModule do
   end
 
   defp decode_rows(dispatch, rest, acc, rem, full, rows) do
-    quote location: :keep do
+    quote location: :keep, generated: true do
```
al2o3cr, line 336: This one covers clauses like:
```
      defp decode_rows(<<?D, rest::binary>>, _, _, rows) do
        {:more, [?D | rest], rows, 0}
      end
```
in the code; is there a performance difference between creating an improper list which is immediately turned back into a binary and passing along the binary directly (maybe related to "small piece of a big buffer" issues?)
```
      defp decode_rows(<<?D, rest::binary>> = whole_thing, _, _, rows) do
        {:more, whole_thing, rows, 0}
      end
```

lib/postgrex/type_module.ex
```diff
@@ -464,7 +464,7 @@ defmodule Postgrex.TypeModule do
         decode_tuple_dispatch(extension, format, rest, oids, n, acc)
       end
 
-    quote do
+    quote generated: true do
```
al2o3cr, line 467: This covers the code starting at line 500:
```
        case Postgrex.Types.fetch(oid, types) do
          {:ok, {:binary, type}} ->
            unquote(oids) = rem - 1

            case [type | types] do
              unquote(dispatch)
            end
```

That `case [type | types] do` code is tricky - it _looks_ like a boring "prepend to list" operation, but AFAIK `types` is never a list (it's a `Postgres.Types.state`!) so it's always making an improper list.

josevalim:

PR #559, opened 2021-07-11
[#558] Support xid8 type

### LINKED ISSUE #558: Support xid8 type
Hello postgrex devs 👋 

[PostgreSQL 13 introduced a new `xid8` type](https://commitfest.postgresql.org/27/2271/) and for its new 64-bit long transaction IDs (basically to ensure uniqueness over the lifespan of the DB). We'd like to be able to query these using Postgrex.

## Actual Behaviour

```
iex(4)> Ecto.Adapters.SQL.query!(MyApp.Repo, "SELECT '1'::xid8;")
** (Postgrex.QueryError) type `xid8` can not be handled by the types module Postgrex.DefaultTypes
    (ecto_sql 3.6.1) lib/ecto/adapters/sql.ex:749: Ecto.Adapters.SQL.raise_sql_call_error/1
[debug] QUERY ERROR db=0.0ms queue=3.8ms idle=262.4ms
SELECT '1'::xid8; []
iex(4)> [error] Postgrex.Protocol (#PID<0.672.0>) disconnected: ** (Postgrex.QueryError) type `xid8` can not be handled by the types module Postgrex.DefaultTypes
```

Side-note: Postgres doesn't have a int8->xid8 cast function, hence the quotes above.

## Desired behaviour

Like `xid`:

```
iex(4)> Ecto.Adapters.SQL.query!(MyApp.Repo, "SELECT '1'::xid;")
[debug] QUERY OK db=0.2ms queue=0.3ms idle=982.4ms
SELECT '1'::xid; []
%Postgrex.Result{
  columns: ["xid"],
  command: :select,
  connection_id: 87864,
  messages: [],
  num_rows: 1,
  rows: [[1]]
}
```

## Solution

We currently run a custom extension:

```elixir
defmodule Postgrex.Extensions.Xid8 do
  @moduledoc false
  import Postgrex.BinaryUtils, warn: false
  use Postgrex.BinaryExtension, send: "xid8send"

  @xid8_range 0..18_446_744_073_709_551_615

  def encode(_) do
    range = Macro.escape(@xid8_range)

    quote location: :keep do
      # BinaryUtils doesn't have uint64 yet
      int when is_integer(int) and int in unquote(range) ->
        <<8::int32, int::unsigned-64>>

      other ->
        raise DBConnection.EncodeError, Postgrex.Utils.encode_msg(other, unquote(range))
    end
  end

  def decode(_) do
    quote location: :keep do
      <<8::int32, int::unsigned-64>> -> int
    end
  end
end
```

I'm happy to come up with a PR if this is something you're interested in and if you could point me in the right direction with regards to:

* testing without hard dependency on PostgreSQL 13
* I noticed that `xid` is in fact cast by `Postgrex.Extensions.OID` - should this also play a role for `xid8`?

Thanks for considering! And thanks for all the work on the library!

Best,
malte

maltoe:
Fixes #558 

This patch adds support for the `xid8` type introduced in PostgreSQL 13.

### DIFF
--- a/lib/postgrex/extensions/xid8.ex
+++ b/lib/postgrex/extensions/xid8.ex
@@ -1,0 +1,25 @@
+defmodule Postgrex.Extensions.Xid8 do
+  @moduledoc false
+  import Postgrex.BinaryUtils, warn: false
+  use Postgrex.BinaryExtension, send: "xid8send"
+
+  @xid8_range 0..18_446_744_073_709_551_615
+
+  def encode(_) do
+    range = Macro.escape(@xid8_range)
+
+    quote location: :keep do
+      int when is_integer(int) and int in unquote(range) ->
+        <<8::int32, int::uint64>>
+
+      other ->
+        raise DBConnection.EncodeError, Postgrex.Utils.encode_msg(other, unquote(range))
+    end
+  end
+
+  def decode(_) do
+    quote location: :keep do
+      <<8::int32, int::uint64>> -> int
+    end
+  end
+end
--- a/lib/postgrex/extensions/xid8.ex
+++ b/lib/postgrex/extensions/xid8.ex
@@ -9,7 +9,7 @@ defmodule Postgrex.Extensions.Xid8 do
     range = Macro.escape(@xid8_range)
 
     quote location: :keep do
-      int when is_integer(int) and int in unquote(range) ->
+      int when int in unquote(range) ->
         <<8::int32, int::uint64>>
 
       other ->
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,10 @@
 # Changelog
 
+## Unreleased
+
+* Enhancements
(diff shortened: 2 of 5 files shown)


lib/postgrex/extensions/xid8.ex
```diff
@@ -0,0 +1,25 @@
...
+  import Postgrex.BinaryUtils, warn: false
+  use Postgrex.BinaryExtension, send: "xid8send"
+
+  @xid8_range 0..18_446_744_073_709_551_615
+
+  def encode(_) do
+    range = Macro.escape(@xid8_range)
+
+    quote location: :keep do
+      int when is_integer(int) and int in unquote(range) ->
```
josevalim, line 12:
```suggestion
      int when int in unquote(range) ->
```

josevalim:

josevalim:
Last one should be fixed in master. I thought the CI ran on the original PR but I forgot we now need to approve them before, so my bad. :)

PR #564, opened 2021-08-22
Fix reraise usage in DBConnection.Query.encode/3 for Postgrex.Copy

michallepicki:
reraise/2 takes stacktrace as second argument,
reraise/3 accepts optional arguments for the exception


josevalim:

PR #565, opened 2021-08-25
Fix possible crash on getting default opts

monshq:
My application could not start because of empty PGPORT env variable. That behaviour was completely unexpected, because I provided ecto with port in application settings. So env variable wasn't going to be used anyway, but prevented application from starting.

### DIFF
--- a/lib/postgrex/utils.ex
+++ b/lib/postgrex/utils.ex
@@ -88,7 +88,8 @@ defmodule Postgrex.Utils do
     |> Keyword.put_new(:password, System.get_env("PGPASSWORD"))
     |> Keyword.put_new(:database, System.get_env("PGDATABASE"))
     |> Keyword.put_new(:hostname, System.get_env("PGHOST") || "localhost")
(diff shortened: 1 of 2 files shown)


lib/postgrex/utils.ex
```diff
@@ -88,7 +88,8 @@ defmodule Postgrex.Utils do
     |> Keyword.put_new(:password, System.get_env("PGPASSWORD"))
     |> Keyword.put_new(:database, System.get_env("PGDATABASE"))
     |> Keyword.put_new(:hostname, System.get_env("PGHOST") || "localhost")
-    |> Keyword.update(:port, normalize_port(System.get_env("PGPORT")), &normalize_port/1)
+    |> Keyword.put_new(:port, System.get_env("PGPORT"))
+    |> Keyword.update(:port, nil, &normalize_port/1)
```
josevalim, line 92:
```suggestion
    |> Keyword.update!(:port, &normalize_port/1)
```

josevalim:

PR #577, opened 2021-11-25
allow specification of replication start lsn

greg-rychlewski:
This will allow the user to specify the LSN number to start replication at. 

It can be useful when, for instance, the server crashes and the replication slot's current position wasn't persisted to disk. This will cause some messages to be sent that the client already received. This will allow the client to restart replication where it left off.


josevalim:

PR #579, opened 2021-11-28
replication functions

greg-rychlewski:
Related to https://github.com/elixir-ecto/postgrex/issues/578.

This PR alters `Postgrex.Replication` so that the initialization function doesn't start the replication process, but instead delegates those actions to different functions that can be used after the connection starts. 

3 functions have been added: `create_slot/4`, `drop_slot/3` and `start_replication/3`.

After replication has started, no other commands can be given to the connection so that the messages don't jumble.

### DIFF
--- a/lib/postgrex/replication.ex
+++ b/lib/postgrex/replication.ex
@@ -143,7 +143,6 @@ defmodule Postgrex.Replication do
             state: nil,
             auto_reconnect: false,
             reconnect_backoff: 500,
-            replication_started: false,
             replication_opts: nil
 
   ## PUBLIC API ##
@@ -510,8 +509,10 @@ defmodule Postgrex.Replication do
        when error in [:error, :disconnect] do
     {:connect, :reconnect, s}
   end
+  
+  defp maybe_restart_replication(%{replication_opts: nil} = s), do: {:ok, s}
 
-  defp maybe_restart_replication(%{replication_started: true} = s) do
+  defp maybe_restart_replication(s) do
     start = command(:start_replication, s.replication_opts)
 
     with {:ok, protocol} <- Protocol.handle_replication(start, s.protocol),
@@ -525,10 +526,10 @@ defmodule Postgrex.Replication do
     end
   end
 
-  defp maybe_restart_replication(s), do: {:ok, s}
+  
 
   ## Queries
-  defp(command(:create_slot, opts)) do
+  defp command(:create_slot, opts) do
     slot = Keyword.fetch!(opts, :slot)
     temporary? = Keyword.get(opts, :temporary, true)
     plugin = Keyword.fetch!(opts, :plugin)
--- a/lib/postgrex/replication.ex
+++ b/lib/postgrex/replication.ex
@@ -302,9 +302,9 @@ defmodule Postgrex.Replication do
 
   ## Options
 
-    * `:wait` - When `true`, waits for an active slot to become inactive before
-      dropping it. When `false`, raises an error when an attemping to drop an active slot.
-      Defaults to `true`.
+    * `:wait` - When `true`, blocks while the slot is being used by a connection.
+      When `false`, returns an error if the slot is being used by a connection.
+      Defaults to `false`.
 
     * `:timeout` - Call timeout.
       Defaults to `5000`.
--- a/lib/postgrex/replication.ex
+++ b/lib/postgrex/replication.ex
@@ -249,7 +249,6 @@ defmodule Postgrex.Replication do
 
     * `:reconnect_backoff` - time (in ms) between reconnection attempts when
       `auto_reconnect` is enabled. Defaults to `500`.
-
   """
   @spec start_link(module(), term(), Keyword.t()) ::
           {:ok, pid} | {:error, Postgrex.Error.t() | term}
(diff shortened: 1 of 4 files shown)


v0idpwn (commented):
Looks great to me! Please note that the `@doc` blocks need formatting!

lib/postgrex/replication.ex
```diff
@@ -325,23 +388,9 @@ defmodule Postgrex.Replication do
...
-             {:ok, protocol} <- Protocol.handle_replication(start, protocol),
-             {:ok, protocol} <- Protocol.checkin(protocol) do
-          {:ok, %{s | protocol: protocol}}
-        else
-          # If we can't start replication, we assume that auto_reconnect can't
-          # solve it either, so we just raise the error as a readable message.
-          {_error, reason, _protocol} ->
-            raise reason
-        end
+        {:ok, %{s | protocol: protocol}}
```
josevalim, line 393: I think `auto_reconnect` will no longer work as expected now. I think that, if replication has started, we should store the replication options in the connection state and automatically restart replication. WDYT?

lib/postgrex/replication.ex
```diff
@@ -286,6 +259,96 @@ defmodule Postgrex.Replication do
     Connection.start_link(__MODULE__, {module, arg, connection_opts}, server_opts)
   end
 
+  @doc """
+  Creates a logical replication slot with the given name and output plugin.
+  By default, PostgreSQL includes the `pgoutput` plugin.
```
josevalim, line 264: Keep the first paragraph short as it is used as a summary by exdoc. :)

```suggestion

  By default, PostgreSQL includes the `pgoutput` plugin.
```

lib/postgrex/replication.ex
```diff
@@ -286,6 +259,96 @@ defmodule Postgrex.Replication do
...
+      Defaults to `true`.
+
+    * `:snapshot` - The type of logical snapshot for the slot. Must be one of
+      `:export`, `:noexport`, or `:use`.
+      Defaults to `:export`.
+
+    * `:timeout` - Call timeout.
+      Defaults to `5000`.
+
+  To better understand the meaning of those options, [see PostgreSQL replication docs](https://www.postgresql.org/docs/14/protocol-replication.html).
```
josevalim, line 287:
```suggestion
  To better understand the meaning of those options,
  [see PostgreSQL replication docs](https://www.postgresql.org/docs/14/protocol-replication.html).
```

lib/postgrex/replication.ex
```diff
@@ -286,6 +259,96 @@ defmodule Postgrex.Replication do
...
+  ## Options
+
+    * `:wait` - When `true`, blocks while the slot is being used by a connection.
+      When `false`, returns an error if the slot is being used by a connection.
+      Defaults to `false`.
+
+    * `:timeout` - Call timeout.
+      Defaults to `5000`.
+
+  To better understand the meaning of those options, [see PostgreSQL replication docs](https://www.postgresql.org/docs/14/protocol-replication.html).
```
josevalim, line 312:
```suggestion
  To better understand the meaning of those options,
  [see PostgreSQL replication docs](https://www.postgresql.org/docs/14/protocol-replication.html).
```

lib/postgrex/replication.ex
```diff
@@ -352,6 +401,49 @@ defmodule Postgrex.Replication do
...
+      {:error, reason, protocol} ->
+        {:reply, {:error, reason}, %{s | protocol: protocol}}
+
+      {:disconnect, reason, protocol} ->
+        reconnect_or_stop(:disconnect, reason, protocol, s)
+    end
+  end
+
+  @doc false
+  def handle_call({name, _opts}, _from, %{replication_started: true} = s)
```
josevalim, line 426: I think you can move this up and remove the duplication with a similar clause above?

josevalim:
This is looking beautiful @greg-rychlewski! I have added one important comment and three small nits.

lib/postgrex/replication.ex
greg-rychlewski: Thank you, good point!. I added this in with some extra details:

- if the slot is temporary then it also must be recreated on restart
- added a warning about duplicate WAL messages on restart
- If replication can't be restarted then I raise an error, same as what was there before. Since I believe in this case the error is unexpected and requires investigation.

greg-rychlewski: Actually this is trickier than I originally thought because you can create a slot that you're not starting replication on. 

What do you think about not trying to recreate the slot and putting a note in the doc stating that replication on temporary slots can't be automatically restarted?

lib/postgrex/replication.ex
```diff
@@ -411,32 +511,61 @@ defmodule Postgrex.Replication do
...
+      {_error, reason, _protocol} ->
+        raise reason
+    end
+  end
+
+  defp maybe_restart_replication(s), do: {:ok, s}
 
-  defp commands(opts) do
+  ## Queries
+  defp(command(:create_slot, opts)) do
```
josevalim, line 531:
```suggestion
  defp command(:create_slot, opts) do
```

lib/postgrex/replication.ex
```diff
@@ -148,13 +142,17 @@ defmodule Postgrex.Replication do
   defstruct protocol: nil,
             state: nil,
             auto_reconnect: false,
-            reconnect_backoff: 500
+            reconnect_backoff: 500,
+            replication_started: false,
+            replication_opts: nil
```
josevalim, line 147: We don't need both `replication_started` and `replication_opts`. If `replication_opts` is nil, then replication has not started. :)

lib/postgrex/replication.ex
josevalim: > What do you think about not trying to recreate the slot and putting a note in the doc stating that replication on temporary slots can't be automatically restarted?

Sounds like a good way to go for now!

greg-rychlewski:
@josevalim Thanks for taking time to review. I believe I got everything you mentioned.

josevalim:

PR #580, opened 2021-12-07
Replication: return results from simple queries + add more simple queries

greg-rychlewski:
This PR does the following:
- adds the ability to decode the rows returned by the simple queries executed on a replication connection
- returns the results from the existing simple queries, such as `create_slot` and `drop_slot` 
- adds the new commands `show`, `identify_server`, `timeline_history` and returns their results

### DIFF
--- a/lib/postgrex/replication.ex
+++ b/lib/postgrex/replication.ex
@@ -418,7 +418,7 @@ defmodule Postgrex.Replication do
   end
 
   @doc """
-  Returns timeline history file for the specified timeline ID.
+  Returns the timeline history file for the specified timeline ID.
 
   Once replication has begun, no other commands can be given and this
   function will return `{:error, :replication_started}`.
--- a/lib/postgrex/replication.ex
+++ b/lib/postgrex/replication.ex
@@ -151,7 +151,14 @@ defmodule Postgrex.Replication do
   @type state :: term
   @type copy :: binary
   @timeout 5000
-  @commands [:create_slot, :drop_slot, :start_replication, :show]
+  @commands [
+    :create_slot,
+    :drop_slot,
+    :start_replication,
+    :show,
+    :identify_system,
+    :timeline_history
+  ]
 
   @doc """
   Callback for process initialization.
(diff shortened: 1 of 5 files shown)


greg-rychlewski:
Converting to draft. I didn't realize the values from simple queries are all text. Need to change some things.

greg-rychlewski:
All fixed.

lib/postgrex/replication.ex
```diff
@@ -359,6 +370,77 @@ defmodule Postgrex.Replication do
...
+      Defaults to `5000`.
+
+  To better understand the meaning of this command,
+  [see PostgreSQL replication docs](https://www.postgresql.org/docs/14/protocol-replication.html).
+  """
+  @spec show(server, String.t(), Keyword.t()) ::
+          {:ok, Postgrex.Result.t()}
+          | {:error, Postgrex.Error.t()}
+          | {:error, :replication_started}
+  def show(pid, name, opts \\ []) do
```
josevalim, line 391: Should we check `name` is a string?

lib/postgrex/replication.ex
```diff
@@ -359,6 +370,77 @@ defmodule Postgrex.Replication do
...
+      Defaults to `5000`.
+
+  To better understand the meaning of this command,
+  [see PostgreSQL replication docs](https://www.postgresql.org/docs/14/protocol-replication.html).
+  """
+  @spec timeline_history(server, String.t(), Keyword.t()) ::
+          {:ok, Postgrex.Result.t()}
+          | {:error, Postgrex.Error.t()}
+          | {:error, :replication_started}
+  def timeline_history(pid, timeline_id, opts \\ []) do
```
josevalim, line 438: Same here, check that timeline_id is a string?

test/replication_test.exs
```diff
@@ -122,9 +130,23 @@ defmodule ReplicationTest do
...
+    {:ok, %Postgrex.Result{} = result} = PR.identify_system(context.repl)
+    assert result.num_rows == 1
+    refute is_nil(result.columns)
+    refute is_nil(result.rows)
+  end
+
+  test "show returns values", context do
+    {:ok, %Postgrex.Result{} = result} = PR.show(context.repl, "SERVER_VERSION")
+    assert result.num_rows == 1
+    refute is_nil(result.columns)
```
josevalim, line 143: Can  we assert on the column name? It should be static, right? For the rows, we can probably do `assert [[_]] = result.rows`?

test/replication_test.exs
```diff
@@ -122,9 +130,23 @@ defmodule ReplicationTest do
     assert Exception.message(error) =~ "replication slot \"postgrex_example\" is active for PID"
   end
 
+  test "identify system returns values", context do
+    {:ok, %Postgrex.Result{} = result} = PR.identify_system(context.repl)
+    assert result.num_rows == 1
+    refute is_nil(result.columns)
```
josevalim, line 136: Can  we assert on the column name? It should be static, right? For the rows, we can probably do `assert [[_]] = result.rows`?

josevalim:
Looks great! 🎉  I have added only some minor comments.

greg-rychlewski:
Thanks! I double checked that the column names are consistent across versions for the output of `create_slot`, 'show` and `identify_system` and they are.

josevalim:

PR #582, opened 2021-12-14
Replication: copy tables

greg-rychlewski:
This PR introduces 2 new functions: `copy_table/3` and `publication_tables/3`. 

It also introduces back pressure for the replication command because this was needed for `copy_table/3` and could be applied to `start_replication/3` for free.

**Motivation**

Would like to eventually expose the tools needed to simulate the `CREATE SUBSCRIPTION` command from Postgres. These 2 functions would be critical to providing the initial snapshot functionality from that command. 

The idea behind Postgres's initial snapshot functionality can be summarized by this excerpt from the `copy_table/3` doc:

```
... a single main replication process alongside secondary replication processes for each table. 
An individual secondary process can be used to copy the initial snapshot for a given table and
then stream replication messages for that table until it is caught up to the main process. Once 
caught up, the secondary process can be terminated and control of the table given back to the 
main process. This strategy minimizes the amount of time a transaction must hold onto old 
transaction IDs as well as the amount of time a slot must hold onto old WAL files.
```

**References**

  - https://www.postgresql.fastware.com/blog/logical-replication-tablesync-workers
  - https://github.com/postgres/postgres/blob/e997a0c642860a96df0151cbeccfecbdf0450d08/src/backend/replication/logical/tablesync.c
  - https://www.postgresql.org/docs/current/logical-replication-architecture.html
  - https://www.postgresql.org/docs/current/sql-createsubscription.html

### DIFF
--- a/lib/postgrex/replication.ex
+++ b/lib/postgrex/replication.ex
@@ -144,7 +144,8 @@ defmodule Postgrex.Replication do
             state: nil,
             auto_reconnect: false,
             reconnect_backoff: 500,
-            replication_opts: nil
+            repl_opts: nil,
+            copy_opts: nil
 
   ## PUBLIC API ##
 
@@ -154,13 +155,19 @@ defmodule Postgrex.Replication do
   @timeout 5000
   @max_lsn_component_size 8
   @max_uint64 18_446_744_073_709_551_615
-  @commands [
+  @max_copy_messages 500
+  @copy_done :copy_done
+  @slot_name_prefix "pg_"
+  @slot_name_rand_bytes 18
+  @public_commands [
     :create_slot,
     :drop_slot,
     :start_replication,
     :show,
     :identify_system,
-    :timeline_history
+    :timeline_history,
+    :copy_table,
+    :publication_tables
   ]
 
   @doc """
@@ -171,12 +178,14 @@ defmodule Postgrex.Replication do
   @callback init(term) :: {:ok, state}
 
   @doc """
-  Receives `binary` replication messages.
+  Receives `binary` messages during replication and `Postgrex.Result.t()` messages during table copying.
 
-  The format of the messages are described [under the START_REPLICATION
+  The format of the row values in the copy messages are `text`.
+
+  The format of the replication messages are described [under the START_REPLICATION
   section in PostgreSQL docs](https://www.postgresql.org/docs/14/protocol-replication.html).
 
-  Some messages require explicit acknowledgement, which can be done
+  Some replication messages require explicit acknowledgement, which can be done
   by returning a list of binaries according to the replication protocol.
   """
   @callback handle_data(binary, state) :: {:noreply, [copy], state}
@@ -274,8 +283,8 @@ defmodule Postgrex.Replication do
 
   By default, PostgreSQL includes the `pgoutput` plugin.
 
-  Once replication has begun, no other commands can be given and this
-  function will return `{:error, :replication_started}`.
+  This function will return `{:error, :stream_in_progress}` while a table
+  is being copied or after replication has begun.
 
   ## Options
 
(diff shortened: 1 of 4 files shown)


lib/postgrex/replication.ex
```diff
@@ -672,9 +853,97 @@ defmodule Postgrex.Replication do
...
+    }
+  end
+
+  defp table_columns(table_name, protocol) do
+    show_statement = command(:show, name: "server_version_num")
+
+    with {:ok, show_result, protocol} <- Protocol.handle_simple(show_statement, protocol),
+         %Postgrex.Result{rows: [[version]]} <- show_result,
+         columns_statement_opts <- [server_version: version, table_name: table_name],
+         columns_statement <- command(:table_columns, columns_statement_opts),
```
josevalim, line 916:
```suggestion
         %Postgrex.Result{rows: [[version]]} = show_result,
         columns_statement_opts = [server_version: version, table_name: table_name],
         columns_statement = command(:table_columns, columns_statement_opts),
```

josevalim:
Thank you @greg-rychlewski! Awesome work as usual!

Quick questions:

1. Should we maybe introduce `handle_copy` callback for the copy operations

2. Could we achieve similar behaviour by using a regular Postgrex pool with the stream operation to do the copying? What are the pros? What are the cons?

greg-rychlewski:
Thanks @josevalim! 

Here are my thoughts:

> Should we maybe introduce handle_copy callback for the copy operations

I like the idea of separating them and it could be done pretty easily by pattern matching `handle_info` on `copy_opts`. It makes the code more readable imo, having `handle_copy` and `handle_data` private functions/callbacks instead of pattern matching several times on `handle_data`. Also users shouldn't be sending replies for copy messages so the new callback could have that restriction.

>Could we achieve similar behaviour by using a regular Postgrex pool with the stream operation to do the copying? What are the pros? What are the cons?

I think this could potentially be a good option when the user just wants to copy all the tables and then start replication on everything once they're all done. The main benefits I see are that we would be reducing the complexity of the Replication/Protocol modules and reusing already proven copying/transaction functionality. 

The main drawback I see is not being able to use the copying connection to also do replication. And I think this becomes a deal breaker when implementing logical replication on large databases with high amounts of traffic. 

Using the regular Postgrex connection pool, would have to do something like this:

  - Start a replication process and use it to create a slot so that WAL is retained until copying is done.
  - Have the Postgrex connection pool complete the copying.
  - Start replication from before the copying started to ensure all the changes that happened during copying are synced.

This has the following disadvantages:

  - Potential for uncontrolled WAL build up. The worst consequences being running out of disk space or reaching [wal_keep_size](https://www.postgresql.org/docs/current/runtime-config-replication.html#GUC-WAL-KEEP-SIZE), causing the replication connection to close.
  - Potential for large replication lag since a single connection will be responsible for catching up all the tables.

Using the replication process pool, you could do this (similar to how Postgres `CREATE SUBSCRIPTION` works):

 - Start a main replication process and begin replication immediately so that WAL is not keep around unnecessarily. Replication messages are initially filtered until the table copying is completed.
 - Create a pool of replication processes, each one responsible for a single table at at time. 
 - The replication process for a single table will create a slot so that WAL is retained while copying the table.
 - Once the table has been copied, replication starts (filtering for the table of interest) until it is caught up to the LSN of the main process. Then this process can be killed and the main replication process can start acting on the messages for this table.

This has the following advantages:

  - Table changes that happened during copying are synced in parallel instead of sequentially.
  - The amount of WAL retained is dependent on the time it would take your biggest table to copy as opposed to the time it would take all tables to copy.

josevalim:
Alright, thank you! Let's do the `handle_copy` change and ship it!

greg-rychlewski:
Thanks 😃 ! All done.

josevalim:

PR #583, opened 2021-12-16
Replication: reconnect temp slots

greg-rychlewski:
This PR adds the ability to resume replication on temporary slots, provided the user passes the `plugin` option to `start_replication/3`. Also added some tests for reconnections.

### DIFF
--- a/lib/postgrex/replication.ex
+++ b/lib/postgrex/replication.ex
@@ -594,7 +594,7 @@ defmodule Postgrex.Replication do
           | {:error, :stream_in_progress}
   def copy_table(pid, table_name, slot_name, plugin, opts \\ []) do
     {timeout, opts} = Keyword.pop(opts, :timeout, @timeout)
-    opts = opts |> Keyword.put(:snapshot, :use)
+    opts = opts |> Keyword.put(:snapshot, :use) |> Keyword.put_new(:temporary, true)
     opts = [table_name: table_name, slot_name: slot_name, plugin: plugin] ++ opts
     call(pid, {:copy_table, opts}, timeout)
   end
--- a/lib/postgrex/replication.ex
+++ b/lib/postgrex/replication.ex
@@ -357,12 +357,12 @@ defmodule Postgrex.Replication do
   Starts logical replication on the given slot.
 
   If the slot's plugin requires additional options, the `plugin_opts`
-  option is required.
+  option must be specified.
 
   If the connection was started with `auto_reconnect` set to `true`, then
   replication will automatically restart with the options passed into this
   function. You must ensure your system will not be affected by receiving
-  duplicate WAL updates. Reconnecting temporary slots requires the `plugin`
+  duplicate WAL updates. Restarting temporary slots requires the `plugin`
   option in order to recreate the slot.
 
   This function will return `{:error, :stream_in_progress}` while a table
--- a/lib/postgrex/replication.ex
+++ b/lib/postgrex/replication.ex
@@ -314,9 +314,7 @@ defmodule Postgrex.Replication do
   [see PostgreSQL replication docs](https://www.postgresql.org/docs/14/protocol-replication.html).
   """
   @spec create_slot(server, String.t(), atom(), Keyword.t()) ::
-          {:ok, Postgrex.Result.t()}
-          | {:error, Postgrex.Error.t()}
-          | {:error, :stream_in_progress}
+          {:ok, Postgrex.Result.t()} | {:error, Postgrex.Error.t() | :stream_in_progress}
   def create_slot(pid, slot_name, plugin, opts \\ []) do
     opts = [slot_name: slot_name, plugin: plugin] ++ opts
     {timeout, opts} = Keyword.pop(opts, :timeout, @timeout)
@@ -342,9 +340,7 @@ defmodule Postgrex.Replication do
   [see PostgreSQL replication docs](https://www.postgresql.org/docs/14/protocol-replication.html).
   """
   @spec drop_slot(server, String.t(), Keyword.t()) ::
-          {:ok, Postgrex.Result.t()}
-          | {:error, Postgrex.Error.t()}
-          | {:error, :stream_in_progress}
+          {:ok, Postgrex.Result.t()} | {:error, Postgrex.Error.t() | :stream_in_progress}
   def drop_slot(pid, slot_name, opts \\ []) do
     opts = [slot_name: slot_name] ++ opts
     {timeout, opts} = Keyword.pop(opts, :timeout, @timeout)
@@ -367,7 +363,7 @@ defmodule Postgrex.Replication do
   with a temporary slot, use the `:create_temporary_slot` option instead. This
   provides the information needed to recreate the slot.
 
-  If replication is sucessfully started, this function will return `:ok` when
+  If replication is sucessfully started, this function will return `{:ok, nil}` when
   `:slot_name` is provided or `{:ok, Postgrex.Result()}` when `:create_temporary_slot`
   is provided. `{:ok, Postgrex.Result()}` is the result from creating the temporary
   slot and is identical to what is returned by `create_slot/4`.
@@ -408,32 +404,26 @@ defmodule Postgrex.Replication do
   [see PostgreSQL replication docs](https://www.postgresql.org/docs/14/protocol-replication.html).
   """
   @spec start_replication(server, Keyword.t()) ::
-          :ok
-          | {:ok, Postgrex.Result.t()}
-          | {:error, Postgrex.Error.t()}
-          | {:error, :stream_in_progress}
+          {:ok, Postgrex.Result.t() | nil} | {:error, Postgrex.Error.t() | :stream_in_progress}
   def start_replication(pid, opts \\ []) do
     opts =
-      case Keyword.has_key?(opts, :slot_name) do
-        true ->
+      cond do
+        Keyword.has_key?(opts, :slot_name) ->
           Keyword.drop(opts, [:create_temporary_slot])
 
-        false ->
-          case Keyword.has_key?(opts, :create_temporary_slot) do
-            true ->
-              {slot_opts, opts} = Keyword.pop(opts, :create_temporary_slot)
-              slot_opts = Keyword.put(slot_opts, :temporary, true)
+        Keyword.has_key?(opts, :create_temporary_slot) ->
+          {slot_opts, opts} = Keyword.pop(opts, :create_temporary_slot)
+          slot_opts = Keyword.put(slot_opts, :temporary, true)
 
-              if valid_slot_options?(slot_opts) do
-                slot_opts ++ opts
-              else
-                raise ArgumentError,
-                      "expected :slot_name, :plugin and :snapshot to be provided in :create_temporary_slot"
-              end
-
-            false ->
-              raise ArgumentError, "expected one of :slot_name or :create_temporary slot"
+          if valid_slot_options?(slot_opts) do
+            slot_opts ++ opts
+          else
+            raise ArgumentError,
+                  "expected :slot_name, :plugin and :snapshot to be provided in :create_temporary_slot"
           end
+
+        true ->
+          raise ArgumentError, "expected one of :slot_name or :create_temporary slot"
       end
 
     {timeout, opts} = Keyword.pop(opts, :timeout, @timeout)
@@ -455,9 +445,7 @@ defmodule Postgrex.Replication do
   [see PostgreSQL replication docs](https://www.postgresql.org/docs/14/protocol-replication.html).
   """
   @spec show(server, String.t(), Keyword.t()) ::
-          {:ok, Postgrex.Result.t()}
-          | {:error, Postgrex.Error.t()}
-          | {:error, :stream_in_progress}
+          {:ok, Postgrex.Result.t()} | {:error, Postgrex.Error.t() | :stream_in_progress}
   def show(pid, name, opts \\ []) when is_binary(name) do
     opts = [name: name] ++ opts
     {timeout, opts} = Keyword.pop(opts, :timeout, @timeout)
@@ -479,9 +467,7 @@ defmodule Postgrex.Replication do
   [see PostgreSQL replication docs](https://www.postgresql.org/docs/14/protocol-replication.html).
   """
   @spec identify_system(server, Keyword.t()) ::
-          {:ok, Postgrex.Result.t()}
-          | {:error, Postgrex.Error.t()}
-          | {:error, :stream_in_progress}
+          {:ok, Postgrex.Result.t()} | {:error, Postgrex.Error.t() | :stream_in_progress}
   def identify_system(pid, opts \\ []) do
     {timeout, opts} = Keyword.pop(opts, :timeout, @timeout)
     call(pid, {:identify_system, opts}, timeout)
@@ -502,9 +488,7 @@ defmodule Postgrex.Replication do
   [see PostgreSQL replication docs](https://www.postgresql.org/docs/14/protocol-replication.html).
   """
   @spec timeline_history(server, String.t(), Keyword.t()) ::
-          {:ok, Postgrex.Result.t()}
-          | {:error, Postgrex.Error.t()}
-          | {:error, :stream_in_progress}
+          {:ok, Postgrex.Result.t()} | {:error, Postgrex.Error.t() | :stream_in_progress}
   def timeline_history(pid, timeline_id, opts \\ []) when is_binary(timeline_id) do
     opts = [timeline_id: timeline_id] ++ opts
     {timeout, opts} = Keyword.pop(opts, :timeout, @timeout)
@@ -524,9 +508,7 @@ defmodule Postgrex.Replication do
       Defaults to `5000`.
   """
   @spec publication_tables(server, String.t(), Keyword.t()) ::
-          {:ok, Postgrex.Result.t()}
-          | {:error, Postgrex.Error.t()}
-          | {:error, :stream_in_progress}
+          {:ok, Postgrex.Result.t()} | {:error, Postgrex.Error.t() | :stream_in_progress}
   def publication_tables(pid, publication_name, opts \\ []) when is_binary(publication_name) do
     opts = [publication_name: publication_name] ++ opts
     {timeout, opts} = Keyword.pop(opts, :timeout, @timeout)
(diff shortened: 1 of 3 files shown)


lib/postgrex/replication.ex
```diff
@@ -354,28 +354,35 @@ defmodule Postgrex.Replication do
...
       each, separated by a slash. e.g. `1/F73E0220`.
       Defaults to `0/0`.
 
+    * `:plugin_opts` - It must be a keyword list and is used to configure
+      the output plugin assigned to the given slot.
+
+    * `:plugin` - Used to recreate temporary slots when `auto_reconnect = true`.
+      If not provided, the slot cannot be recreated and an error will be raised
+      during reconnection. Recreated slots will be made with `temporary = true`
+      and `snapshot = :noexport`. See `create_slot/4` for more details.
```
josevalim, line 384: I am not sure if we should couple the plugin with the temporary option. Someone may want to use plugin for other reasons without setting temporary or snapshot on reconnect?

test/replication_test.exs
```diff
@@ -289,6 +294,54 @@ defmodule ReplicationTest do
     end
   end
 
+  describe "auto-reconnect" do
```
josevalim, line 297: 💯

lib/postgrex/replication.ex
greg-rychlewski: Now that you mention this, I am realizing we can reconnect temporary slots without  adding new options using the results from [command(:slot_state, opts)](https://github.com/elixir-ecto/postgrex/blob/master/lib/postgrex/replication.ex#L1017). This will tell us the plugin and whether the slot is temporary or not, given the slot name. 

If you like this route, there would be a couple questions:

1. We can't receive the `:snapshot` value from this command. I'm not sure a user will be able to act on an exported snapshot when reconnecting, so I think it could make sense to set this as `:noexport`.
2. Should permanent slots be recreated on reconnect or just temporary ones? I'm thinking they shouldn't because if a permanent slot is disappearing between disconnect and reconnect probably the user would want to know. 

Anyway, those are just my initial thoughts. Please let me know what you think.

josevalim: My suggestion would be this:

1. If `temporary: true`, we recreate. If we also need the plugin option, you should raise if temporary+auto_reconnect are given and there is no plugin. If a snapshot is also required, then we will request it to be added too.

2. If `temporary: false`, we don't recreate.

This way the user intent is always clear. WDYT?

greg-rychlewski: Perfect! I will make the change.

greg-rychlewski: I made the changes but just wanted to double check with you a couple things:

1. I validate before starting replication just so the user doesn't find a surprise when they reconnect after a few days. Wanted to make sure that is in line with how you would normally do things.
2. Is there a good way to test that an exception is raised inside the connection process? I tried a few things, `assert_raise` and trapping exits but couldn't find a way to get it to work.

josevalim: You should be able to trap exits and then do assert_receive on the EXIT message that gets delivered to your inbox.

regarding the raising, perfect.

greg-rychlewski: Ah I was missing `catch_exit` when I tried to trap exits before. Working now. Thank you for the help.

lib/postgrex/replication.ex
```diff
@@ -542,7 +556,7 @@ defmodule Postgrex.Replication do
       it is not needed anymore. See `create_slot/4` for more details.
       Defaults to `true`.
 
-    * `:slot_opts` - Keyword list of options to create the slot with.
+    * `:slot_opts` - Keyword list of options used to create the slot.
       Only the keys `:slot_name`, `:temporary` and `:plugin` are allowed.
       By default, a temporary slot with the `:pgoutput` plugin is created.
       See `create_slot/4` for more details.
```
josevalim, line 562: Isn't it confusing that we have different defaults here? Does `copy_table` only make sense for temporarily slots?

lib/postgrex/replication.ex
```diff
@@ -691,13 +705,20 @@ defmodule Postgrex.Replication do
...
+    with :ok <- validate_reconnect_options(s.auto_reconnect, temporary, opts),
+         {:ok, protocol} <- Protocol.handle_replication(statement, s.protocol),
          {:ok, protocol} <- Protocol.checkin(protocol) do
       s = %{s | protocol: protocol, repl_opts: opts}
       {:reply, :ok, s}
     else
+      :error ->
+        # If auto_reconnect is enabled and the temporary option is given,
+        # raise an error if any of the keys in temporary are missing.
+        raise ArgumentError, "expected :plugin and :snapshot to be given in the :temporary option"
```
josevalim, line 720: We should validate this in the client, better than crashing in the process later on?

josevalim: Oh, ignore me. We only check this here because we only try to create this slot here if it is temporary. Hrm...

I am thinking that maybe we should change the `start_replication` API slightly. What if instead of `:temporary`, we call it `create_temporary_slot: [...]`, and then it will be applied both on `start_replication` and on auto-reconnect, making sure both the initial behaviour and the reconnect behaviour are the same. The idea is that, if the slot is temporary, it is always automatically started by `start_replication` for you. You don't have to do it yourself. WDYT?

Sorry about the back and forth, but it is a bit tricky to figure out all of the API details.

lib/postgrex/replication.ex
```diff
@@ -566,12 +580,12 @@ defmodule Postgrex.Replication do
     slot_opts =
       slot_opts
       |> Keyword.take([:slot_name, :temporary, :plugin])
-      |> Keyword.put_new(:plugin, :pgoutput)
       |> Keyword.put_new(:slot_name, copy_table_slot_name())
+      |> Keyword.put_new(:plugin, :pgoutput)
+      |> Keyword.put_new(:temporary, true)
```
josevalim, line 585: I think those options (plugin/temporary/snapshot) are not documented on `copy_table`. Maybe we want to nest them under `temporary` (or under `create_temporary_slot`, as per the previous comment)?

lib/postgrex/replication.ex
greg-rychlewski: > I am thinking that maybe we should change the start_replication API slightly. What if instead of :temporary, we call it create_temporary_slot: [...], and then it will be applied both on start_replication and on auto-reconnect, making sure both the initial behaviour and the reconnect behaviour are the same. The idea is that, if the slot is temporary, it is always automatically started by start_replication for you. You don't have to do it yourself. WDYT?

I like it. So the idea would be, for both `start_replication` and `copy table`, that you either provide your own permanent slot or provide the options needed to create/recreate a temporary slot. There would be a few other changes to consider in this case:

1. Might want to remove the `temporary` option from `create_slot` because any temporary slot created won't work as expected during auto-reconnect when passed into these commands. It could be considered (and possibly renamed to) `create_permanent_slot`?  If this change isn't made then I think it gets a bit ugly when a user provides a temporary slot. Either:
    - The commands crash upon reconnect
    - We need to make another option for the user to provide the reconnect options
    - We have to get the reconnect options from the `slot_state` query, which wouldn't have the `snapshot` option.

2. `start_replication` and `copy_table` must have 2 options: `slot_name` and `create_temp_slot` and one of them is required but not both. If both are provided then raise?

>Sorry about the back and forth, but it is a bit tricky to figure out all of the API details.

No worries, I agree lol. I second guessed many things while I was writing these. I'm happy to try to make the API better.

lib/postgrex/replication.ex
greg-rychlewski: `Does copy_table only make sense for temporarily slots?`

No you're completely right, this is confusing. Addressed this in your other comment.

lib/postgrex/replication.ex
greg-rychlewski: That's a good point. Addressed in your other comment.

josevalim:
> Might want to remove the temporary option from create_slot because any temporary slot created won't work as expected during auto-reconnect when passed into these commands

I would say it is fine to keep it as part of `create_slot` but we should document that, if they want temporary slots, they should consider having it automatically created under `copy_tables` and `start_replication`, especially if they are using `autoreconnect`. This way we keep the API complete but provide proper guidance.

> start_replication and copy_table must have 2 options: slot_name (to use an existing permanent slot) and create_temp_slot and one of them is required but not both. If both are provided then raise?

Sounds good to me!

greg-rychlewski:
There's one other thing I thought of for `copy_table`. The `drop_slot` option will be more confusing with the new API because we shouldn't be automatically dropping the slots they bring themselves.

The idea behind this option was that copying is temporary and there might be some people that just want to know the `consistent_point` from the slot so they know where to start replication after copying. There are 2 options I see:

1. add `:drop` to the keyword `:create_temporary_slot` keyword list. 
2. Remove the command altogether and put a note that the user should drop the slot if they don't want it once copying is done.

I'm leaning towards 2 because it would be nice if `create_temporary_slot` was simple and consistent between the 2 commands. And it seems like the idea behind this module is to give people low level control over everything, so don't necessarily need to baby them with the ability to automatically drop the slot. Please let me know what you think!

greg-rychlewski:
Sorry I actually forgot about something important. I don't think it will make sense to have the user provide their own slot for `copy_table`. The reason is because the slot is used to create a snapshot just before copying so that you can resume replication from its `consistent_point` once copying is done.

We could make the arguments to `copy_table` similar to the ones for `create_slot`, with the addition of `table_name`. i.e. `table_name`, `slot_name`, `plugin` are required and `temporary` and `drop_slot` are optional with certain defaults?

greg-rychlewski:
Apologies for the multiple comments. I just had one last idea that I think simplifies the `copy_table` command. Similar to what I said in my last comment: 

>We could make the arguments to copy_table similar to the ones for create_slot, with the addition of table_name. i.e. table_name, slot_name, plugin are required and temporary and drop_slot are optional with certain defaults?

But remove the `drop_slot` option entirely and limit the scope of this function to being used when you want to start replication on the same slot that you are copying. And if that is not required, advise people to use a regular Postgrex connection. I try to give a simplified version of [what we talked about in the last PR](https://github.com/elixir-ecto/postgrex/pull/582#issuecomment-994819532) to help people understand where this would be useful. And later on could write a guide with a real example I create. If you think that would be a good addition to this module.

josevalim:
Simpler is good! I usually say, we can always add stuff later, removing is the hard part. So less is better!

greg-rychlewski:
Thanks for all the feedback. I completed the changes we talked about. I'll summarize here:

1. Refactor api for `start_replication`
   - Provide `slot_name` and `create_temporary_slot` options. The former is meant for existing slots while the latter is a keyword list used to create temporary slots for the user. This is to ensure we have the information needed to recreate the slot on auto_reconnect. I put a note in the docs about the dangers of using auto_reconnect and supplying a temporary `slot_name`
2. Refactor api for `copy_table` so that it is similar to `create_slot` but with an additional `table_name` argument. This replaces the way it was before where the user didn't have to give their intent and I picked defaults for the slot options. Also removed `drop_slot` and altered the description of the command to say that it's mean to be used for more complicated scenarios and that a regular postgrex connection might suffice if their situation is simpler.

lib/postgrex/replication.ex
```diff
@@ -386,10 +407,35 @@ defmodule Postgrex.Replication do
...
+                slot_opts ++ opts
+              else
+                raise ArgumentError,
+                      "expected :slot_name, :plugin and :snapshot to be provided in :create_temporary_slot"
+              end
+
+            false ->
+              raise ArgumentError, "expected one of :slot_name or :create_temporary slot"
+          end
+      end
```
josevalim, line 437:
```suggestion
      cond do
        Keyword.has_key?(opts, :slot_name) ->
          Keyword.drop(opts, [:create_temporary_slot])

        Keyword.has_key?(opts, :create_temporary_slot) ->
          {slot_opts, opts} = Keyword.pop(opts, :create_temporary_slot)
          slot_opts = Keyword.put(slot_opts, :temporary, true)

          if valid_slot_options?(slot_opts) do
            slot_opts ++ opts
          else
            raise ArgumentError,
                  "expected :slot_name, :plugin and :snapshot to be provided in :create_temporary_slot"
          end

        true ->
          raise ArgumentError, "expected one of :slot_name or :create_temporary slot"
      end
```

lib/postgrex/replication.ex
```diff
@@ -693,10 +717,11 @@ defmodule Postgrex.Replication do
   def handle_call({:start_replication, opts}, _from, %{repl_opts: nil, copy_opts: nil} = s) do
     statement = command(:start_replication, opts)
 
-    with {:ok, protocol} <- Protocol.handle_replication(statement, s.protocol),
+    with {:ok, slot, protocol} <- maybe_create_slot(opts, s.protocol),
+         {:ok, protocol} <- Protocol.handle_replication(statement, protocol),
          {:ok, protocol} <- Protocol.checkin(protocol) do
-      s = %{s | protocol: protocol, repl_opts: opts}
-      {:reply, :ok, s}
+      reply = (slot && {:ok, slot}) || :ok
```
josevalim, line 723: I wonder if we should do `{:ok, slot}` and then we return `{:ok, nil}` if there is no slot?

lib/postgrex/replication.ex
```diff
@@ -927,25 +954,12 @@ defmodule Postgrex.Replication do
...
-      if state_result.rows == [[]] do
-        create_statement = command(:create_slot, opts)
-        Protocol.handle_simple(create_statement, protocol)
-      else
-        {:ok, nil, protocol}
-      end
+  defp valid_slot_options?(opts) do
+    with true <- Keyword.has_key?(opts, :slot_name),
+         true <- Keyword.has_key?(opts, :plugin),
+         true <- Keyword.has_key?(opts, :temporary),
```
josevalim, line 960: Don't need to check for temporary as you are the one adding it. :)

```suggestion
```

lib/postgrex/replication.ex
```diff
@@ -488,65 +534,52 @@ defmodule Postgrex.Replication do
...
-
-  If the connection was started with `auto_reconnect` set to `true`, then
+  Streams a copy of the given table.
+
+  This function is meant to aid in the initial synchronization of large
+  databases with high traffic. For simpler situations, it may be desirable
+  to use a non-replication connection to copy the table.
+
+  In addition to copying the table, a replication slot will be created using
+  the `slot_name`, `plugin` and `:temporary` values passed into this function.
```
josevalim, line 544:
```suggestion
  the `:slot_name`, `:plugin`, and `:temporary` values passed into this function.
```

lib/postgrex/replication.ex
```diff
@@ -488,65 +534,52 @@ defmodule Postgrex.Replication do
...
+  The replication slot is returned to the user so it can be used to catch up
+  with the changes that occur during copying. The slot's `:snapshot` value
+  is automatically set to `:use` to ensure the user can start replication where
+  the copying transaction left off.
```
josevalim, line 548:
```suggestion
  In addition to copying the table, a replication slot will be created using
  the `slot_name` and `plugin` arguments passed into this function. The slot
  is temporary by default but it can be changed via the `:temporary` option.
  The replication slot is returned to the user so it can be used to catch up
  with the changes that occur during copying. The slot's `:snapshot` value
  is automatically set to `:use` to ensure the user can start replication where
  the copying transaction left off.
```

josevalim (commented):
Just some minor comments left and we are good to go! :100:

lib/postgrex/replication.ex
greg-rychlewski: I think it's a good option. At least it seems more consistent to me: always return `{:ok, created_slot}` and some case no slot is created/returned.

greg-rychlewski:
Thanks! All done. When changing the type spec for `:ok` --> `{:ok, nil}` I noticed I was too verbose in all the specs, using `{:error, Postgrex.Error.t()} | {:error, :streaming_in_progress}` instead of `{:error, Postgrex.Error.t() | :streaming_in_progress}` so I updated those.

josevalim:

OctopusRage:
is copy_table deprecated now? im getting stuck how to do full table copy then continue the replication stream

OctopusRage:
sorry im still new in postgres replication.
can i do full snapshot by providing this `EXPORT_SNAPSHOT` at the end of query?
```
  def handle_connect(%__MODULE__{slot: slot} = state) do
    query = "CREATE_REPLICATION_SLOT #{slot} TEMPORARY LOGICAL pgoutput EXPORT_SNAPSHOT"

    Logger.debug("[create slot] query=#{query}")

    {:query, query, %{state | state: :create_slot}}
  end
  ```
  
  or do i need to do `pg_export_snapshot()` and then save the max txid? and continue with code above?

**Comment by greg-rychlewski** [comment]
That should work. And you can listen for the result using `handle_result`

**Comment by OctopusRage** [comment]
sry if im misunderstand something, 
so this is my code
```elixir

  @impl true
  def handle_connect(%__MODULE__{slot: slot} = state) do
    query = "CREATE_REPLICATION_SLOT #{slot} LOGICAL pgoutput EXPORT_SNAPSHOT"

    Logger.debug("[create slot] query=#{query}")

    {:query, query, %{state | state: :create_slot}}
  end

  @impl true
  def handle_result(
        [%Postgrex.Result{} | _] = x_result,
        %__MODULE__{state: :create_slot, publications: pubs} = state
      ) do
    opts = [proto_version: 1, publication_names: pubs]
    IO.inspect x_result
    

    query = "START_REPLICATION SLOT #{state.slot} LOGICAL 0/0 #{escape_options(opts)}"

    Logger.debug("[start streaming] query=#{query}")

    {:stream, query, [], %{state | state: :streaming}}
  end
```

`x_result` give me snapshotname result
```elixir
    [
      %Postgrex.Result{
        command: :create,
        columns: ["slot_name", "consistent_point", "snapshot_name", "output_plugin"],
        rows: [["articles_slot", "1/83165168", "00000007-000009E7-1", "pgoutput"]],
        num_rows: 1,
        connection_id: 155671,
        messages: []
      }
    ]
    
 ```
 then i try to use that snapshot name in other db connection using 
```
BEGIN ISOLATION LEVEL REPEATABLE READ;
SET TRANSACTION SNAPSHOT '00000007-000009E7-1';
``` 
then i got error
 `invalid snapshot identifier: "00000007-000009E7-1"
 where did i do wrong here?

**Comment by OctopusRage** [comment]
i finally make it work by pausing the process using Process.sleep in `handle_result`
```
def handle_result(
        [%Postgrex.Result{} | _] = x_result,
        %__MODULE__{state: :create_slot, publications: pubs} = state
      ) do
    opts = [proto_version: 1, publication_names: pubs]
    IO.inspect x_result

    query = "START_REPLICATION SLOT #{state.slot} LOGICAL 0/0 #{escape_options(opts)}"

    Logger.debug("[start streaming] query=#{query}")
    Process.sleep(10_000_000)

    {:stream, query, [], %{state | state: :streaming}}
  end
```
thx for helping

PR #586, opened 2021-12-22
Notification callback module

sorentwo:
I am opening this to get discussion and feedback going.

The goal is to make `Postgrex.Notifications` flexible enough to handle running simple arbitrary queries, in addition to the listen/unlisten/notify work it does currently. To make that possible without exposing private APIs, `Notifications` accepts a callback module with a default implementation that maintains the current functionality, but is extensible enough for people to implement their own notification logic.

#### Steps

- [x] Base notifications callback module
- [x] Simple query execution and handling
- [x] Listen/Unlisten/Notification abstraction
- [x] Extract SimpleConnection
- [x] Doc updates

### DIFF
--- a/lib/postgrex/notifications.ex
+++ b/lib/postgrex/notifications.ex
@@ -76,11 +76,7 @@ defmodule Postgrex.Notifications do
             protocol: nil,
             auto_reconnect: false,
             reconnect_backoff: 500,
-            connected: false,
             state: nil
-
-  # TODO: Return :eventually tuple when connected: false
-  # TODO: Handle reply on error/auto-connect
 
   ## PUBLIC API ##
 
@@ -96,31 +92,46 @@ defmodule Postgrex.Notifications do
   @callback init(term) :: {:ok, state}
 
   @doc """
-  Callback for state management after connection or reconnection.
+  Invoked after connecting.
+
+  This may be called multiple times if `:auto_reconnect` is true.
   """
-  @callback connect(state) :: {:ok, state}
+  @callback handle_connect(state) :: {:noreply, state} | {:query, query, state}
+
+  @doc """
+  Invoked after disconnection, regardless of `:auto_reconnect`.
+  """
+  @callback handle_disconnect(state) :: {:noreply, state}
 
   @doc """
   Callback for `call/3`.
+
+  Replies must be sent with `reply/2`.
   """
   @callback handle_call(term, GenServer.from(), state) ::
-              {:noreply, state} | {:reply, term, state} | {:query, query, state}
+              {:noreply, state}
+              | {:query, query, state}
 
   @doc """
-  Callback for unstructured messages.
+  Callback for `Kernel.send/2`.
   """
-  @callback handle_info(term, state) ::
-              {:noreply, state} | {:reply, term, state} | {:query, query, state}
+  @callback handle_info(term, state) :: {:noreply, state} | {:query, query, state}
 
   @doc """
   Callback for processing or relaying query results.
   """
-  @callback handle_result(Result.t(), state) :: {:noreply, state}
+  @callback handle_result(Postgrex.Result.t() | Postgrex.Error.t(), state) :: {:noreply, state}
 
   @doc """
   Callback for processing or relaying pubsub notifications.
   """
   @callback handle_notification(binary, binary, state) :: {:noreply, state}
+
+  @optional_callbacks handle_info: 2,
+                      handle_call: 3,
+                      handle_result: 2,
+                      handle_connect: 1,
+                      handle_disconnect: 1
 
   @doc """
   Replies to the given client.
@@ -267,7 +278,7 @@ defmodule Postgrex.Notifications do
 
         if opts[:sync_connect] do
           case connect(:init, state) do
-            {:ok, _, _} = ok -> ok
+            {:ok, _} = ok -> ok
             {:backoff, _, _} = backoff -> backoff
             {:stop, reason, _} -> {:stop, reason}
           end
@@ -278,7 +289,7 @@ defmodule Postgrex.Notifications do
   end
 
   @doc false
-  def connect(_, %{state: {mod, mod_state}} = s) do
+  def connect(_, %{state: {mod, mod_state}} = state) do
     opts =
       case Keyword.get(opts(), :configure) do
         {module, fun, args} -> apply(module, fun, [opts() | args])
@@ -288,23 +299,24 @@ defmodule Postgrex.Notifications do
 
     case Protocol.connect(opts) do
       {:ok, protocol} ->
-        # TODO: This needs to execute a query after connection
-        {:noreply, mod_state} = mod.connect(mod_state)
+        state = %{state | protocol: protocol}
 
-        {:ok, %{s | protocol: protocol, state: {mod, mod_state}}, s.idle_interval}
+        with {:noreply, state, _} <- maybe_handle(mod, :handle_connect, [mod_state], state) do
+          {:ok, state}
+        end
 
       {:error, reason} ->
-        if s.auto_reconnect do
-          {:backoff, s.reconnect_backoff, s}
+        if state.auto_reconnect do
+          {:backoff, state.reconnect_backoff, state}
         else
-          {:stop, reason, s}
+          {:stop, reason, state}
         end
     end
   end
 
   @doc false
   def handle_call(msg, from, %{state: {mod, mod_state}} = state) do
-    handle(mod, :handle_call, [msg, from, mod_state], state)
+    maybe_handle(mod, :handle_call, [msg, from, mod_state], state)
   end
 
   @doc false
@@ -325,22 +337,30 @@ defmodule Postgrex.Notifications do
       {:ok, protocol} ->
         {:noreply, %{state | protocol: protocol}, state.idle_interval}
 
+      {:unknown, protocol} ->
+        maybe_handle(mod, :handle_info, [msg, mod_state], %{state | protocol: protocol})
+
       {error, reason, protocol} ->
         reconnect_or_stop(error, reason, protocol, state)
     end
   end
 
   def handle_info(msg, %{state: {mod, mod_state}} = state) do
-    handle(mod, :handle_info, [msg, mod_state], state)
+    maybe_handle(mod, :handle_info, [msg, mod_state], state)
+  end
+
+  defp maybe_handle(mod, fun, args, state) do
+    if function_exported?(mod, fun, length(args)) do
+      handle(mod, fun, args, state)
+    else
+      {:noreply, state}
+    end
   end
 
   defp handle(mod, fun, args, state) do
     case apply(mod, fun, args) do
       {:noreply, mod_state} ->
         {:noreply, %{state | state: {mod, mod_state}}, state.idle_interval}
-
-      {:reply, reply, mod_state} ->
-        {:reply, reply, %{state | state: {mod, mod_state}}, state.idle_interval}
 
       {:query, query, mod_state} ->
         opts = [notify: &mod.handle_notification(&1, &2, mod_state)]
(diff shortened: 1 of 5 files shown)


lib/postgrex/notifications.ex
```diff
@@ -188,38 +214,39 @@ defmodule Postgrex.Notifications do
...
-    if idle_timeout do
-      require Logger
-
-      Logger.warn(
-        ":idle_timeout in Postgrex.Notifications is deprecated, " <>
-          "please use :idle_interval instead"
-      )
-    end
+  def init({mod, args, opts}) do
+    with {:ok, mod_state} <- mod.init(args) do
```
josevalim, line 218: I would explicitly check that it has to return {:error, _} or :ignore otherwise, instead of relying on connection.

lib/postgrex/notifications.ex
```diff
@@ -78,18 +78,43 @@ defmodule Postgrex.Notifications do
...
+  """
+  @callback handle_result(Result.t(), state) :: {:noreply, state}
 
   @doc false
   def child_spec(opts) do
-    %{
-      id: __MODULE__,
-      start: {__MODULE__, :start_link, [opts]}
-    }
+    opts =
```
v0idpwn, line 110: Perhaps this match can be done in the function head or with `case`? Don't see the point for `if match?`

v0idpwn (commented):
Looks great so far, just a minor comment :)

lib/postgrex/notifications.ex
sorentwo: Agreed, it is pretty silly. This is only a temporary change to get the tests working anyhow. Soon there will be two distinct modules with their own `child_spec/1`.

sorentwo:
The changes have progressed nicely and I was able to make a 95% functional custom notification
callback in Oban. Between my various `TODO` comments, failing test cases, and what I discovered with the new implementation, there are three outstanding issues:

1. How to execute queries on connect. That may not be a problem if we can send ourselves a message like `send(self(), :reconnect)`.
2. How to know if there is a connection. The server maintains a `connected` boolean that was formerly used to control whether listen/unlisten got an `:ok` or `:eventually` response. The callback module could track the status internally via the `connect` callback, but there isn't any place to flip that back to false on error/while waiting for a reconnect.
3. How to pass unhandled messages to the `handle_info` callback, as they're currently passed to `Protocol.handle_info/3`. This one seems the messiest.

Some possible solutions:

1. Change the `connect` spec back to `{:ok, state}` and send ourselves a message on connect.
2. Introduce a `disconnect` callback for pairity with `connect`.
3. Pass all messages through `Protocol.handle_info/3`, but pass along a function for when nothing matches (the `:unknown` case). That could look something like this:

```elixir
opts = [
   notify: &mod.handle_notification(&1, &2, mod_state),
   unknown: &mod.handle_info(&1, mod_state)
]

case Protocol.handle_info(msg, opts, protocol) do
```

Then the default `unknown` (not the best name) function would log the unknown message as it does now. WDYT?

lib/postgrex/notifications.ex
```diff
@@ -74,22 +74,87 @@ defmodule Postgrex.Notifications do
...
+
+  @doc """
+  Callback for processing or relaying query results.
+  """
+  @callback handle_result(Postgrex.Result.t() | Postgrex.Error.t(), state) :: {:noreply, state}
+
+  @doc """
+  Callback for processing or relaying pubsub notifications.
+  """
+  @callback handle_notification(binary, binary, state) :: {:noreply, state}
```
josevalim, line 128: I think the state in handle_notification is ignored and not given back, right? Perhaps the return type here should be `:ok`. Potentially even call the callback `notify` so people don't expect it to have the same return types as `handle_`.

josevalim:
This looks great to me! I think what we need to do next is to rename, test, and document the new API in isolation. My suggestion would be to introduce a module named `Postgrex.Simple` or `Postgrex.SimpleConnection`. We will document it can perform simple queries, including notifications, and `Postgrex.Notifications` is built on top of it. WDYT?

lib/postgrex/notifications.ex
```diff
@@ -285,107 +329,194 @@ defmodule Postgrex.Notifications do
...
+        with {:ok, result, protocol} <- Protocol.handle_simple(query, opts, state.protocol),
+             {:ok, protocol} <- Protocol.checkin(protocol) do
+          state = %{state | protocol: protocol}
+
+          handle(mod, :handle_result, [result, mod_state], from, state)
+        else
+          {error, reason, protocol} ->
+            from && reply(from, {__MODULE__, reason})
+
+            {:noreply, state, _} = maybe_handle(mod, :handle_disconnect, [mod_state], state)
```
josevalim, line 378: I think we should move this to `reconnect_or_stop` and invoke it only if auto_reconnect is set to true?

sorentwo: Hmm. I think it should run before any disconnect because it could be about more than changing the `mod_state`. For example, imagine you want some side-effect in the handler, e.g., notifying other processes that the connection is terminating.

sorentwo:
> We will document it can perform simple queries, including notifications, and `Postgrex.Notifications` is built on top of it. WDYT?

That's the plan 👍

lib/postgrex/notifications.ex
sorentwo: Good call. Changing it to `notify` and `:ok` 👍

sorentwo:
Alright, I've extracted `SimpleConnection` and tracked down a few fixes that the unit tests exposed. There is a slight issue with a `Notifications` test that expects an `{:eventually, _}` result but gets an error instead:

```
     ** (exit) exited in: Postgrex.SimpleConnection.call(#PID<0.265.0>, {:listen, "channel2"}, 5000)
         ** (EXIT) %DBConnection.ConnectionError{message: "tcp send: closed", reason: :error, severity: :error}
```

That's desirable for other calls, but it breaks the contract for `listen/unlisten`. Aside from that issue, this is only missing `SimpleConnection` docs.

lib/postgrex/simple_connection.ex
```diff
@@ -0,0 +1,274 @@
...
+
+          handle(mod, :handle_result, [result, mod_state], from, state)
+        else
+          {:error, %Postgrex.Error{} = error, protocol} ->
+            handle(mod, :handle_result, [error, mod_state], from, %{state | protocol: protocol})
+
+          {:disconnect, reason, protocol} ->
+            from && reply(from, {__MODULE__, reason})
+
+            reconnect_or_stop(:disconnect, reason, protocol, state)
```
greg-rychlewski, line 255:
```suggestion
          {:disconnect, reason, protocol} ->
            case reconnect_or_stop(:disconnect, reason, protocol, state) do
              {:stop, _, _} = stop ->
                stop

              {:connect, _, _} = connect ->
                from && reply(from, async_reply(mod_state, reason))
                connect
            end
        end
    end
  end

  defp async_reply(%{async_reply: reply} = mod_state, _reason), do: async_reply
  defp async_reply(_ = _mod_state, reason), do: {__MODULE__, reason}
```

@sorentwo I think you need something like this. It will fix your failing test and also it seems necessary for the notification module to function if the user tries to issue a listen command before the connection process handles the disconnect message from postgres.

josevalim: I think we should just remove the `from && ...` and document that, if auto_reconnect is true, you can fail and never get a reply unless you handle this case in `handle_disconnect`.

josevalim: Here is the approach I mentioned above implemented for replication: https://github.com/elixir-ecto/postgrex/commit/bf793fa20c8d7dfce1c5424ca5ea325d27e11e3d - I will push some docs soon.

lib/postgrex/notifications.ex
```diff
@@ -61,35 +61,21 @@ defmodule Postgrex.Notifications do
...
-  @type server :: GenServer.server()
+  @timeout 5000
 
   @doc false
   def child_spec(opts) do
-    %{
-      id: __MODULE__,
-      start: {__MODULE__, :start_link, [opts]}
-    }
+    %{id: __MODULE__, start: {SimpleConnection, :start_link, [__MODULE__, [], opts]}}
```
josevalim, line 78:
```suggestion
    %{id: __MODULE__, start: {__MODULE__, :start_link, [opts]}}
```

lib/postgrex/notifications.ex
```diff
@@ -187,205 +167,120 @@ defmodule Postgrex.Notifications do
...
-  end
-
-  defp reestablish_listener({ref, {channel, pid}}, {:ok, s, timeout}) do
-    case do_listen(channel, pid, ref, nil, s) do
-      {:noreply, s, _} -> {:cont, {:ok, s, timeout}}
-      error -> {:halt, error}
-    end
-  end
+        if map_size(state.listener_channels[channel]) == 0 do
+          state = update_in(state.listener_channels, &Map.delete(&1, channel))
```
josevalim, line 239: nitpick:

```suggestion
          {_, state} = pop_in(state.listener_channels[channel])
```

lib/postgrex/simple_connection.ex
greg-rychlewski: What do you think about changing the arity of `handle_disconnect` to 2 so it can receive whether or not the disconnect is a `:stop` or `:reconnect`? I think it might be necessary here so the Notifications module can reply with `:eventually` only on reconnect. But might be mistaken.

josevalim: @greg-rychlewski the module will always know if it is started with auto-reconnect or not, so it can figure it out if necessary. :)

lib/postgrex/notifications.ex
```diff
@@ -121,10 +107,7 @@ defmodule Postgrex.Notifications do
   """
   @spec start_link(Keyword.t()) :: {:ok, pid} | {:error, Postgrex.Error.t() | term}
   def start_link(opts) do
-    {server_opts, opts} = Keyword.split(opts, [:name])
-    opts = Keyword.put_new(opts, :sync_connect, true)
-    connection_opts = Postgrex.Utils.default_opts(opts)
-    Connection.start_link(__MODULE__, connection_opts, server_opts)
+    SimpleConnection.start_link(__MODULE__, [], opts)
```
greg-rychlewski, line 110:
```suggestion
    auto_reconnect = opts[:auto_reconnect] || false
    SimpleConnection.start_link(__MODULE__, [auto_reconnect: auto_reconnect], opts)
```

I think you need something like this and then to add `auto_reconnect` to the state so that on `handle_disconnect` you can reply with `{:eventually, ref}` if auto reconnect is true.

greg-rychlewski: Actually I don't think it's this simple because you only want `:eventually` on the listen command and not the regular queries.

sorentwo: The listen command already issues an eventually reply when it isn’t connected. Without the auto-reply in error the API remains intact and the tests pass.

greg-rychlewski: I think the current refactor might be missing something important the old module was handling. But you guys let me know what you think.

If you look at the [listener_query](https://github.com/elixir-ecto/postgrex/blob/master/lib/postgrex/notifications.ex#L341) function in the old module, it is issuing the `:eventually` reply in 2 place: 

1. if the state is saying the socket is disconnected
2. if the state doesn't know the socket is disconnect and the user tries to issue a listen command and gets a disconnect message

I think the test with the auto-reply was exposing this gap between the old and new modules. i.e. this might happen:

1. Postgres server disconnects
2. Client issues listen command before the process knows the server disconnected (i.e. its state is not updated)
3. Client can't send eventually reply

sorentwo: You’re right about the gap in functionality there being exposed by the failing test. The disconnect handler needs to reply with `:eventually` in this case, but should only do so if `auto_reconnect` is enabled.

Will fix 👍

lib/postgrex/notifications.ex
```diff
@@ -69,13 +69,20 @@ defmodule Postgrex.Notifications do
 
   require Logger
 
-  defstruct [:from, :ref, connected: false, listeners: %{}, listener_channels: %{}]
+  defstruct [
+    :from,
+    :ref,
+    auto_reconnect: false,
```
sorentwo, line 75: Tracking the `auto_reconnect` status now so we only reply with `:eventually` on query error when auto_reconnect is true.

lib/postgrex/notifications.ex
```diff
@@ -232,11 +249,11 @@ defmodule Postgrex.Notifications do
       %{^ref => {channel, _pid}} ->
         Process.demonitor(ref, [:flush])
 
-        state = update_in(state.listeners, &Map.delete(&1, ref))
-        state = update_in(state.listener_channels[channel], &Map.delete(&1, ref))
+        {_, state} = pop_in(state.listeners[ref])
+        {_, state} = pop_in(state.listener_channels[channel][ref])
```
sorentwo, line 253: 🤯 (makes you wonder why there isn't a `delete_in`...)

josevalim:

PR #587, opened 2021-12-22
Bring the ability of running custom queries to Postgrex.Replication

josevalim:
/cc @greg-rychlewski 

I am thinking of doing similar to copy/replication, and rather the command that starts it come from the process.

### DIFF
--- a/lib/postgrex/replication.ex
+++ b/lib/postgrex/replication.ex
@@ -71,7 +71,19 @@ defmodule Postgrex.Replication do
 
         @impl true
         def init(:ok) do
+          send(self(), :create_slot)
           {:ok, %{}}
+        end
+
+        @impl true
+        def handle_info(:create_slot, state) do
+          query = "CREATE_REPLICATION_SLOT postgrex TEMPORARY LOGICAL pg_output NOEXPORT_SNAPSHOT"
+          {:query, query, state}
+        end
+
+        @impl true
+        def handle_result(_result, state) do
+          {:noreply, state}
         end
 
         @impl true
@@ -101,10 +113,13 @@ defmodule Postgrex.Replication do
           database: "demo_dev",
           username: "postgres",
         )
-      Postgrex.Replication.create_slot(pid, "postgrex", :pgoutput)
-      Postgrex.Replication.start_replication(pid, "postgrex",
+
+      Postgrex.Replication.start_replication(
+        pid,
+        "postgrex",
         plugin_opts: [proto_version: 1, publication_names: "postgrex_example"]
       )
+
       Process.sleep(:infinity)
 
   ## `use` options
@@ -156,16 +171,6 @@ defmodule Postgrex.Replication do
   @max_lsn_component_size 8
   @max_uint64 18_446_744_073_709_551_615
   @max_copy_messages 500
-  @public_commands [
-    :create_slot,
-    :drop_slot,
-    :start_replication,
-    :show,
-    :identify_system,
-    :timeline_history,
-    :copy_table,
-    :publication_tables
-  ]
 
   @doc """
   Callback for process initialization.
@@ -183,7 +188,8 @@ defmodule Postgrex.Replication do
   Some messages require explicit acknowledgement, which can be done
   by returning a list of binaries according to the replication protocol.
   """
-  @callback handle_data(binary, state) :: {:noreply, [copy], state}
+  @callback handle_data(binary, state) ::
+              {:noreply, [copy], state} | {:query, String.t(), state}
 
   @doc """
   Receives copy messages.
@@ -193,21 +199,33 @@ defmodule Postgrex.Replication do
   to signal copying has completed.
   """
   @callback handle_copy(Postgrex.Result.t() | {:copy_done, String.t()}, state) ::
-              {:noreply, [copy], state}
+              {:noreply, [copy], state} | {:query, String.t(), state}
 
   @doc """
   Callback for `Kernel.send/2`.
   """
-  @callback handle_info(term, state) :: {:noreply, [copy], state}
+  @callback handle_info(term, state) ::
+              {:noreply, [copy], state} | {:query, String.t(), state}
 
   @doc """
   Callback for `call/3`.
 
   Replies must be sent with `reply/2`.
   """
-  @callback handle_call(term, GenServer.from(), state) :: {:noreply, [copy], state}
+  @callback handle_call(term, GenServer.from(), state) ::
+              {:noreply, [copy], state} | {:query, String.t(), state}
 
-  @optional_callbacks handle_info: 2, handle_call: 3, handle_copy: 2
+  @doc """
+  Handles result from a query command.
+
+  If any callback returns `{:query, String.t(), state}`,
+  then this callback will be immediatelly called with
+  the result of the query.
+  """
+  @callback handle_result(%Postgrex.Result{}, state) ::
+              {:noreply, [copy], state} | {:query, String.t(), state}
+
+  @optional_callbacks handle_info: 2, handle_call: 3, handle_copy: 2, handle_result: 2
 
   @doc """
   Replies to the given client.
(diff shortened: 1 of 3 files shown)


lib/postgrex/replication.ex
```diff
@@ -836,6 +690,23 @@ defmodule Postgrex.Replication do
...
+                handle(mod, :handle_result, [result, mod_state], %{s | protocol: protocol}, from)
+
+              {error, reason, protocol} ->
+                from && reply(from, {error, reason})
+                reconnect_or_stop(error, reason, protocol, %{s | state: {mod, mod_state}})
+            end
+
+          _ ->
+            reply(from, {:error, :stream_in_progress})
+            raise "cannot perform queries after replication or copying has started"
```
greg-rychlewski, line 708: I think just the reply should be good, otherwise it will kill the replication/copying?

lib/postgrex/replication.ex
```diff
@@ -101,10 +113,13 @@ defmodule Postgrex.Replication do
           database: "demo_dev",
           username: "postgres",
         )
-      Postgrex.Replication.create_slot(pid, "postgrex", :pgoutput)
-      Postgrex.Replication.start_replication(pid, "postgrex",
+
+      Postgrex.Replication.start_replication(
+        pid,
+        "postgrex",
```
greg-rychlewski, line 119:
```suggestion
        slot_name: "postgrex",
```
I forgot to update this in the last PR.

greg-rychlewski:
I had a thought about the general approach, in case you find it useful:

It feels like there is a bit of a tension trying to implement the query functionality purely through callbacks, since the user doesn't have access to the Protocol module. It's kind of like they are using their own handmade function to generate the query and then call a provided execute function, but in a bit of a roundabout/indirect way.

It might be easier to grok if there was a function like `Postgrex.Replication.execute` the user could pass their queries into. And then `handle_result` could be used to catch the `{:ok, _}`/`{:error, _}`. That's just my initial thought though. I don't think getting used to this way is particularly difficult either.

BTW when you start on switching the copying functionality, I read more of the documentation and it sounds like you can send multiple statements in a single message and they will be automatically executed in the same transaction. So could be useful instead of doing the multiple single statements the way it is now:

From 53.2.2.1. Multiple Statements In A Simple Query of https://www.postgresql.org/docs/current/protocol-flow.html
```
When a simple Query message contains more than one SQL statement (separated by semicolons), those statements are executed as a single transaction, unless explicit transaction control commands are included to force a different behavior.
```

josevalim:
> It feels like there is a bit of a tension trying to implement the query functionality purely through callbacks, since the user doesn't have access to the Protocol module. It's kind of like they are using their own handmade function to generate the query string and then call a provided execute function, but in an indirect way.

I think it is rather the opposite, most people won't be using `Postgrex.Replication.execute` because the queries would most likely common from within the process itself. So the primary goal is to support the use case of a "self-driving process", if that makes any sense. :)

lib/postgrex/replication.ex
```diff
@@ -193,21 +199,33 @@ defmodule Postgrex.Replication do
...
+              {:noreply, [copy], state} | {:query, iodata, state}
 
   @doc """
   Callback for `call/3`.
 
   Replies must be sent with `reply/2`.
   """
-  @callback handle_call(term, GenServer.from(), state) :: {:noreply, [copy], state}
+  @callback handle_call(term, GenServer.from(), state) ::
+              {:noreply, [copy], state} | {:query, String.t(), state}
```
sorentwo, line 216:
```suggestion
              {:noreply, [copy], state} | {:query, iodata, state}
```

lib/postgrex/replication.ex
```diff
@@ -193,21 +199,33 @@ defmodule Postgrex.Replication do
...
+
+  @doc """
+  Handles result from a query command.
+
+  If any callback returns `{:query, String.t(), state}`,
+  then this callback will be immediatelly called with
+  the result of the query.
+  """
+  @callback handle_result(%Postgrex.Result{}, state) ::
+              {:noreply, [copy], state} | {:query, String.t(), state}
```
sorentwo, line 226:
```suggestion
              {:noreply, [copy], state} | {:query, iodata, state}
```

lib/postgrex/replication.ex
josevalim: I think it is fine to raise because it is an error in the application logic most likely? Especially because queries come from within.

lib/postgrex/replication.ex
```diff
@@ -836,6 +690,23 @@ defmodule Postgrex.Replication do
...
+      {:query, query, mod_state} ->
+        case s do
+          %{repl_opts: nil, copy_opts: nil} ->
+            case Protocol.handle_simple(query, [], s.protocol) do
+              {:ok, %Postgrex.Result{} = result, protocol} ->
+                handle(mod, :handle_result, [result, mod_state], %{s | protocol: protocol}, from)
+
+              {error, reason, protocol} ->
+                from && reply(from, {error, reason})
+                reconnect_or_stop(error, reason, protocol, %{s | state: {mod, mod_state}})
```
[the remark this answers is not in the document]
josevalim: I like this option. @sorentwo, you will probably need to consider something similar. Maybe `handle_result` receives `{:ok, _} | {:error, Postgrex}`. Although it may also return `:disconnect` and then we must disconnect.

josevalim:
@greg-rychlewski @sorentwo I have tried to port the `copy` bits to the query API and one of the annoying things is that doing multiple queries require multiple callbacks. For example, I would need to do this:

```elixir
def handle_call({:copy_tables, table}, _from, state) do
  {:query, "BEGIN", %{state | step: {:copy_table, 1}}}
end

def handle_result({:ok, _}, _from, %{step: %{step: {:copy_table, 1}} = state) do
  {:query, "CREATE_TEMPORARY_SLOT", %{state | step: {:copy_table, 2}}}
end

def handle_result({:ok, _}, _from, %{step: %{step: {:copy_table, 2}} = state) do
  {:query, "SHOW ...", %{state | step: {:copy_table, 3}}}
end

def handle_result({:ok, %Postgrex.Result{} = ...}, _from, %{step: %{step: {:copy_table, 3}} = state) do
  {:stream, "COPY ...", %{state | step: nil}}
end
```

Maybe we have it be `{:queries, [...], ...}` where you pass a list of queries and you return `{:ok, [list_of_results]} | {:error, [successes], failure}`? Or is the above "fine"?

greg-rychlewski:
> Maybe we have it be {:queries, [...], ...} where you pass a list of queries and you return {:ok, [list_of_results]} | {:error, [successes], failure}? Or is the above "fine"?

I like the idea of accepting a list of query strings/returning the batched results. I guess one difficulty is knowing when to return, but maybe you can have something in the state for the # of remaining statements in the current transaction.

lib/postgrex/replication.ex
greg-rychlewski: Makes sense especially with the queries coming from the process.

greg-rychlewski:
The copy tables command is pretty opinionated, trying to mimic the functionality from Postgres `CREATE SUBSCRIPTION`. You could potentially forget about that and leave it up to the user/other libraries to issue the correct commands.

Then maybe you just need a way to give users the ability to checkin/set restart behaviour when they are streaming.

sorentwo:
I'd prefer to run a single query at a time rather than adding logic around partial results or multiple returns.

Can we use a `DO` block to accomplish multiple queries instead? For example, this is what the notifier will issue on reconnect:

```
"DO $$BEGIN #{listen_statements} END$$"
```

lib/postgrex/replication.ex
```diff
@@ -734,30 +606,12 @@ defmodule Postgrex.Replication do
...
   end
 
-  @doc false
-  def handle_call({name, _opts}, _from, s) when name in @public_commands,
-    do: {:reply, {:error, :stream_in_progress}, s}
-
-  @doc false
   def handle_call(msg, from, %{state: {mod, mod_state}} = s) do
-    handle(mod, :handle_call, [msg, from, mod_state], s)
+    handle(mod, :handle_call, [msg, from, mod_state], s, from)
```
sorentwo, line 614: It seems strange to have `from` as the last argument, as it's usually the `state`. What about swapping them?

```suggestion
    handle(mod, :handle_call, [msg, from, mod_state], from, s)
```

josevalim:
Hi @greg-rychlewski! I should be done with the refactoring. Now we don't know how to do any query but we should give you control over executing regular queries and streaming. The only downside is that once we enter streaming mode, there is no way to get out, which is necessary after copying tables. Can you please take a look and let me know what you think?

lib/postgrex/replication.ex
```diff
@@ -175,53 +183,83 @@ defmodule Postgrex.Replication do
...
   @doc """
-  Replies to the given client.
+  Callback for `:query` outputs.
 
-  Wrapper for `GenServer.reply/2`.
+  If any callback returns `{:query, iodata, state}`,
+  then this callback will be immediatelly called with
+  the result of the query.
+  """
+  @callback handle_result(%Postgrex.Result{} | %Postgrex.Error{}, state) ::
```
sorentwo, line 239: Is this purposefully using the literal struct for `Result` rather than a type? I noticed that the `Error` type is simply the struct anyhow.

josevalim: We should use the types, yeah.

lib/postgrex/replication.ex
```diff
@@ -694,140 +439,45 @@ defmodule Postgrex.Replication do
...
-  end
-
-  @doc false
-  def handle_call({name, _opts}, _from, s) when name in @public_commands,
-    do: {:reply, {:error, :stream_in_progress}, s}
-
-  @doc false
   def handle_call(msg, from, %{state: {mod, mod_state}} = s) do
-    handle(mod, :handle_call, [msg, from, mod_state], s)
+    handle(mod, :handle_call, [msg, from, mod_state], from, s)
```
sorentwo, line 443: `handle_call` is listed as an optional callback and this is using `handle` rather than `maybe_handle`. It makes sense to keep it optional, so perhaps `from` should be passed to `mabye_handle` too?

josevalim: It is optional because you may never call it. But if you call it, it has to be implemented. So the current code is correct!

sorentwo: Ah, of course! 👍

lib/postgrex/replication.ex
```diff
@@ -694,140 +439,45 @@ defmodule Postgrex.Replication do
...
         reconnect_or_stop(error, reason, protocol, s)
     end
   end
 
-  defp copy_recv_dispatch(copies, %{copy_opts: nil} = s), do: handle_data(copies, s)
-  defp copy_recv_dispatch(copies, s), do: handle_copy(copies, s)
-
   defp handle_data([], s), do: {:noreply, s}
 
   defp handle_data([copy | copies], %{state: {mod, mod_state}} = s) do
```
greg-rychlewski, line 463: In the `Postgrex.Protocol` module, the `handle_copy_recv` function is signalling the end of the regular copying:

```Elixir
{:ok, msg_ready(status: postgres), buffer} ->
  s = %{s | postgres: postgres, buffer: buffer}
  {:ok, Enum.reverse([:copy_done | copies]), s}
```

This could be used to reset the state/exit streaming after copying is done. Just need to document + add it to the type spec for `handle_data`.

I was reading more and it seems like having this sort of signal is also crucial when starting replication on a timeline that's not the latest one. [From the Postgres docs](https://www.postgresql.org/docs/current/protocol-replication.html):


> After streaming all the WAL on a timeline that is not the latest one, the server will end streaming by exiting the COPY mode. 
>
> When the client acknowledges this by also exiting COPY mode, the server sends a result set with one row and two columns, indicating the next timeline in this server's history. 
>
> The first column is the next timeline's ID (type int8), and the second column is the WAL location where the switch happened (type text). 
>
> Finally, the server sends two CommandComplete messages (one that ends the CopyData and the other ends the START_REPLICATION itself), and is ready to accept a new command.

greg-rychlewski:
@josevalim This is really elegant. I love it. I just had 2 comments for you to consider.

lib/postgrex/replication.ex
```diff
@@ -175,53 +183,83 @@ defmodule Postgrex.Replication do
...
-  Replies to the given client.
+  Callback for `:query` outputs.
 
-  Wrapper for `GenServer.reply/2`.
+  If any callback returns `{:query, iodata, state}`,
+  then this callback will be immediatelly called with
+  the result of the query.
+  """
+  @callback handle_result(%Postgrex.Result{} | %Postgrex.Error{}, state) ::
+              {:noreply, ack, state}
```
greg-rychlewski, line 240:
```suggestion
              {:noreply, state}
```

Would you want to change it to this, and then add another clause in the handler:

```elixir
{:noreply, mod_state} ->
  {:noreply, %{s | state: {mod, mod_state}}}
```

I don't believe any regular queries require ack. I'm pretty sure it's only when you need to let the server know your replication status.

lib/postgrex/replication.ex
```diff
@@ -836,210 +486,54 @@ defmodule Postgrex.Replication do
...
+        s = %{s | state: {mod, mod_state}}
+        max_messages = opts[:max_messages] || @max_messages
 
-  defp max_copy_messages(s), do: Keyword.get(s.repl_opts, :max_messages, @max_copy_messages)
+        with {:ok, protocol} <- Protocol.handle_streaming(query, s.protocol),
+             {:ok, protocol} <- Protocol.checkin(protocol) do
+          {:noreply, %{s | protocol: protocol, streaming: max_messages}}
+        else
+          {error_or_disconnect, reason, protocol} ->
+            reconnect_or_stop(error_or_disconnect, reason, protocol, s)
```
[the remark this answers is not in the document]
josevalim: @greg-rychlewski my thinking is that those errors are fine because it almost always means an error in your logic. :)

lib/postgrex/replication.ex
josevalim: Done!

josevalim:

PR #591, opened 2022-01-23
separate multiple results in simple queries

greg-rychlewski:
Previously a query like `SELECT 1; SELECT 2;` would pass `%PostgrexResult{rows:[["1"], ["2"]]}` to the `handle_result` callback. Now it will send `[%PostgrexResult{rows:[["1"]]}, %PostgrexResult{rows:[["2"]]}]`.

This implementation reuses the `:query` return value and the `handle_result` callback. Another way to go is to make a new return value/callback for multiple result queries. The former is a bit more straightforward imo but the latter is more in line with MyXQL's semantics. 

Actually, now that I'm writing this, I'm thinking it's probably better to keep the semantics the same between the drivers, even if the implementation is completely different. But please let me know what you think. Thanks!

One more thing I realized: it will add some extra complexity to use MyXQL's semantics because we'd have to return  errors if a multi-result query is issued using `:query`.

### DIFF
--- a/lib/postgrex/protocol.ex
+++ b/lib/postgrex/protocol.ex
@@ -1030,7 +1030,7 @@ defmodule Postgrex.Protocol do
   ## replication/notifications
 
   @spec handle_simple(String.t() | iolist(), state) ::
-          {:ok, Postgrex.Result.t() | [Postgrex.Result.t()], state}
+          {:ok, [Postgrex.Result.t()], state}
           | {:error, Postgrex.Error.t(), state}
           | {:disconnect, %DBConnection.ConnectionError{}, state}
   def handle_simple(statement, opts \\ [], %{buffer: buffer} = s) do
@@ -1068,9 +1068,8 @@ defmodule Postgrex.Protocol do
         error_ready(s, status, err, buffer)
 
       {:ok, msg_ready(status: postgres), buffer} ->
-        results = format_simple_results(results)
         s = %{s | postgres: postgres, buffer: buffer}
-        {:ok, results, s}
+        {:ok, Enum.reverse(results), s}
 
       {:ok, msg, buffer} ->
         {s, status} = handle_msg(s, status, msg)
@@ -1080,10 +1079,6 @@ defmodule Postgrex.Protocol do
         dis
     end
   end
-
-  defp format_simple_results([result]), do: result
-  defp format_simple_results([_ | _] = results), do: Enum.reverse(results)
-  defp format_simple_results(results), do: results
 
   @spec handle_copy_send([binary], state) ::
           :ok
--- a/lib/postgrex/protocol.ex
+++ b/lib/postgrex/protocol.ex
@@ -1039,7 +1039,7 @@ defmodule Postgrex.Protocol do
 
     case msg_send(%{s | buffer: nil}, msgs, buffer) do
       :ok ->
-        recv_simple(s, status, [], [], [], [], buffer)
+        recv_simple(s, status, [], [], [], buffer)
 
       {:disconnect, err, s} ->
         {:disconnect, err, s}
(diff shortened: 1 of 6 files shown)


josevalim:
If we are going to add this, then I think the result should always be a list. What do you think @sorentwo?

sorentwo:
Keeping it consistent by always making the result a list would be ideal.

josevalim:
Agreed. It will require breaking changes to the just released v0.16, but I can release a new one and retire it.

greg-rychlewski:
thanks both. I updated it to be a list result, but there is a hiccup I need to investigate. it seems like replication connections don't accept multiple commands in simple mode, even ones like `SELECT 1; SELECT 2;`. you can see the failing test. making this a draft for now.

greg-rychlewski:
@josevalim I got an answer from Tom Lane, so this is probably as authoritative as it gets :):
https://www.postgresql.org/message-id/1786010.1642990001%40sss.pgh.pa.us

The gist:
This looks like a bug and playing around with it uncovered some other parts of the parser that need work.

What it means for Postgrex:
- We can't return more than 1 result on a replication connection until this is fixed in Postgres. So returning a list is kind of weird/confusing.
- If we don't return a list right now and wait for this to be fixed in Postgres, then when it is fixed it will require users to upgrade Postgres just to be able to use this library's replication. So if they are using replication and an older version of Postgres, they can't upgrade this library to obtain non-replication improvements.

It seems like using a list right now and documenting why it's a list might be the least bad option?

test/notification_test.exs
```diff
@@ -145,7 +145,8 @@ defmodule NotificationTest do
       disconnect(context.pid_ps)
 
       # Also attempt to subscribe while it is down
-      assert {:eventually, ref2} = PN.listen(context.pid_ps, "channel2")
+      assert {ok_or_eventually, ref2} = PN.listen(context.pid_ps, "channel2")
+      assert ok_or_eventually in [:ok, :eventually]
```
greg-rychlewski, line 149: This was a heisen test that failed after one of my pushes.

test/replication_connection_test.exs
```diff
@@ -107,11 +113,15 @@ defmodule ReplicationTest do
 
   describe "handle_result" do
     test "on result", context do
-      assert {:ok, %Postgrex.Result{}} = PR.call(context.repl, {:query, "SELECT 1"})
+      assert {:ok, [%Postgrex.Result{}]} = PR.call(context.repl, {:query, "SELECT 1"})
+    end
+
+    test "on multiple results", context do
+      assert {:error, %Postgrex.Error{}} = PR.call(context.repl, {:query, "SELECT 1; SELECT 2;"})
     end
```
v0idpwn, line 121: As I understand, this tests the postgres bug. Should we?

lib/postgrex/protocol.ex
```diff
@@ -1049,30 +1049,31 @@ defmodule Postgrex.Protocol do
...
 
       {:ok, msg_data_row(values: values), buffer} ->
         row = Types.decode_simple(values, s.types)
-        recv_simple(s, status, columns, [row | rows], tags, buffer)
+        recv_simple(s, status, results, columns, [row | rows], tags, buffer)
 
       {:ok, msg_command_complete(tag: tag), buffer} ->
-        recv_simple(s, status, columns, rows, [tag | tags], buffer)
+        result = done(s, status, columns, Enum.reverse(rows), Enum.reverse([tag | tags]))
+        recv_simple(s, status, [result | results], [], [], [], buffer)
```
v0idpwn, line 1064: It seems to me that this function now never receives something other than `[]` as `tags`, so maybe the `tags` param could be removed altogether?
 
```suggestion
        result = done(s, status, columns, Enum.reverse(rows), [tag])
        recv_simple(s, status, [result | results], [], [], buffer)
```

test/replication_connection_test.exs
josevalim: Yeah, it is better not to test it as it will start breaking when we change CI?

greg-rychlewski: good call, removed

lib/postgrex/protocol.ex
greg-rychlewski: ty nice catch, updated

lib/postgrex/simple_connection.ex
```diff
@@ -175,8 +182,12 @@ defmodule Postgrex.SimpleConnection do
 
   @doc """
   Callback for processing or relaying queries executed via `{:query, query, state}`.
+  Either a list of sucessful query results or an error will be passed to this callback.
```
josevalim, line 185:
```suggestion

  Either a list of successful query results or an error will be passed to this callback.
```

josevalim:

PR #593, opened 2022-01-26
docs:  fix typo in typedoc

adampash:
When the [hexdocs are generated](https://hexdocs.pm/postgrex/Postgrex.Types.html#t:oid/0), the link to postgres oid documentation is bad (points [here](http://www.postgresql.org/docs/9.4/static/datatype-oid.html.) instead of [here](http://www.postgresql.org/docs/9.4/static/datatype-oid.html)). This just updates to markdown so the link doesn't include the period at the end of the sentence.

### DIFF
--- a/lib/postgrex/types.ex
+++ b/lib/postgrex/types.ex
@@ -8,7 +8,7 @@ defmodule Postgrex.Types do
 
   @typedoc """
   PostgreSQL internal identifier that maps to a type. See
-  http://www.postgresql.org/docs/9.4/static/datatype-oid.html.
+  [http://www.postgresql.org/docs/9.4/static/datatype-oid.html](http://www.postgresql.org/docs/9.4/static/datatype-oid.html).
   """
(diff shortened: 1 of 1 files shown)


lib/postgrex/types.ex
```diff
@@ -8,7 +8,7 @@ defmodule Postgrex.Types do
 
   @typedoc """
   PostgreSQL internal identifier that maps to a type. See
-  http://www.postgresql.org/docs/9.4/static/datatype-oid.html.
+  [http://www.postgresql.org/docs/9.4/static/datatype-oid.html](http://www.postgresql.org/docs/9.4/static/datatype-oid.html).
```
wojtekmach, line 11:
```suggestion
  <https://www.postgresql.org/docs/9.4/static/datatype-oid.html>.
```

wojtekmach:
Nice catch, thanks!

PR #595, opened 2022-02-11
Fix! Replication streaming can be resumed after reconnect

drowzy:
When `:auto_reconnect` is set to true it isn't possible to resume the replication after a reconnect due to the process being in a streaming state. 
Any `{:query, ...}` or `{:stream, ...}` from in the `handle_connect` callback end up in the [`stream_in_progress`](https://github.com/elixir-ecto/postgrex/blob/master/lib/postgrex/replication_connection.ex#L552) and the warning `09:21:28.465 [warning] received query while stream is already in progress` is printed.

The actual replication is not resumed / restarted just the connection.
This PR just sets `:streaming` to `nil` on disconnect so that one can resume / recreate the replication slot in `handle_connect`.

Also a documentation fix for the correct plugin name 😄

### DIFF
--- a/lib/postgrex/replication_connection.ex
+++ b/lib/postgrex/replication_connection.ex
@@ -75,7 +75,7 @@ defmodule Postgrex.ReplicationConnection do
 
         @impl true
         def handle_connect(state) do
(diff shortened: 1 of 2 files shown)


josevalim:

PR #597, opened 2022-02-14
Add `:ping_timeout`

### LINKED ISSUE #596: Add `:ping_timeout`

wojtekmach:
Closes #596

### DIFF
--- a/lib/postgrex/protocol.ex
+++ b/lib/postgrex/protocol.ex
@@ -20,6 +20,7 @@ defmodule Postgrex.Protocol do
             types: nil,
             null: nil,
             timeout: nil,
+            ping_timeout: nil,
             parameters: %{},
             queries: nil,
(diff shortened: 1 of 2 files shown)


lib/postgrex/protocol.ex
```diff
@@ -3098,7 +3116,14 @@ defmodule Postgrex.Protocol do
         msg_recv(s, timeout, IO.iodata_to_binary([buffer | data]))
 
       {:error, reason} ->
-        disconnect(s, tag(mod), "recv", reason, IO.iodata_to_binary(buffer))
+        action =
+          if s.postgres == :idle do
+            "recv (idle)"
```
wojtekmach, line 3121: this is so that we have a better idea where the error came from:

image

josevalim (approved):

PR #601, opened 2022-04-27
Implement the Table.Reader protocol for query result

jonatanklosko:
:cat2:

### DIFF
--- a/lib/postgrex/result.ex
+++ b/lib/postgrex/result.ex
@@ -27,7 +27,7 @@ end
 
 if Code.ensure_loaded?(Table.Reader) do
   defimpl Table.Reader, for: Postgrex.Result do
-    def init(%{columns: columns} = result) when columns in [nil, []] do
+    def init(%{columns: columns}) when columns in [nil, []] do
       {:rows, %{columns: []}, []}
     end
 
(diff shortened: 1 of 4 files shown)


lib/postgrex/result.ex
```diff
@@ -24,3 +24,15 @@ defmodule Postgrex.Result do
 
   defstruct [:command, :columns, :rows, :num_rows, :connection_id, messages: nil]
 end
+
+if Code.ensure_loaded?(Table.Reader) do
+  defimpl Table.Reader, for: Postgrex.Result do
+    def init(result) when result.columns in [nil, []] do
```
josevalim, line 30:
```suggestion
    def init(%{columns: columns} = result) when columns in [nil, []] do
```

lib/postgrex/result.ex
```diff
@@ -24,3 +24,15 @@ defmodule Postgrex.Result do
 
   defstruct [:command, :columns, :rows, :num_rows, :connection_id, messages: nil]
 end
+
+if Code.ensure_loaded?(Table.Reader) do
+  defimpl Table.Reader, for: Postgrex.Result do
+    def init(%{columns: columns} = result) when columns in [nil, []] do
```
jonatanklosko, line 30:
```suggestion
    def init(%{columns: columns}) when columns in [nil, []] do
```

josevalim:

PR #604, opened 2022-05-31
Support Unix sockets in `:hostname` and `PGHOST`

### LINKED ISSUE #603: Treat PGHOST that start with `/` (slash) as socket directory
It is how it is defined in the [PostgreSQL environment variable list](https://www.postgresql.org/docs/current/libpq-envars.html).

> Name of host to connect to. If a host name looks like an absolute path name, it specifies Unix-domain communication rather than TCP/IP communication; the value is the name of the directory in which the socket file is stored. (On Unix, an absolute path name begins with a slash. On Windows, paths starting with drive letters are also recognized.) If the host name starts with @, it is taken as a Unix-domain socket in the abstract namespace (currently supported on Linux and Windows).

The reason for that is that I like to have DB per project and I setup it via Direnv. Here I set the `PGHOST` variable, but in addition to that, only for Postgrex, I need to do additional dance to support the `socket_dir` option in Posrgrex configuration which is a little bit of a pain.

hauleth:
Close #603

### DIFF
--- a/lib/postgrex/utils.ex
+++ b/lib/postgrex/utils.ex
@@ -83,16 +83,23 @@ defmodule Postgrex.Utils do
   """
   @spec default_opts(Keyword.t()) :: Keyword.t()
   def default_opts(opts) do
+    {field, value} = extract_host(System.get_env("PGHOST"))
+
     opts
     |> Keyword.put_new(:username, System.get_env("PGUSER") || System.get_env("USER"))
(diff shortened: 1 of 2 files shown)


josevalim:
Beautiful! :green_heart: :blue_heart: :purple_heart: :yellow_heart: :heart:

hauleth:
I am not 100% sure how Erlang works with abstract UNIX sockets, but it should work OOTB. I haven't updated the documentation though about it.

Issue #608, opened 2022-06-10
SSL Config - no longer accepts functions?


jamesvl:
I'm getting the following error when building a release via `mix release`:

> (Mix) Could not read configuration file. It has invalid configuration terms such as functions, references, and pids. Please make sure your configuration is made of numbers, atoms, strings, maps, tuples and lists. The following entries are wrong:

> <snipped> ... customize_hostname_check: [match_fun: #Function<6.108018300/2 in :public_key.pkix_verify_hostname_match_fun/1>],

My Repo config is set up like

```elixir
config :portals, MyApp.Repo,
  username: "myuser",
  database: "mydb",
  hostname: "something.rds.amazonaws.com",
  port: 5432,
  ssl: true,
  ssl_opts: [
    verify: :verify_peer,
    cacertfile: "priv/cert/aws-bundle.pem",
    # Erlang option - needs to be a charlist, not a binary
    server_name_indication: 'something.rds.amazonaws.com',
    customize_hostname_check: [match_fun: :public_key.pkix_verify_hostname_match_fun(:https)],
    depth: 3
  ]
```

I took this directly from the [Postgrex SSL docs](https://hexdocs.pm/postgrex/Postgrex.html#start_link/1-ssl-client-authentication); am I missing something with my config, or did `mix` change behaviors and passing a function is no longer permitted?

josevalim:
Passing a function there was never permitted via config but I believe Ecto supervisors support an `init` callback and you can return those specific `ssl_opts` from the init callback.

Something like this:

```
config :portals, MyApp.Repo,
  aws_ssl: true
```

and then in your Ecto.Repo:

def init(_type, config) do
  config =
    if config[:aws_ssl] do
      Keyword.put(config, :ssl_opts, ...)
    else
      config
    end

  {:ok, config}
end
```

And if this works, please do send a pull request to update the docs?

jamesvl:
Thank you - I'll investigate and submit a PR for the docs for this.

PR #609, opened 2022-06-11
Update docs for configuring :ssl_opts (#608)

jamesvl:
Config files do not allowing passing functions, so the docs give an example on how to use the `init` callback of the Ecto supervisor.

Based on recommendation given in #608


josevalim:

PR #611, opened 2022-06-25
Infinity Regression Test

greg-rychlewski:
Based on the comments by @lukaszsamson  here I thought it might be a good idea to add a regression test: https://github.com/elixir-ecto/postgrex/pull/606#issuecomment-1164969404.

It feels a bit borderline to me because it's a bit like testing that Postgres is returning an error when it should. But it's also relevant to this library because it's not blocking the attempts to use invalid values.

Even if this isn't merged it's still probably of interest to see how the tests turn out. I tested on pg 13 locally and it returns an error.

### DIFF
--- a/test/query_test.exs
+++ b/test/query_test.exs
@@ -719,7 +719,7 @@ defmodule QueryTest do
     :ok = query("CREATE TABLE IF NOT EXISTS test_infinity (num numeric)", [])
     :ok = query("INSERT INTO test_infinity VALUES ($1)", [1], [])
 
-   assert %Postgrex.Error{postgres: %{code: :invalid_binary_representation}} =
+    assert %Postgrex.Error{postgres: %{code: :invalid_binary_representation}} =
              query("UPDATE test_infinity SET num = $1 WHERE num = $2", [pos_inf, 1], [])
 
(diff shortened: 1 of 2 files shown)


josevalim:
@greg-rychlewski can you pls rebase? :)

greg-rychlewski:
sorry I accidentally pushed to this repo with something meant for here =(. fixing now

greg-rychlewski:
ok now it's good. sorry about that.

test/query_test.exs
```diff
@@ -677,6 +677,55 @@ defmodule QueryTest do
     assert [[neg_inf]] == query("SELECT $1::numeric", [neg_inf])
   end
 
+  @tag max_pg_version: "14.0"
```
greg-rychlewski, line 680: If these tests are merged, I was debating what the tag should be because we need it to be exclusive. This is less verbose but could be confusing because of the min tag behaviour. I could also make it `max_pg_version_exclusive` or something similar.

josevalim:
Yeah, I am honestly a -1 on the tests because we are mostly testing PG specific behaviour and, once PG supports a new feature, I don't think we need to test that older versions did not support said feature. :)

greg-rychlewski:
I had a feeling this wasn't the greatest idea. Thanks for looking it over :).

PR #615, opened 2022-07-12
Accept `nil` as first argument in `disconnect/2`

acco:
The first argument in `Protocol.disconnect/2` is not currently used. Sometimes we find ourselves wanting to call `disconnect/2` without an exception (so `nil` is the first argument). This change removes the Dialyzer warning that occurs when you pass in `nil` as the first argument.

### DIFF
--- a/lib/postgrex/protocol.ex
+++ b/lib/postgrex/protocol.ex
@@ -203,7 +203,7 @@ defmodule Postgrex.Protocol do
   end
 
   @impl true
-  @spec disconnect(Exception.t(), state) :: :ok
+  @spec disconnect(Exception.t() | nil, state) :: :ok
   def disconnect(_, s) do
     # cancel the request first otherwise PostgreSQL will log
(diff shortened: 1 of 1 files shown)


lib/postgrex/protocol.ex
```diff
@@ -203,7 +203,7 @@ defmodule Postgrex.Protocol do
   end
 
   @impl true
-  @spec disconnect(Exception.t(), state) :: :ok
+  @spec disconnect(Exception.t() | nil, state) :: :ok
```
v0idpwn, line 206: I guess we could virtually accept `term()`?

v0idpwn (approved):

josevalim:
Unforutnately this is private API so I don't feel comfortable with changing it for direct calls nor assuming that we will never use the first argument. My suggestion is to build a dummy exception, such as `RuntimeError.exception("disconnect")` and passing that instead. Thanks!

PR #624, opened 2022-10-15
Cache SCRAM client key

greg-rychlewski:
Closes https://github.com/elixir-ecto/postgrex/issues/623.

This will allow connections to skip calculating the salted hashed password and the client key. But can't do the same for the client signature because of the nonce.


greg-rychlewski:
Just want to double check: the server key isn't being validated right? I'm not able to see it anywhere. Asking because then I have to cache the hashed password in addition to the client key.

josevalim (approved):
Looks great! We also have some unit tests for LockedCache in EXLA that we could copy over!

greg-rychlewski:
Thanks for showing me the locked cache. That's a cool idea.

PR #625, opened 2022-10-16
Add SCRAM server signature verification

greg-rychlewski:
Adds a final step in the SCRAM authentication to verify the server signature is correct.

Disclaimer: I'm for sure not an expert in this stuff. I just studied how libpq implemented it: https://github.com/postgres/postgres/blob/master/src/interfaces/libpq/fe-auth-scram.c.


josevalim (approved):

josevalim:
@rbishop can you please try this one too? Then we can ship a new release!

rbishop:
@josevalim Just tested it out and everything is working great :+1: thanks a lot!

PR #628, opened 2022-10-19
Fixes PUBLICATION name in docs 

andrewoswald:
to match `START_REPLICATION` `publication_names` ([line 84](https://github.com/elixir-ecto/postgrex/blob/master/lib/postgrex/replication_connection.ex#L84)).

This fixes the error
`message: "publication \"postgrex_example\" does not exist"`
otherwise experienced if `auto_reconnect: false`. If auto_reconnect is `true`, it will simply reconnect after error, but won't display any error messages, which is failing silently to anyone trying the example exs script.

You can observe `handle_disconnect` being called after _any_ time the connection gets wal content, which led to me experimenting with setting `auto_reconnect` to `false` in hopes an error message would surface (and it did).

### DIFF
--- a/lib/postgrex/replication_connection.ex
+++ b/lib/postgrex/replication_connection.ex
@@ -38,16 +38,16 @@ defmodule Postgrex.ReplicationConnection do
   Then you must create a publication to be replicated.
   This can be done in any session:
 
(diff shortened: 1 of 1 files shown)


josevalim:

PR #631, opened 2022-10-24
Decode regconfig

greg-rychlewski:
So that queries like this will work: `query("select $1::text::regconfig", ["english"])`


greg-rychlewski:
ah I need to rebase master to get rid of the bug report

josevalim (approved):

PR #634, opened 2022-12-12
Add Multirange extension

greg-rychlewski:
This PR adds support for the multirange data type. This was introduced into Postgres in version 14. Someone expressed interest in it here: https://groups.google.com/g/elixir-ecto/c/yVVBLLcLCD4.

A multirange looks like an array of ranges but it is different in a few ways. Postgres will remove overlaps if they exist, the binary protocol is different and the oid and sub oids are classified differently in the pg_type/pg_range catalogs

References:

- Postgres docs: https://www.postgresql.org/docs/14/rangetypes.html
- Binary protocol: https://github.com/postgres/postgres/blob/master/src/backend/utils/adt/multirangetypes.c#L328

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -41,6 +41,7 @@ jobs:
           - version: 11
           - version: 12
           - version: 13
+          - version: 14
 
         pair:
           - elixir: 1.11.4
--- a/README.md
+++ b/README.md
@@ -46,7 +46,7 @@ iex> Postgrex.query!(pid, "INSERT INTO comments (user_id, text) VALUES (10, 'hey
     array           [1, 2, 3]
     composite type  {42, "title", "content"}
     range           %Postgrex.Range{lower: 1, upper: 5}
-    multirange      [%Postgrex.Range{lower: 1, upper: 5}, %Postgrex.Range{lower: 20, upper: 23}]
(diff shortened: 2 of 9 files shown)


wojtekmach:
This is great. My only question is whether this should be `[%Postgrex.Range{}]` or `%Postgrex.MultiRange{}`. The latter offers interesting opportunities for an Enumerable implementation, e.g.:

```elixir
multirange = [
  %Postgrex.Range{lower: 1, upper: 2},
  %Postgrex.Range{lower: 5, upper: 6}
]

Enum.to_list(multirange)
#=> [1, 2, 5, 6]

1 in multirange
#=> true
```

However I can see people expecting `Enum.to_list(multirange) #=> [%Postgrex.Range{}, %Postgrex.Range{}]` instead. We don't implement Enumerable for Postgrex.Range at the moment anyway.

So yeah I think it's fine to go with the implementation as is but curious if the above resonate with you at all. :)

greg-rychlewski:
I think going with your suggestion is better. Even if we're not doing anything with ranges now, we'll be keeping the opportunity open for both of them.

Also I just realized that Postgres allows arrays of ranges and they are not the same as a multirange. Being able to distinguish between the two seems important.

v0idpwn (approved):

greg-rychlewski:
@josevalim Just wanted to double check, since you're the one most affected by this. Should I give you a chance to review or you don't mind me merging since wojtekmach and voidpwn looked it over?

josevalim:
I didn't review because they did ;)

PR #636, opened 2022-12-19
lquery and ltree extensions

greg-rychlewski:


josevalim (approved):

Nezteb:
Super neat!

Does this mean folks won't have to define their own extensions anymore to be able to use `ltree`? i.e. https://github.com/elixir-ecto/postgrex/issues/184 and https://github.com/elixir-ecto/postgrex/issues/306

greg-rychlewski:
As long as you are using Postgres 13.0+ you can use Postgrex. Before that the binary protocol didn't exist for it, only the text protocol.

For Postgres < 13.0 you have a couple options supporting the text protocol:

- https://github.com/josemrb/ecto_ltree
   - only downside is it can't support arrays
- https://github.com/greg-rychlewski/postgrex_text_ext
   - lets you configure any data type you want to use the text protocol. so can support ltree/lquery in array form as well

PR #639, opened 2023-01-25
Explain unnamed prepared statements in docs

greg-rychlewski:
Based on [this comment](https://github.com/elixir-ecto/postgrex/issues/497#issuecomment-1404321658), it seems helpful to highlight this. 

It's a bit hidden in the Postgres docs,  you'd have to know where to find it in the protocol pages https://www.postgresql.org/docs/current/protocol-flow.html.

### DIFF
--- a/lib/postgrex.ex
+++ b/lib/postgrex.ex
@@ -390,7 +390,8 @@ defmodule Postgrex do
 
   For unnamed prepared statements, pass an empty string for `name`. This can be useful
   when trying to avoid the generic query plan Postgres creates for named prepared
-  statements.
+  statements. You may also set `prepare: :unnamed` at the connection level so that
+  every prepared statement using that connection will be unnamed.
 
   ## Options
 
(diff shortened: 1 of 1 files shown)


lib/postgrex.ex
```diff
@@ -377,16 +377,21 @@ defmodule Postgrex do
...
+  query string. To execute execute the query call `execute/4`. To close the prepared query
   call `close/3`. See `Postgrex.Query` for the query data.
 
   This function may still raise an exception if there is an issue with types
   (`ArgumentError`), connection (`DBConnection.ConnectionError`), ownership
   (`DBConnection.OwnershipError`) or other error (`RuntimeError`).
 
+  For unnamed prepared statements, pass an empty string for `name`. This can be useful
+  when trying to avoid the generic query plan Postgres creates for named prepared
+  statements.
```
josevalim, line 393: Should we mention you can also set `prepare: :unnamed` at the connection level?

josevalim (approved):

PR #643, opened 2023-02-10
Use gen_statem in Postgrex.SimpleConnection

whatyouhide:
First pass, but seems to be quite promising. 🙃

### DIFF
--- a/lib/postgrex/notifications.ex
+++ b/lib/postgrex/notifications.ex
@@ -234,7 +234,7 @@ defmodule Postgrex.Notifications do
   end
 
   @impl true
-  def handle_call({:listen, channel}, {pid, _} = from, state) do
+  def handle_call({:listen, channel}, {_, {_, pid}} = from, state) do
     ref = Process.monitor(pid)
 
     state = put_in(state.listeners[ref], {channel, pid})
--- a/lib/postgrex/notifications.ex
+++ b/lib/postgrex/notifications.ex
@@ -237,7 +237,7 @@ defmodule Postgrex.Notifications do
   end
 
   @impl true
-  def handle_call({:listen, channel}, {_, {_, pid}} = from, state) do
+  def handle_call({:listen, channel}, {pid, _} = from, state) do
     ref = Process.monitor(pid)
 
     state = put_in(state.listeners[ref], {channel, pid})
--- a/lib/postgrex/notifications.ex
+++ b/lib/postgrex/notifications.ex
@@ -47,7 +47,7 @@ defmodule Postgrex.Notifications do
   ## A note on casing
 
   While PostgreSQL seems to behave as case-insensitive, it actually has a very
-  perculiar behaviour on casing. When you write:
+  peculiar behaviour on casing. When you write:
 
       SELECT * FROM POSTS
 
@@ -74,6 +74,9 @@ defmodule Postgrex.Notifications do
     2. If you cannot wrap the channel name in quotes when sending a notification,
        then make sure to give the lowercased channel name when listening
   """
+
+  @typedoc since: "0.17.0"
+  @type server :: :gen_statem.from()
 
   alias Postgrex.SimpleConnection
 
@@ -146,7 +149,7 @@ defmodule Postgrex.Notifications do
 
     * `:timeout` - Call timeout (default: `#{@timeout}`)
   """
-  @spec listen(:gen_statem.server_ref(), String.t(), Keyword.t()) ::
+  @spec listen(server, String.t(), Keyword.t()) ::
           {:ok, reference} | {:eventually, reference}
   def listen(pid, channel, opts \\ []) do
     SimpleConnection.call(pid, {:listen, channel}, Keyword.get(opts, :timeout, @timeout))
@@ -155,7 +158,7 @@ defmodule Postgrex.Notifications do
   @doc """
   Listens to an asynchronous notification channel `channel`. See `listen/2`.
   """
-  @spec listen!(:gen_statem.server_ref(), String.t(), Keyword.t()) :: reference
+  @spec listen!(server, String.t(), Keyword.t()) :: reference
   def listen!(pid, channel, opts \\ []) do
     {:ok, ref} = listen(pid, channel, opts)
     ref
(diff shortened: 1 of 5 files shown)


josevalim:
LGTM too.

lib/postgrex/simple_connection.ex
```diff
@@ -307,41 +346,51 @@ defmodule Postgrex.SimpleConnection do
...
-  def handle_info(:timeout, %{protocol: protocol} = state) do
+  def handle_event({:call, from}, {msg, caller_pid}, @state, %{state: {mod, mod_state}} = state) do
+    # We have to do a hack here to carry the actual caller PID over to the handle_call/3
+    # callback, because gen_statem uses a proxy process to do calls with timeout != :infinity.
+    # This results in the caller PID not being the same as the PID in the "from" tuple,
+    # so things like Postgrex.Notifications cannot use that "from"'s PID to register
+    # notification handlers. This approach is paired with reconstructing the proper
+    # "from" tuple in the reply/2 function in this module.
+    {from_pid, from_ref} = from
+    callback_from = {from_pid, {from_ref, caller_pid}}
```
josevalim, line 374: Generally speaking, the second element of `from` must be opaque and I think we are going to be in trouble if we ask users to look into it. I think we should rather tell users of this library to not rely on `from` and instead explicitly pass the caller too.

josevalim: In other words, I don't think the `from` should be passed on the caller here. Rather the `.listen` function in `notifications` should explicitly pass the `pid`.

josevalim: I sent a similar PR to Oban which may also be affected by this: https://github.com/sorentwo/oban/pull/843

whatyouhide: @josevalim yes, the second element of from _is_ opaque here outside of `SimpleConnection`. This is to be backwards compatible for folks using the `SimpleConnection` behaviour, so that they're unlikely to have to change their callbacks.

whatyouhide:
Aaaand all tests pass 🎉

lib/postgrex/notifications.ex
```diff
@@ -146,7 +146,7 @@ defmodule Postgrex.Notifications do
 
     * `:timeout` - Call timeout (default: `#{@timeout}`)
   """
-  @spec listen(GenServer.server(), String.t(), Keyword.t()) ::
+  @spec listen(:gen_statem.server_ref(), String.t(), Keyword.t()) ::
```
josevalim, line 149: `@type server :: :gen_statem.server_ref()` at the top?

whatyouhide: Sure, done ✅

lib/postgrex/simple_connection.ex
josevalim: That's the point. Your fix has `handle_call` looking at the second element of `from`, but there was no guarantee that would be of any shape whatsoever _before_ this PR.

josevalim: Perhaps you meant to write this:

```suggestion
    # We have to do a hack here to carry the actual caller PID over to the handle_call/3
    # callback, because gen_statem uses a proxy process to do calls with timeout != :infinity.
    # This results in the caller PID not being the same as the PID in the "from" tuple,
    # so things like Postgrex.Notifications cannot use that "from"'s PID to register
    # notification handlers. This approach is paired with reconstructing the proper
    # "from" tuple in the reply/2 function in this module.
    {from_pid, from_ref} = from
    callback_from = {caller_pid, from}
```

That would be a bit more okayish (and backwards compatible).

whatyouhide: Yeah jeez, that's exactly what I wanted to do. I mixed things up 🙈 Thanks for catching that!

whatyouhide: (and fixed)

test/simple_connection_test.exs
```diff
@@ -114,6 +114,7 @@ defmodule SimpleConnectionTest do
 
   describe "auto-reconnect" do
     @tag opts: [auto_reconnect: true]
+    @tag :focus
```
josevalim, line 117:
```suggestion
```

josevalim:
@whatyouhide it seems we have some heisen-failures now in the test suite. :(

josevalim:
You can try `elixir --erl "+T 9" -S mix test` to see if you have better luck reproducing it locally.

whatyouhide:
@josevalim some race conditions in the tests themselves I think, not in the gen_statem code. I sprinkled some tracing in there, and it seems like the Heisenfailingtest is gone.

test/simple_connection_test.exs
```diff
@@ -118,14 +118,34 @@ defmodule SimpleConnectionTest do
...
+      :erlang.trace(:all, true, [:call])
+      match_spec = [{:_, [], [{:return_trace}]}]
+      :erlang.trace_pattern({SC, :_, :_}, match_spec, [:local])
+
+      task =
+        Task.async(fn ->
+          # Simulate a small delay, because it used to cause flaky test failures
+          # in the past.
+          # https://github.com/elixir-ecto/postgrex/actions/runs/4151824827/jobs/7182370012
+          Process.sleep(10)
```
josevalim, line 131: Do we need to sleep if we have the trace?

whatyouhide: I think yes as a regression. I added a comment for it too 👍

whatyouhide:
I rebased on #644 and remove the direct dependency on `:connection`.

test/simple_connection_test.exs
whatyouhide: In my tests tracing `SC.call` was enough. I triggered CI with just that and if it passes I think we're ok to merge that, otherwise yeah I can trace the query getting to the connection.

josevalim:

lib/postgrex/simple_connection.ex
hkrutzer: Will this change after https://github.com/erlang/otp/pull/7081 ?

PR #644, opened 2023-02-11
Use gen_statem in Postgrex.ReplicationConnection

whatyouhide:
Similar to https://github.com/elixir-ecto/postgrex/pull/643, but for the other connection. Once the other PR is merged, we can also remove the direct dependency to connection in the meantime.


josevalim:

PR #655, opened 2023-06-23
Comment on how range retrieval works

nathanl:
Based on https://elixirforum.com/t/postgrex-range-behavior-is-confusing/56654

### DIFF
--- a/lib/postgrex/builtins.ex
+++ b/lib/postgrex/builtins.ex
@@ -52,6 +52,10 @@ defmodule Postgrex.Range do
   @moduledoc """
   Struct for PostgreSQL `range`.
 
+  Note that PostgreSQL itself does not return ranges exactly as stored:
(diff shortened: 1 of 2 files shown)


lib/postgrex/builtins.ex
```diff
@@ -52,6 +52,10 @@ defmodule Postgrex.Range do
   @moduledoc """
   Struct for PostgreSQL `range`.
 
+  Note that PostgreSQL itself does not return ranges exactly as stored:
+  `SELECT '(1,5)'::int4range` returns `[2,5)`, which is equivalent in terms of the values included in the range ([PostgreSQL docs](https://www.postgresql.org/docs/current/rangetypes.html#RANGETYPES-IO)).
```
josevalim, line 56:
```suggestion
  `SELECT '(1,5)'::int4range` returns `[2,5)`, which is equivalent in terms
  of the values included in the range ([PostgreSQL docs](https://www.postgresql.org/docs/current/rangetypes.html#RANGETYPES-IO)).
```

josevalim:

PR #660, opened 2023-09-14
feat: add ability to pass comments in query statements

btkostner:
This allows the ability to pass in Postgres comments as a prefix or suffix to the actual query statement. This is useful for some third party services which link traces (OpenTelemetry, Jaeger, etc) to actual database queries. For instance, here is [the code](https://github.com/DataDog/dd-trace-js/blob/master/packages/dd-trace/src/plugins/database.js#L41) that DataDog uses for tracing between services and databases.

This adds a `prefix_comment` and `suffix_comment` option to execute options which gets added to the query statement in the `String.Chars` implementation. I _did not_ do any sort of escaping or manipulation to the comments aside from wrapping them in comment blocks.

Conversation in the elixir-ecto group: https://groups.google.com/g/elixir-ecto/c/QjOJp12WgK0/m/gK-m5Dc0AAAJ

### DIFF
--- a/lib/postgrex.ex
+++ b/lib/postgrex.ex
@@ -63,6 +63,8 @@ defmodule Postgrex do
 
   @type execute_option ::
           {:decode_mapper, (list -> term)}
+          | {:prefix_comment, String.t()}
+          | {:suffix_comment, String.t()}
           | option
 
   @max_rows 500
@@ -317,6 +319,8 @@ defmodule Postgrex do
     * `:mode` - set to `:savepoint` to use a savepoint to rollback to before the
     query on error, otherwise set to `:transaction` (default: `:transaction`);
     * `:cache_statement` - Caches the query with the given name
+    * `:prefix_comment` - Comment string to put before the statement
+    * `:suffix_comment` - Comment string to put after the statement
 
   ## Examples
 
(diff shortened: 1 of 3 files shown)


btkostner:
Hmmm, after looking at the actual sqlcommenter spec, this would be better as a key value list instead of an arbitrary string. Give me a bit to update.

josevalim:
Thanks for the PR. Given Postgrex works with generated queries, why do we need these options? Can't you prepend/append the comment directly to the query?

btkostner:
So this might be better in `ecto_sql`, but the end result is being able to use the `Ecto.Repo.prepare_query/3` callback to add comments to all queries being ran. Right now this has to return `{Ecto.Query.t(), Keyword.t()}` and `Ecto.Query.t()` does not have a concept of comments. I figured passing down the comments as options all the way down to postgrex was the best bet, but I'm not super familiar with the Ecto code base. These are the options I've come up with:

1) Add a comment field to `Ecto.Query.t()`. `ecto_sql` can then support adding the comment to the statement string

2) Have comments as an option and have `ecto_sql` add the comment to the statement string

3) Have comments as an option and have `postgrex` add the comment to the statement string

End result being something like this (pseudo code):

```elixir
def prepare_query(_operation, query, _opts) do
  trace_id = get_otel_trace_id()
  
  "/*trace_id='#{trace_id}'*/ #{to_string(query)}"  
end
```

josevalim:
The problem with that, which came up in previous discussions, is that you have to pretty much disable the query statement cache, as each query is now different. I would recommend looking at other discussions in `ecto_sql` if you can find them and we can continue the discussion there, but I don't think this is a postgrex concern. :) Thanks!

btkostner:
Sorry for the late reply but I was unable to find any discussion in [elixir-ecto/ecto_sql](https://github.com/elixir-ecto/ecto_sql/issues?q=statement+comment) or [elixir-ecto/ecto](https://github.com/elixir-ecto/ecto/issues?q=statement+comment) about statement comments. Everything I found was related to table structure comments in migrations. Sorry if I missed something obvious.

As for the cache, the example above would need statement caching disabled. There is also a slimmer version of this that takes the service name, env, and version which would remain consistent and _not_ require the statement caching to be disabled. All in all, I think there are valid use cases for comments like this that wouldn't need the query statement cache disabled, but that should be at the discretion of the developer.

v0idpwn:
Check the mailing lists, @btkostner

PR #663, opened 2023-10-27
Add configuration to skip querying of comp_oids

jeregrine:
It is becoming more [common](https://github.com/elixir-ecto/postgrex/issues/645#issuecomment-1462931379) to use the "[postgres wire protocol](https://github.com/elixir-ecto/postgrex/issues?q=is%3Aissue+wire+protocol+is%3Aclosed)" for non-postgres projects. Many of these projects "support" the basic type bootstrap query, but do not support the [Array+subquery syntax nor custom types](https://github.com/elixir-ecto/postgrex/blob/master/lib/postgrex/types.ex#L90). 

I added a configuration option to pass to the repo/start_link that enables this:

`skip_comp_oids: true|false` 

Appears to work in my test cases but its pretty difficult to write an actual test for this without mocking the connection? Any guidance would be helpful.

### DIFF
--- a/lib/postgrex.ex
+++ b/lib/postgrex.ex
@@ -167,6 +167,10 @@ defmodule Postgrex do
       See the [PostgreSQL docs](https://www.postgresql.org/docs/current/ddl-schemas.html#DDL-SCHEMAS-PATH)
       for more details.
 
+    * `:disable_composite_types` - Set to `true` to disable composite types support.
+      This is useful when using Postgrex against systems that do not support composite types
+      (default: `false`).
+
 
   `Postgrex` uses the `DBConnection` library and supports all `DBConnection`
   options like `:idle`, `:after_connect` etc. See `DBConnection.start_link/2`
--- a/lib/postgrex.ex
+++ b/lib/postgrex.ex
@@ -171,7 +171,6 @@ defmodule Postgrex do
       This is useful when using Postgrex against systems that do not support composite types
       (default: `false`).
 
-
   `Postgrex` uses the `DBConnection` library and supports all `DBConnection`
   options like `:idle`, `:after_connect` etc. See `DBConnection.start_link/2`
   for more information.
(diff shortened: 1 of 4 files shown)


josevalim:
If we skip comp_oids, it means we will be unable to encode/decode records. If we want to go down this route, I think this option should be called "disable_composite_types". :)

jeregrine:
> If we skip comp_oids, it means we will be unable to encode/decode records. If we want to go down this route, I think this option should be called "disable_composite_types". :)

sounds good, I'll change it on monday. I don't think people using postgrex for its wire protocol care too much about records or custom types at all.

greg-rychlewski:
Is it just turning subquery results into an array or do these databases not support arrays at all? I was thinking it might be worth benchmarking a join with `array_agg`, if that is generally supported?

jeregrine:
Its not 100% clear to me, it seems that cockroachdb struggled with the comp oids subquery as well as some sqlite implementations that I saw. In these cases the likelihood they support records is low. As far as I can tell go/rust libs did not have this issue cause they compile default oids into the lib unless a type is defined. On Oct 28, 2023, at 2:31 AM, Greg Rychlewski ***@***.***> wrote:﻿
Is it just turning subquery results into an array or do these databases not support arrays at all? I was thinking it might be worth benchmarking a join with array_agg, if that is generally supported?

jeregrine:
@josevalim changes made.

lib/postgrex.ex
```diff
@@ -167,6 +167,10 @@ defmodule Postgrex do
       See the [PostgreSQL docs](https://www.postgresql.org/docs/current/ddl-schemas.html#DDL-SCHEMAS-PATH)
       for more details.
 
+    * `:disable_composite_types` - Set to `true` to disable composite types support.
+      This is useful when using Postgrex against systems that do not support composite types
+      (default: `false`).
+
```
josevalim, line 173:
```suggestion
```

josevalim:

PR #666, opened 2023-12-04
Handle duplicate column names in Table.Reader implementation

jonatanklosko:
`Table.Reader` expects unique column names, since it uses the keys as row/columns map keys. We decided to suffix duplicate names with `_n`.


josevalim:

PR #669, opened 2024-03-28
feat: `commit_comment` option for prepending a SQL comment to a `COMMIT` statement

bceskavich:
We're using this option at @knocklabs to inject comments for all transactions executed via the `Ecto.Repo.transaction/2` callback, which is gives us more granular insight into transaction performance in our RDS performance insights dashboard. We'd like to upstream it!

I wasn't sure where to add unit test coverage for this, if anywhere. Nor was I sure how I should try to sanitize the input, if at all.

More context in the ecto-elixir Google Group discussion here: https://groups.google.com/g/elixir-ecto/c/vUImb5ZX_CI

### DIFF
--- a/lib/postgrex/protocol.ex
+++ b/lib/postgrex/protocol.ex
@@ -3390,6 +3390,10 @@ defmodule Postgrex.Protocol do
     {:disconnect, %{err | connection_id: connection_id}, %{s | buffer: buffer}}
   end
 
+  defp disconnect(s, %Postgrex.QueryError{} = err, buffer) do
+    {:disconnect, err, %{s | buffer: buffer}}
+  end
+
   defp disconnect(s, %RuntimeError{} = err, buffer) do
     {:disconnect, err, %{s | buffer: buffer}}
   end
--- a/lib/postgrex/protocol.ex
+++ b/lib/postgrex/protocol.ex
@@ -12,6 +12,9 @@ defmodule Postgrex.Protocol do
   @nonposix_errors [:closed, :timeout]
   @max_rows 500
   @text_type_oid 25
+  @commit_comment_validation_error Postgrex.QueryError.exception(
+                                     "`:comment_comment` option cannot contain sequence \"*/\""
+                                   )
 
   defstruct sock: nil,
             connection_id: nil,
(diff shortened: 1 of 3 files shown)


josevalim:
Thank you. I have some questions:

1. Should we use `--` or `/* ... */`? Do they work the same?
2. Regardless of the form that we choose, we should probably validate it against SQL injection.
3. We'd also need docs. :)

josevalim:
I think we should go with `/*` then because it will allow multi-line comments. Then, to validate it doesn't inject SQL, we just need to validate it doesn't have "*/" in it.

bceskavich:
@josevalim I just pushed updates to add docs + validate against SQL injection. For testing the validation flow, should that live in `TransactionTest`?

josevalim:
SOunds good yeah

bceskavich:
Alright @josevalim just pushed up a unit test. I think this is ready for another review.

lib/postgrex/protocol.ex
```diff
@@ -12,6 +12,9 @@ defmodule Postgrex.Protocol do
   @nonposix_errors [:closed, :timeout]
   @max_rows 500
   @text_type_oid 25
+  @commit_comment_validation_error Postgrex.QueryError.exception(
+                                     "`:comment_comment` option cannot contain sequence \"*/\""
```
dkuku, line 16: typo

dkuku:
Should this be named differently? `transaction_comment` maybe.

lib/postgrex/protocol.ex
bceskavich: Ah whoops, thanks!

lib/postgrex/protocol.ex
```diff
@@ -537,16 +540,19 @@ defmodule Postgrex.Protocol do
           {:ok, Postgrex.Result.t(), state}
           | {DBConnection.status(), state}
           | {:disconnect, %RuntimeError{}, state}
-          | {:disconnect, %DBConnection.ConnectionError{} | Postgrex.Error.t(), state}
+          | {:disconnect,
+             %DBConnection.ConnectionError{} | Postgrex.Error.t() | Postgrex.QueryError.t(),
```
josevalim, line 544: Let's return Postgrex.Error. It is not worth introducing a new exception type only for this case. WDYT? :)

bceskavich: Fine with me!

josevalim: We can also return a RuntimeError if you prefer.

bceskavich: Went with Postgrex.Error in https://github.com/elixir-ecto/postgrex/pull/669/commits/af319f9ad753023e180ec7883cd16b39674711aa

josevalim:

PR #673, opened 2024-05-03
ReplicationConnection: allow callback to trigger disconnect

DaemonSnake:
Currently the only way to stop the replication connection workflow is to raise / exit from within a callback.
This allows callback to return {:disconnect, reason} and either stop cleanly (if auto_reconnect:false) or restart the replication connection (if set to true)


josevalim:
Looks great, can we have a simple test? One for auto_connect: false is enough. :)

DaemonSnake:
sure thing, will do ^^

DaemonSnake:
done

josevalim:

PR #677, opened 2024-05-17
Streamline SSL experience

josevalim:
@chrismccord, let me know if this works. For now, this means you need to explicitly set `ssl: :verify_full` in your config file but we can support "ssl=verify_full" via query string too.

### DIFF
--- a/lib/postgrex.ex
+++ b/lib/postgrex.ex
@@ -117,9 +117,9 @@ defmodule Postgrex do
     * `:idle_interval` - Ping connections after a period of inactivity in milliseconds.
       Defaults to 1000ms;
 
-    * `:ssl` - Enables SSL. Setting it to `true` enables SSL without host verification,
+    * `:ssl` - Enables SSL. Setting it to `true` enables SSL without server certificate verification,
       which emits a warning. Instead, prefer to set it to a keyword list, with either
-      `:cacerts` or `:cacertfile` pointing to the server certificate, to enable hostname
+      `:cacerts` or `:cacertfile` set to a CA trust store, to enable server certificate
       verification. Defaults to `false`;
 
     * `:socket_options` - Options to be given to the underlying socket
--- a/lib/postgrex.ex
+++ b/lib/postgrex.ex
@@ -47,8 +47,7 @@ defmodule Postgrex do
           | {:connect_timeout, timeout}
           | {:handshake_timeout, timeout}
           | {:ping_timeout, timeout}
-          | {:ssl, boolean | :verify_full}
-          | {:ssl_opts, [:ssl.tls_client_option()]}
+          | {:ssl, boolean | [:ssl.tls_client_option()]}
           | {:socket_options, [:gen_tcp.connect_option()]}
           | {:prepare, :named | :unnamed}
           | {:transactions, :strict | :naive}
@@ -118,14 +117,10 @@ defmodule Postgrex do
     * `:idle_interval` - Ping connections after a period of inactivity in milliseconds.
       Defaults to 1000ms;
 
-    * `:ssl` - Set to `:verify_full` to enable full SSL verification, including peer
-      and hostname. You must also specify either `:cacerts` or `:cacertfile` in
-      `:ssl_opts`. Alternatively, you may set it to `true` for enabling SSL, but
-      use `:ssl_opts` to opt-in and control verification. Defaults to `false`;
-
-    * `:ssl_opts` - A list of ssl options, see the
-      [`tls_client_option`](http://erlang.org/doc/man/ssl.html#type-tls_client_option)
-      from the ssl docs. Only applies if `:ssl` is also enabled;
+    * `:ssl` - Enables SSL. Setting it to `true` enables SSL without host verification,
+      which emits a warning. Instead, prefer to set it to a keyword list, with either
+      `:cacerts` or `:cacertfile` pointing to the server certificate, to enable hostname
+      verification. Defaults to `false`;
 
     * `:socket_options` - Options to be given to the underlying socket
       (applies to both TCP and UNIX sockets);
--- a/lib/postgrex.ex
+++ b/lib/postgrex.ex
@@ -191,37 +191,13 @@ defmodule Postgrex do
   ## SSL client authentication
 
   When connecting to Postgres or CockroachDB instances over SSL it is idiomatic to use
-  certificate authentication. Config files do not allowing passing functions,
-  so use the `init` callback of the Ecto supervisor.
+  certificate authentication:
 
-  In your Repository configuration:
+      [ssl: [cacertfile: System.get_env("DB_CA_CERT_FILE")]]
 
-      config :app, App.Repo,
-        ssl: String.to_existing_atom(System.get_env("DB_SSL_ENABLED", "true")),
-        verify_ssl: true
-
-  And in App.Repo, set your `:ssl_opts`:
-
-      def init(_type, config) do
-        config =
-          if config[:verify_ssl] do
-            Keyword.put(config, :ssl_opts, my_ssl_opts(config[:hostname]))
-          else
-            config
-          end
-
-        {:ok, config}
-      end
-
-      def my_ssl_opts(server) do
-        [
-          verify: :verify_peer,
-          cacertfile: System.get_env("DB_CA_CERT_FILE"),
-          server_name_indication: String.to_charlist(server),
-          customize_hostname_check: [match_fun: :public_key.pkix_verify_hostname_match_fun(:https)],
-          depth: 3
-        ]
-      end
+  The server name indication (SNI) will be automatically set based on the `:hostname`
+  configuration, if one was provided. Other options, such as `depth: 3`, may be necessary
+  depending on the server.
 
   ## PgBouncer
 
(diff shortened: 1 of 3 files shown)


lib/postgrex/protocol.ex
```diff
@@ -117,14 +125,37 @@ defmodule Postgrex.Protocol do
...
     }
 
     connect_endpoints(endpoints, sock_opts ++ @sock_opts, connect_timeout, s, status, [])
   end
 
+  defp default_ssl_opts(:verify_full, opts) do
+    case Keyword.fetch(opts, :hostname) do
+      {:ok, hostname} ->
+        [
+          server_name_indication: hostname,
```
wojtekmach, line 140:
```suggestion
          server_name_indication: String.to_charlist(hostname),
```

lib/postgrex/protocol.ex
```diff
@@ -117,14 +125,37 @@ defmodule Postgrex.Protocol do
...
+          customize_hostname_check: [
+            match_fun: :public_key.pkix_verify_hostname_match_fun(:https)
+          ]
```
wojtekmach, line 143: Can we also add `verify: :verify_peer`? I believe it's not enabled by default.

```suggestion
          ],
          verify: :verify_peer
```

wojtekmach (approved):
Just one comment about charlist and this is good to go. I tested it on DigitalOcean and it worked!

```elixir
Mix.install([
  {:postgrex, github: "elixir-ecto/postgrex", branch: "jv-add-ssl-verify-full"}
])

{:ok, pid} =
  Postgrex.start_link(
    hostname: System.fetch_env!("DO_HOST"),
    port: 25060,
    username: "doadmin",
    password: System.fetch_env!("DO_PASSWORD"),
    database: "defaultdb",
    ssl: :verify_full,
    ssl_opts: [
      cacertfile: "/Users/wojtek/Downloads/ca-certificate.crt"
    ]
  )

IO.inspect(Postgrex.query!(pid, "SELECT NOW()", []))
```

lib/postgrex/protocol.ex
wojtekmach: Sorry, please disregard this, I think it _is_ using verify_peer by default:

If I use just `ssl: :verify_full` I get this error:

```
11:13:24.657 [error] Postgrex.Protocol (#PID<0.203.0>) failed to connect: ** (DBConnection.ConnectionError) ssl connect: Options (or their values) can not be combined: [{verify,verify_peer},
                                                {cacerts,undefined}] - {:options, :incompatible, [verify: :verify_peer, cacerts: :undefined]}
```

Unfortunately I can't just pass `ssl_opts: [cacerts: :public_key.cacerts_get()]` because DO, like I'm sure most servers, are not using public certs? Passing:

```
ssl_opts: [
  cacertfile: "/Users/wojtek/Downloads/ca-certificate.crt"
]
```

that I got from them fixes the error.

lib/postgrex.ex
```diff
@@ -118,7 +118,10 @@ defmodule Postgrex do
     * `:idle_interval` - Ping connections after a period of inactivity in milliseconds.
       Defaults to 1000ms;
 
-    * `:ssl` - Set to `true` if ssl should be used (default: `false`);
+    * `:ssl` - Set to `true` if ssl should be used (default: `false`). You may also
+      set it to `:verify_full`, which enables peer and hostname validation. You must
+      specify either `:cacerts` or `:cacertfile` in `:ssl_opts` if `:verify_full`
+      is enabled;
```
josevalim, line 124:
```suggestion
    * `:ssl` - Set to `:verify_full` to enable full SSL verification, including peer
      and hostname. You must also specify either `:cacerts` or `:cacertfile` in
      `:ssl_opts`. Alternatively, you may set it to `true` for enabling SSL, but
      use `:ssl_opts` to opt-in and control verification. Defaults to `false`.
```

josevalim:
@voltone, please let me know if you have thoughts here, the goal is to mirror this in myxql as well.

chrismccord (approved):
:shipit:

lib/postgrex/protocol.ex
```diff
@@ -81,12 +81,29 @@ defmodule Postgrex.Protocol do
...
 
+    {ssl_opts, opts} =
+      case Keyword.pop(opts, :ssl, false) do
+        {false, opts} ->
+          {nil, opts}
+
+        {true, opts} ->
+          Logger.warning(
+            "setting ssl: true on your database connection offers only limited protection, " <>
+              "as the hostname is not verified. Set \"ssl: [cacertfile: path/to/file]\" instead"
```
voltone, line 97:
```suggestion
              "as the server's certificate is not verified. Set \"ssl: [cacertfile: path/to/file]\" instead"
```

lib/postgrex/protocol.ex
```diff
@@ -117,14 +134,27 @@ defmodule Postgrex.Protocol do
...
+      ssl: ssl_opts,
       target_server_type: target_server_type,
       search_path: opts[:search_path]
     }
 
     connect_endpoints(endpoints, sock_opts ++ @sock_opts, connect_timeout, s, status, [])
   end
 
+  defp default_ssl_opts(opts) do
+    case Keyword.fetch(opts, :hostname) do
```
voltone, line 146: If hostname is not set and PGHOST is used instead, the `server_name_indication` won't reflect the actual hostname, and server name verification will fail (as the TLS handshake is started on an established TCP connection and `:ssl` does not know the hostname that was used to establish the connection in the first place).

So either the docs would have to state that `hostname` option is mandatory when using `verify_peer`, or the logic to set `server_name_indication` would have to move to a later place where the actual hostname is known, no?

voltone (commented):
There are a few more places in the docs where ssl options are described, including for example the `endpoints` list under "Failover with SSL support". I don't have time right now to trace the behavior of `endpoints` and understand the implications to propose updates, sorry. Thought I'd give my feedback now before this gets merged. I might have time tonight...

lib/postgrex.ex
```diff
@@ -118,11 +117,10 @@ defmodule Postgrex do
...
 
-    * `:ssl` - Set to `true` if ssl should be used (default: `false`);
-
-    * `:ssl_opts` - A list of ssl options, see the
-      [`tls_client_option`](http://erlang.org/doc/man/ssl.html#type-tls_client_option)
-      from the ssl docs;
+    * `:ssl` - Enables SSL. Setting it to `true` enables SSL without host verification,
+      which emits a warning. Instead, prefer to set it to a keyword list, with either
+      `:cacerts` or `:cacertfile` pointing to the server certificate, to enable hostname
+      verification. Defaults to `false`;
```
voltone, line 123: Setting `verify_peer` enables server certificate verification, with hostname verification being just a part of that. Without it, the server certificate is completely ignored, not just the hostname.

Also, `:cacerts` and `:cacertfile` need to point to a CA trust store, not to the server certificate.

```suggestion
    * `:ssl` - Enables SSL. Setting it to `true` enables SSL without server certificate verification,
      which emits a warning. Instead, prefer to set it to a keyword list, with either
      `:cacerts` or `:cacertfile` set to a CA trust store, to enable server certificate
      verification. Defaults to `false`;
```

lib/postgrex/protocol.ex
josevalim: I think this is good. PGHOST is read on `Postgrex.start_link` and set as the `opts[:hostname]`. If endpoints is set, that wins over hostname, but we will still use the hostname for SNI. I will update the docs!

voltone: So if `hostname` is never supposed to be absent, why do we need the `case` statement?

josevalim: @voltone do you know if :server_name_indicator has any effect if we have `verify: :verify_none`?

josevalim:
Thank you @voltone, I believe I have addressed all of your concerns. However, there is one problem, we are always setting `server_name_indicator`, even in the cases where we would previously set only `ssl: true` (i.e. when `verify: :verify_none`). Let me know if this can be a concern because then we can set `server_name_indicator: disabled` instead.

josevalim:
> However, there is one problem, we are always setting server_name_indicator, even in the cases where we would previously set only ssl: true (i.e. when verify: :verify_none)

From some tests this does not seem to be an issue and, if it happens, folks can always set it to `:disabled`.

wojtekmach (approved):
I tested this on Supabase, CockroachDB, DigitalOcean PostgreSQL, and PlanetScale (which is MySQL, but same idea) and it works well! I'll work on similar change for MyXQL soon.

josevalim:

ruslandoga:
👋

It seems like setting `?ssl=true` overwrites the keyword options at https://github.com/elixir-ecto/ecto/blob/cba494787b8e98aa0038f6c35e2f44d3661f2620/lib/ecto/repo/supervisor.ex#L28

```elixir
config :app, Repo,
  url: "ecto://postgresql://your_username:your_password@cluster-do-user-1234567-0.db.ondigitalocean.com:25060/defaultdb?ssl=true",
  ssl: [cacertfile: "/path/to/cacert.pem"]
```
results in
```elixir
Postgrex.child_spec([repo: Repo, telemetry_prefix: [:app, :repo], otp_app: :app, ...etc..., ssl: true, pool: DBConnection.ConnectionPool])
```
and a warning
```
[warning] setting ssl: true on your database connection offers only limited protection, as the server's certificate is not verified. Set "ssl: [cacertfile: path/to/file]" instead
```

I wonder if URL parsing logic should be updated to avoid overwriting `ssl` cacerts?

josevalim:
Yes, we should probably fix it in Ecto. Maybe we ignore `ssl: true` if there is already a list in place (and we emit a warning).

lib/postgrex/protocol.ex
```diff
@@ -81,12 +81,29 @@ defmodule Postgrex.Protocol do
...
+          {nil, opts}
+
+        {true, opts} ->
+          Logger.warning(
+            "setting ssl: true on your database connection offers only limited protection, " <>
+              "as the server's certificate is not verified. Set \"ssl: [cacertfile: path/to/file]\" instead"
+          )
+
+          # Read ssl_opts for backwards compatibility
+          Keyword.pop(opts, :ssl_opts, [])
```
ruslandoga, line 101: 👋 

Sorry for posting on an old PR, but should the default be `verify_peer: :none`? it seems like, by default, `:ssl` still tries to verify the peer.

```elixir
iex(1)> :ssl.connect(~c"google.com", 443, [])
{:error, {:options, :incompatible, [verify: :verify_peer, cacerts: :undefined]}}
```

wojtekmach: I think previous OTP defaulted to :verify_none and more recent to :verify_peer and so I believe the idea behind this legacy `ssl: true` was to keep whatever that particular OTP defaulted to however `ssl: list` would force secure defaults.

wojtekmach: Do you know if this change broke any apps? Cause if the warning implies less security we should change the warning not intentionally make things less secure. Sorry I don't have an easy way to check how it all works right now.

ruslandoga: No, I don't think it broke any apps. My suggestion was purely a QoL improvement :)

PR #682, opened 2024-06-26
Default to system certificate store

greg-rychlewski:
Does this fit in with the recent changes to the `:ssl` option? If the user provides ssl options but does not supply `:cacerts` or `:cacertfile` then it uses the erlang default.

### DIFF
--- a/lib/postgrex/protocol.ex
+++ b/lib/postgrex/protocol.ex
@@ -101,7 +101,7 @@ defmodule Postgrex.Protocol do
           Keyword.pop(opts, :ssl_opts, [])
 
         {ssl_opts, opts} when is_list(ssl_opts) ->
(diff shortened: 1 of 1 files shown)


josevalim:
My understanding is that most of them have their own certificates. For the few that don’t, being explicit maybe does not hurt? Thoughts?

greg-rychlewski:
My experience in this area isn't too prolific, but in the few things I've used what you said is true.

Where this PR came from is I had it on my todo list after a couple people said it would be useful in the google group: https://groups.google.com/g/elixir-ecto/c/31K2gjLVGRs.

But I'm happy to keep things the way they are. Your call!

josevalim:
Exactly. I think it is a sane default for the web but I am not sure if it works for databases. So let's skip it for now!

PR #683, opened 2024-06-28
Allow custom extensions to use options

greg-rychlewski:
Currently the extensions users create themselves cannot access the options passed to the type module. A few of the options, like `:null` and `:decode_binary` seem like they should be passed to them. For example, we previously asked users to define their own `ltree` and `lquery` extensions and our own implementations are using the `:decode_binary` option. Also, our own documented example of a custom extension is passed the opts https://hexdocs.pm/postgrex/Postgrex.Extension.html

### DIFF
--- a/lib/postgrex/type_module.ex
+++ b/lib/postgrex/type_module.ex
@@ -920,9 +920,12 @@ defmodule Postgrex.TypeModule do
     end
   end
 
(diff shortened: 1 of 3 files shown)


greg-rychlewski:
I had a misunderstanding. We ask the user to pass a tuple if they want opts. Closing this

PR #684, opened 2024-07-01
Allow extensions to access type modifiers 

greg-rychlewski:
Closes https://github.com/elixir-ecto/postgrex/pull/676

The main idea is: 

- Postgres returns a type modifier in the [row description message](https://www.postgresql.org/docs/current/protocol-message-formats.html#PROTOCOL-MESSAGE-FORMATS-ROWDESCRIPTION) along with the type oid. This modifier is where the timestamp precision information is stored. Other types can store information there as well, like decimal precision or length of  varchar.
- The decoder needs access to the type modifier so we allow the decode function to pattern match against it. I don't think it can be in the state because that is global to all decode instances. 
- I also considered having a new callback to post-process the decoding. I don't think that simplifies the code much and also arrays would have to do a second pass to apply the post-processing.
- Super types have type modifiers as well but through my investigation the only one that has a type modifier that makes sense is array. It will be the type modifier of the elements. For other things like ranges and records, the modifier is not matching the one on the subtype. So I do not pass the modifier of the super extensions down to the sub types unless it's an array.
- The inlined type module defines a lot of functions extension/arity and care had to be taken that adding the type modifier did not cause arities to overlap. That's why I pass `nil` for things like tuples where they don't care about the modifier. So that the different arity functions are preserved.

I only applied to changes for `timestamp` but I can do it for the other time related extensions if this seems like a good idea.

### DIFF
--- a/lib/postgrex/protocol.ex
+++ b/lib/postgrex/protocol.ex
@@ -955,7 +955,7 @@ defmodule Postgrex.Protocol do
   defp set_search_path_recv(s, status, buffer) do
     case msg_recv(s, :infinity, buffer) do
       {:ok, msg_row_desc(fields: fields), buffer} ->
-        {[@text_type_oid], ["search_path"]} = columns(fields)
+        {[@text_type_oid], ["search_path"], _} = columns(fields)
         set_search_path_recv(s, status, buffer)
 
       {:ok, msg_data_row(), buffer} ->
@@ -1014,7 +1014,7 @@ defmodule Postgrex.Protocol do
        ) do
     case msg_recv(s, :infinity, buffer) do
       {:ok, msg_row_desc(fields: fields), buffer} ->
-        {[@text_type_oid], ["transaction_read_only"]} = columns(fields)
+        {[@text_type_oid], ["transaction_read_only"], _} = columns(fields)
         check_target_server_type_recv(s, status, buffer)
 
       {:ok, msg_data_row(values: values), buffer} ->
@@ -1611,8 +1611,9 @@ defmodule Postgrex.Protocol do
        )
        when ref == nil or protocol_types != query_types do
     with {:ok, s, buffer} <- recv_parse(s, status, buffer),
-         {:ok, param_oids, result_oids, columns, s, buffer} <- recv_describe(s, status, buffer) do
-      describe(s, query, param_oids, result_oids, columns, buffer)
+         {:ok, param_oids, result_oids, result_mods, columns, s, buffer} <-
+           recv_describe(s, status, buffer) do
+      describe(s, query, param_oids, result_oids, result_mods, columns, buffer)
     else
       {:error, %Postgrex.Error{} = error, s, buffer} ->
         {:error, %{error | query: query.statement}, s, buffer}
@@ -1626,13 +1627,13 @@ defmodule Postgrex.Protocol do
     %Query{param_oids: param_oids, result_oids: result_oids, columns: columns} = query
 
     with {:ok, s, buffer} <- recv_parse(s, status, buffer),
-         {:ok, ^param_oids, ^result_oids, ^columns, s, buffer} <-
+         {:ok, ^param_oids, ^result_oids, _, ^columns, s, buffer} <-
            recv_describe(s, status, param_oids, buffer) do
       query_put(s, query)
       {:ok, query, s, buffer}
     else
-      {:ok, ^param_oids, new_result_oids, new_columns, s, buffer} ->
-        redescribe(s, query, new_result_oids, new_columns, buffer)
+      {:ok, ^param_oids, new_result_oids, new_result_mods, new_columns, s, buffer} ->
+        redescribe(s, query, new_result_oids, new_result_mods, new_columns, buffer)
 
       {:error, %Postgrex.Error{}, _, _} = error ->
         error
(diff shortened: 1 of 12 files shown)


lib/postgrex/protocol.ex
```diff
@@ -3185,9 +3187,13 @@ defmodule Postgrex.Protocol do
...
   defp columns(fields) do
-    fields
-    |> Enum.map(fn row_field(type_oid: oid, name: name) -> {oid, name} end)
-    |> :lists.unzip()
+    {oids_mods, names} =
+      fields
+      |> Enum.map(fn row_field(type_oid: oid, type_mod: mod, name: name) -> {{oid, mod}, name} end)
+      |> :lists.unzip()
+
+    {oids, mods} = :lists.unzip(oids_mods)
```
josevalim, line 3195: You can keep it as a flat list and use unzip3: https://www.erlang.org/doc/apps/stdlib/lists.html#unzip3/1

lib/postgrex/protocol.ex
```diff
@@ -1759,9 +1760,10 @@ defmodule Postgrex.Protocol do
     {:ok, query}
   end
 
-  defp describe_result(%{types: types}, query, result_oids, columns) do
+  defp describe_result(%{types: types}, query, result_oids, result_mods, columns) do
     with {:ok, result_info} <- fetch_type_info(result_oids, types),
-         {result_formats, result_types} = Enum.unzip(result_info) do
+         {result_formats, result_types} = Enum.unzip(result_info),
+         result_types = Enum.zip(result_types, result_mods) do
```
josevalim, line 1766: This and the previous line can be moved inside the `do-block`.

lib/postgrex/type_module.ex
```diff
@@ -382,11 +382,12 @@ defmodule Postgrex.TypeModule do
   defp decode_row_dispatch(extension, :super_binary, rest, acc, rem, full, rows) do
     [clause] =
       quote do
-        [{unquote(extension), sub_oids, sub_types} | types] ->
+        [{{unquote(extension), sub_oids, sub_types}, mod} | types] ->
```
josevalim, line 385: Can we avoid nesting tuples if possible?

lib/postgrex/type_module.ex
```diff
@@ -920,6 +1021,22 @@ defmodule Postgrex.TypeModule do
     end
   end
 
+  defp split_super_decode({:->, _, [head, body]}) do
+    case head do
+      [{:when, _, [pattern, sub_oids, sub_types, guard]}] ->
```
josevalim, line 1026: Super extensions are private, so we can update all of them to receive the mod and simplify the logic here.

josevalim:
@greg-rychlewski there is another approach which we can take here, which is to say match on `mod` as `var!(mod)` and then tell library authors that they can access the modifier as `var!(mod)`. This means we don't need to change the pattern matching logic and that we can add other "arguments" in the future. I am don't have a preference at the moment.

Btw, does this also work if you have an array of timestamps? Can we forward the mod information down?

greg-rychlewski:
>@greg-rychlewski there is another approach which we can take here, which is to say match on mod as var!(mod) and then tell library authors that they can access the modifier as var!(mod). This means we don't need to change the pattern matching logic and that we can add other "arguments" in the future

This sounds like a nice option to me. I will try it out in its own commit so we can take a look and revert it if it's no good.


> Btw, does this also work if you have an array of timestamps? Can we forward the mod information down?

Thankfully yes. Even though we only get the modifier for the array type, Postgres makes it the same as the sub- type. I added a test for it here: https://github.com/elixir-ecto/postgrex/pull/684/files#diff-dbb00c889130c35f21ecd6d123b9dc135fc6b1478fefb50fada337914e4ef94bR1773

For other super types this doesn't seem to be the case. Ranges and Records do not have any information about the sub type modifiers.

lib/postgrex/type_module.ex
greg-rychlewski: I can do it if I flatten during parsing the row description method like this: https://github.com/elixir-ecto/postgrex/pull/684/commits/7057e83fe6fcf5a7ad4e68f4348b1f78e8bb3da9#diff-0da854f0c1cda9486d776c72ecda6a2e595a7667b72688669bbd80d6b80f0f96R1768

greg-rychlewski:
@josevalim I gave it a shot two ways:

1. Using the original method (and addressing your comments): [please see this commit](https://github.com/elixir-ecto/postgrex/pull/684/commits/7057e83fe6fcf5a7ad4e68f4348b1f78e8bb3da9)
2. Using the var! method. Please see the latest commit. I'm not too well versed in this type of stuff so not sure if this is exactly what you had in mind. But I got the tests to pass. The one thing I notice though is I get an unused attribute warning for all the extensions not using `var!(mod)`. I'm not sure if it's a limitation of the method or I'm just doing something wrong. this is an example of the warning:

>warning: variable "mod" is unused (if the variable is not meant to be used, prefix it with an underscore)

josevalim:
You can add `_ = var!(mod)` to each body to silence it. I will review with care tomorrow!

josevalim (approved):
It looks great to me! The only question is if we want to use a name more specific than `mod`, but we can decide this once we expose it to users.

PR #689, opened 2024-07-06
Remove max year from timestamptz

greg-rychlewski:
I think we can just do this? Postgres will return an error if the timestamp is out of range with the message 

>message: "timestamp out of range:..."


wojtekmach:
Sounds good to me. What do you think about adding a test? Curious if the error differs between pg versions we support.

greg-rychlewski:
Good idea. I added a test for the negative bound because the upper bound will crash on most elixir versions.

josevalim (approved):

wojtekmach:
@greg-rychlewski I was rather thinking about this:

```elixir
  test "decode timestamptz out of bounds", context do
    %Postgrex.Error{postgres: %{message: message}} =
      query("SELECT $1::timestamptz", [~U[-4713-01-01 00:00:00Z]])

    assert message =~ "timestamp out of range"
  end
```

so it goes through encoding which was the change here. WDYT? If this sounds good to you, Im happy to make the change on main.

greg-rychlewski:
sounds good, please go for it:) thank you

PR #694, opened 2024-07-23
Fix typo in docs for Types.define/3

DriesCruyskens:
I'm not completely sure this is what you meant but I think `how Postgrex encodes and decodes data coming from Postgrex` seems to be a duplicate by mistake.


josevalim:

PR #696, opened 2024-07-26
Add support for Elixir Durations

greg-rychlewski:
Addresses https://github.com/elixir-ecto/postgrex/issues/688 besides the deprecation of `Postgrex.Interval`.

The following changes were made:

- allow encoding of `Duration` structs for postgres interval extension
- allow decoding postgres interval types to `Duration` structs if the option `:interval_decode_type` is set to `Duration`
- both of the above two are only allowed if the elixir version is at least 1.17.0

Some notes:

- when breaking down the postgres response (microseconds, days and months) into the fields for the `Duration` struct, I went with the same style we use for `Postgrex.Interval`: translate the lower units into the higher units as much as possible
- the type modifier postgres returns for intervals is not an integer from 0 to 6. i will try to find out what it means but it's not well documented so probably have to go through their source code. for now i always return precision of 6 because they are giving us microseconds in the response
- from what i can see we are not documenting anywhere the options that can be passed to any of our extensions. in a follow-up PR i'll try to collect them all and put them in a good place

### DIFF
--- a/lib/postgrex/extensions/interval.ex
+++ b/lib/postgrex/extensions/interval.ex
@@ -75,7 +75,7 @@ defmodule Postgrex.Extensions.Interval do
     def encode(_) do
       quote location: :keep do
         %Postgrex.Interval{months: months, days: days, secs: seconds, microsecs: microseconds} ->
-          microseconds = 1_000_000 * secs + microseconds
+          microseconds = 1_000_000 * seconds + microseconds
           <<16::int32(), microseconds::int64(), days::int32(), months::int32()>>
 
         other ->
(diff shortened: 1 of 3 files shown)


lib/postgrex/extensions/interval.ex
```diff
@@ -3,23 +3,93 @@ defmodule Postgrex.Extensions.Interval do
...
-    quote location: :keep do
-      %Postgrex.Interval{months: months, days: days, secs: secs, microsecs: microsecs} ->
-        microsecs = secs * 1_000_000 + microsecs
-        <<16::int32(), microsecs::int64(), days::int32(), months::int32()>>
-
-      other ->
-        raise DBConnection.EncodeError, Postgrex.Utils.encode_msg(other, Postgrex.Interval)
+  def init(opts), do: Keyword.get(opts, :interval_decode_type, Postgrex.Interval)
+
+  if Version.match?(System.version(), ">= 1.17.0") do
```
josevalim, line 8:
```suggestion
  if Code.ensure_loaded?(Duration) do
```

josevalim (approved):

greg-rychlewski:
Would you want to add a :duration type to ecto?

josevalim:
We should! :)

PR #699, opened 2024-07-27
Respect precision for intervals

greg-rychlewski:
In the previous PR we hardcoded the precision to 6 because we didn't know how to deal with the type modifier.

I got pointed in the right direction by the nice people at Postgres. The type modifier for intervals takes into account two different things:

1. The allowed fields (i.e. only `days`, `years`, `days to years` and stuff like that), represented as a 16-bit bit field
2. The precision represented as a 16 bit integer

The right most 16 bits in the type modifier are the precision. I was pointed to the right places in their code [here](https://git.postgresql.org/gitweb/?p=postgresql.git;a=blob;f=src/backend/utils/adt/timestamp.c;h=69fe7860ede062fc8be42e7545b35e69c3e068c4;hb=HEAD#l1151) and [here](https://git.postgresql.org/gitweb/?p=postgresql.git;a=blob;f=src/include/utils/timestamp.h;h=a6ce03ed46045fb748daf02f0f6e06492ad5a181;hb=HEAD#l81)

We could probably do something about the fields modifier as well but that is a lot more complex. And we ignore it already today. So going to think on it a bit and save it for later.

### DIFF
--- a/lib/postgrex/extensions/interval.ex
+++ b/lib/postgrex/extensions/interval.ex
@@ -6,7 +6,7 @@ defmodule Postgrex.Extensions.Interval do
   def init(opts), do: Keyword.get(opts, :interval_decode_type, Postgrex.Interval)
 
   if Code.ensure_loaded?(Duration) do
-    import Bitwise
+    import Bitwise, warn: false
     @default_precision 6
     @precision_mask 0xFFFF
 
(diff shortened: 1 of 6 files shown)


wojtekmach:
import Bitwise, warn: false

josevalim (approved):

greg-rychlewski:
Just to be extra cautious:

Is it safe to assume across elixir/erlang versions that `-1 &&& 0xFFFF` is always `0xFFFF`

josevalim:
Yes. Otherwise it would have to be a breaking change in Erlang.

PR #702, opened 2024-08-06
Fix dialyzer issues with interval extension, part 2

greg-rychlewski:
Really not sure about the latest warning

```elixir
deps/postgrex/lib/postgrex/type_module.ex:1084:call
The function call will not succeed.

:erlang.band(_mod :: nil, 65535)

will never return since the 1st arguments differ
from the success typing arguments:

(integer(), integer())
```

@oliver-kriska  could you try this branch before we merge? ty

### DIFF
--- a/lib/postgrex/extensions/interval.ex
+++ b/lib/postgrex/extensions/interval.ex
@@ -52,14 +52,11 @@ defmodule Postgrex.Extensions.Interval do
     def decode(type) do
       quote location: :keep, generated: true do
         <<16::int32(), microseconds::int64(), days::int32(), months::int32()>> ->
-          type_mod = var!(mod)
-          precision = if type_mod, do: type_mod &&& unquote(@precision_mask)
-
           unquote(__MODULE__).decode_interval(
             microseconds,
             days,
             months,
-            precision,
+            var!(mod),
             unquote(type)
           )
       end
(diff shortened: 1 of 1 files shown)


oliver-kriska:
unfortunately still same issue

```
deps/postgrex/lib/postgrex/type_module.ex:1084:call
The function call will not succeed.

:erlang.band(_ :: nil, 65535)

will never return since the 1st arguments differ
from the success typing arguments:

(integer(), integer())
```

lib/postgrex/extensions/interval.ex
```diff
@@ -52,7 +52,8 @@ defmodule Postgrex.Extensions.Interval do
     def decode(type) do
       quote location: :keep do
```
josevalim, line 53: You can also do:

```suggestion
      quote location: :keep, generated: true do
```

and not have to worry about pleasing Dialyzer. :)

greg-rychlewski:
`generated: true` it is :) . this warning seems pretty unreasonable

oliver-kriska:
still same
```elixir
def decode(type) do
      quote location: :keep, generated: true do
        <<16::int32(), microseconds::int64(), days::int32(), months::int32()>> ->
          type_mod = var!(mod)
          precision = if type_mod, do: type_mod &&& unquote(@precision_mask)

          unquote(__MODULE__).decode_interval(
            microseconds,
            days,
            months,
            precision,
            unquote(type)
          )
      end
    end
```
and error:
```

deps/postgrex/lib/postgrex/type_module.ex:1084:call
The function call will not succeed.

:erlang.band(_ :: nil, 65535)

will never return since the 1st arguments differ
from the success typing arguments:

(integer(), integer())
```

josevalim:
@greg-rychlewski maybe move the band to the `decode_interval` functions?

greg-rychlewski:
@oliver-kriska sorry, one more time when you have the chance :)

greg-rychlewski:
thanks to @josevalim . i had no clue what was happening

Issue #704, opened 2024-08-26
Arrays of custom timerange types with more than one element worked in 0.18, do not work in 0.19


jdav-dev:
### Elixir version

Elixir 1.17.2 (compiled with Erlang/OTP 27)

### Database and Version

PostgreSQL 16.4

### Postgrex Version

0.19.1

### Current behavior

A minimal script to reproduce the issue:

```elixir
Mix.install([
  {:postgrex, "~> 0.19.0"}
])

{:ok, conn} =
  Postgrex.start_link(username: "postgres", password: "postgres", database: "postgres")

Postgrex.query!(conn, "DROP TABLE IF EXISTS availability;", [])
Postgrex.query!(conn, "DROP TYPE IF EXISTS timerange;", [])

Postgrex.query!(
  conn,
  """
  CREATE OR REPLACE FUNCTION time_subtype_diff(x time(0), y time(0)) RETURNS float8 AS
  'SELECT EXTRACT(EPOCH FROM (x - y))' LANGUAGE sql STRICT IMMUTABLE;
  """,
  []
)

Postgrex.query!(
  conn,
  """
  CREATE TYPE timerange AS RANGE (
    subtype = time(0),
    subtype_diff = time_subtype_diff
  );
  """,
  []
)

Postgrex.query!(conn, "CREATE TABLE availability (hours timerange[]);", [])

Postgrex.query!(conn, "INSERT INTO availability (hours) VALUES ($1);", [
  [
    struct!(Postgrex.Range,
      lower: ~T[09:00:00],
      upper: ~T[12:00:00],
      lower_inclusive: true,
      upper_inclusive: true
    ),
    struct!(Postgrex.Range,
      lower: ~T[14:00:00],
      upper: ~T[17:00:00],
      lower_inclusive: true,
      upper_inclusive: true
    )
  ]
])

Postgrex.query!(conn, "SELECT * FROM availability;", [])
|> inspect(pretty: true)
|> IO.puts()
```

The above script crashes with the following:

```
** (CaseClauseError) no case clause matching: {{Postgrex.Extensions.Time, nil}, nil}
    (postgrex 0.19.1) lib/postgrex/type_module.ex:1084: Postgrex.DefaultTypes.decode_list/2
    (postgrex 0.19.1) lib/postgrex/type_module.ex:1084: Postgrex.DefaultTypes."Elixir.Postgrex.Extensions.Range"/5
[...]

PR #705, opened 2024-08-26
Fix nested super type modifier handling

greg-rychlewski:
Closes https://github.com/elixir-ecto/postgrex/issues/704

The issue is with nested super types we may pass `{extension, modifier}` down so we end up with `{{extension, modifier}, nil}`. 

We want to flatten it out and pass `nil` for the range modifier because we don't know how the Postgres modifier for ranges works yet. Passing `nil` will have it default to what it was before we started handling modifiers.

### DIFF
--- a/test/query_test.exs
+++ b/test/query_test.exs
@@ -531,6 +531,27 @@ defmodule QueryTest do
                }
              ]
            ] = query("SELECT '[2014-1-1,2014-12-31)'::tsrange", [])
+
+    assert [
+      [
+        [
+          %Postgrex.Range{
+            lower: ~N[2014-01-01 00:00:00.000000],
+            upper: ~N[2014-12-31 00:00:00.000000],
+            lower_inclusive: true,
+            upper_inclusive: false
+          },
+          %Postgrex.Range{
+            lower: ~N[2014-01-01 00:00:00.000000],
+            upper: ~N[2014-12-31 00:00:00.000000],
+            lower_inclusive: true,
+            upper_inclusive: false
+          }
+        ]
+      ]
+    ]
+
+    query("SELECT ARRAY['[2014-1-1,2014-12-31)'::tsrange, '[2014-1-1,2014-12-31)'::tsrange]", [])
   end
 
   @tag min_pg_version: "14.0"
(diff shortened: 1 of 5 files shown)


greg-rychlewski:
I don't think this is the right solution actually. Since it happens with list of length 2 instead of 1 we are not unpacking the type somewhere we should be....will keep looking.

jdav-dev:
I can't say whether or not this is the right solution, but this branch does work for my script from #704.

test/query_test.exs
```diff
@@ -531,6 +531,27 @@ defmodule QueryTest do
...
+            lower: ~N[2014-01-01 00:00:00.000000],
+            upper: ~N[2014-12-31 00:00:00.000000],
+            lower_inclusive: true,
+            upper_inclusive: false
+          }
+        ]
+      ]
+    ]
+
+    query("SELECT ARRAY['[2014-1-1,2014-12-31)'::tsrange, '[2014-1-1,2014-12-31)'::tsrange]", [])
```
jdav-dev, line 554: I assume this assertion is to be checked against the query.

```suggestion
    assert [
             [
               [
                 %Postgrex.Range{
                   lower: ~N[2014-01-01 00:00:00.000000],
                   upper: ~N[2014-12-31 00:00:00.000000],
                   lower_inclusive: true,
                   upper_inclusive: false
                 },
                 %Postgrex.Range{
                   lower: ~N[2014-01-01 00:00:00.000000],
                   upper: ~N[2014-12-31 00:00:00.000000],
                   lower_inclusive: true,
                   upper_inclusive: false
                 }
               ]
             ]
           ] =
             query(
               "SELECT ARRAY['[2014-1-1,2014-12-31)'::tsrange, '[2014-1-1,2014-12-31)'::tsrange]",
               []
             )
```

greg-rychlewski: thank you!

jdav-dev:
> This one was a doozy. We cannot modify the variable `type` in the extensions otherwise when you have an array of super types it will pass the modified sub-type to the next member of the array. So I made it a more explicit name that shouldn't conflict.

I just came to the same conclusion.  This is the actual issue!

PR #708, opened 2024-10-11
update_error_codes

dkuku:
It was not updated for a long time in the repo but upstream last change was 2 weeks ago.
There are some new errors as you can see in the diff
https://github.com/postgres/postgres/blob/master/src/backend/utils/errcodes.txt


v0idpwn (approved):

josevalim:

PR #709, opened 2024-10-11
allow to  append comment to query

dkuku:
Recently there was introduced commit_comment to prepend comment before the commit keyword
I wanted to propose something similar for all queries - It will be useful for prepending [sqlcommenter](https://github.com/QuerySorcery/sqlcommenter) comments to queries but any other info is possible. 
I tried different hacks to get around this limitation but this has to be implemented in postgrex just before the query is sent to the database.
I found the place where to add this code by trial and error, any suggestion welcome how to make it better.
Also is it possible to test it somehow? Currently I'm logging it in the database but it would be nice to log the data that is sent to the database. Maybe testing the encoder is enough?
```
2024-10-11 21:44:02.715 BST,"kuku","postgrex_test",624085,"127.0.0.1:55774",67098e12.985d5,1,"SELECT",2024-10-11 21:44:02 BST,3/2286,0,LOG,00000,"execute <unnamed>: /*comment*/
SELECT 321",,,,,,,,,"","client backend",,0  
```

### DIFF
--- a/lib/postgrex.ex
+++ b/lib/postgrex.ex
@@ -289,6 +289,9 @@ defmodule Postgrex do
   @spec query(conn, iodata, list, [execute_option]) ::
           {:ok, Postgrex.Result.t()} | {:error, Exception.t()}
   def query(conn, statement, params, opts \\ []) do
+    prepare? = !Keyword.get(opts, :comment)
+    opts = Keyword.put(opts, :postgrex_prepare, prepare?)
+
     if name = Keyword.get(opts, :cache_statement) do
       query = %Query{name: name, cache: :statement, statement: IO.iodata_to_binary(statement)}
 
--- a/lib/postgrex.ex
+++ b/lib/postgrex.ex
@@ -363,7 +363,8 @@ defmodule Postgrex do
           {:ok, Postgrex.Query.t()} | {:error, Exception.t()}
   def prepare(conn, name, statement, opts \\ []) do
     query = %Query{name: name, statement: statement}
-    opts = Keyword.put(opts, :postgrex_prepare, true)
+    prepare? = !Keyword.get(opts, :comment)
+    opts = Keyword.put(opts, :postgrex_prepare, prepare?)
     DBConnection.prepare(conn, query, opts)
   end
 
@@ -373,7 +374,8 @@ defmodule Postgrex do
   """
   @spec prepare!(conn, iodata, iodata, [option]) :: Postgrex.Query.t()
   def prepare!(conn, name, statement, opts \\ []) do
-    opts = Keyword.put(opts, :postgrex_prepare, true)
+    prepare? = !Keyword.get(opts, :comment)
+    opts = Keyword.put(opts, :postgrex_prepare, prepare?)
     DBConnection.prepare!(conn, %Query{name: name, statement: statement}, opts)
   end
 
--- a/lib/postgrex.ex
+++ b/lib/postgrex.ex
@@ -67,6 +67,10 @@ defmodule Postgrex do
   @max_rows 500
   @timeout 15_000
 
+  @comment_validation_error Postgrex.Error.exception(
+                              message: "`:comment` option cannot contain sequence \"*/\""
+                            )
+
   ### PUBLIC API ###
 
   @doc """
@@ -294,6 +298,8 @@ defmodule Postgrex do
   @spec query(conn, iodata, list, [execute_option]) ::
           {:ok, Postgrex.Result.t()} | {:error, Exception.t()}
   def query(conn, statement, params, opts \\ []) do
+    :ok = validate_comment(opts)
+
     if name = Keyword.get(opts, :cache_statement) do
       query = %Query{name: name, cache: :statement, statement: IO.iodata_to_binary(statement)}
 
@@ -321,6 +327,20 @@ defmodule Postgrex do
     case DBConnection.prepare_execute(conn, query, params, opts) do
       {:ok, _, result} -> {:ok, result}
       {:error, _} = error -> error
+    end
+  end
+
+  defp validate_comment(opts) do
+    case Keyword.get(opts, :comment) do
+      nil ->
+        :ok
+
+      comment when is_binary(comment) ->
+        if String.contains?(comment, "*/") do
+          raise @comment_validation_error
+        else
+          :ok
+        end
     end
   end
 
--- a/lib/postgrex.ex
+++ b/lib/postgrex.ex
@@ -289,9 +289,6 @@ defmodule Postgrex do
   @spec query(conn, iodata, list, [execute_option]) ::
           {:ok, Postgrex.Result.t()} | {:error, Exception.t()}
   def query(conn, statement, params, opts \\ []) do
-    prepare? = !Keyword.get(opts, :comment)
-    opts = Keyword.put(opts, :postgrex_prepare, prepare?)
-
     if name = Keyword.get(opts, :cache_statement) do
       query = %Query{name: name, cache: :statement, statement: IO.iodata_to_binary(statement)}
 
@@ -366,7 +363,7 @@ defmodule Postgrex do
           {:ok, Postgrex.Query.t()} | {:error, Exception.t()}
   def prepare(conn, name, statement, opts \\ []) do
     query = %Query{name: name, statement: statement}
-    prepare? = !Keyword.get(opts, :comment)
+    prepare? = Keyword.get(opts, :comment) == nil
     opts = Keyword.put(opts, :postgrex_prepare, prepare?)
     DBConnection.prepare(conn, query, opts)
   end
@@ -377,7 +374,7 @@ defmodule Postgrex do
   """
   @spec prepare!(conn, iodata, iodata, [option]) :: Postgrex.Query.t()
   def prepare!(conn, name, statement, opts \\ []) do
-    prepare? = !Keyword.get(opts, :comment)
+    prepare? = Keyword.get(opts, :comment) == nil
     opts = Keyword.put(opts, :postgrex_prepare, prepare?)
     DBConnection.prepare!(conn, %Query{name: name, statement: statement}, opts)
   end
--- a/lib/postgrex.ex
+++ b/lib/postgrex.ex
@@ -169,6 +169,11 @@ defmodule Postgrex do
       This is useful when using Postgrex against systems that do not support composite types
       (default: `false`).
 
+    * `:comment` - When a binary string is provided, appends the given text as a comment to the
+      query.  This can be useful for tracing purposes, such as when using SQLCommenter or similar
+      tools to track query performance and behavior. Note that including a comment disables query
+      caching since each query with a different comment is treated as unique (default: `nil`).
+
   `Postgrex` uses the `DBConnection` library and supports all `DBConnection`
   options like `:idle`, `:after_connect` etc. See `DBConnection.start_link/2`
   for more information.
--- a/lib/postgrex/protocol.ex
+++ b/lib/postgrex/protocol.ex
@@ -346,19 +346,21 @@ defmodule Postgrex.Protocol do
   end
 
   def handle_prepare(%Query{name: ""} = query, opts, s) do
-    prepare? = Keyword.get(opts, :postgrex_prepare, false)
-    status = new_status(opts, prepare: prepare?)
+    prepare = Keyword.get(opts, :postgrex_prepare, false)
+    status = new_status(opts, prepare: prepare)
 
-    if prepare? do
-      parse_describe_close(s, status, query)
-    else
-      comment = Keyword.get(opts, :comment)
+    case prepare do
+      true ->
+        parse_describe_close(s, status, query)
 
-      if is_binary(comment) && String.contains?(comment, "*/") do
-        raise @comment_validation_error
-      else
-        parse_describe_flush(s, status, query, comment)
-      end
+      false ->
+        comment = Keyword.get(opts, :comment)
+
+        if is_binary(comment) && String.contains?(comment, "*/") do
+          raise @comment_validation_error
+        else
+          parse_describe_flush(s, status, query, comment)
+        end
     end
   end
 
@@ -371,11 +373,11 @@ defmodule Postgrex.Protocol do
     if new_query = cached_query(s, query) do
       {:ok, new_query, s}
     else
-      prepare? = Keyword.get(opts, :postgrex_prepare, false)
-      status = new_status(opts, prepare: prepare?)
+      prepare = Keyword.get(opts, :postgrex_prepare, false)
+      status = new_status(opts, prepare: prepare)
 
       result =
-        case prepare? do
+        case prepare do
           true -> close_parse_describe(s, status, query)
           false -> close_parse_describe_flush(s, status, query)
         end
@@ -1605,7 +1607,7 @@ defmodule Postgrex.Protocol do
   end
 
   defp parse_describe_comment_msgs(query, comment, tail) when is_binary(comment) do
-    statement = query.statement <> "/* #{comment} */"
+    statement = query.statement <> "/*#{comment}*/"
     query = %{query | statement: statement}
     parse_describe_msgs(query, tail)
   end
@@ -3392,7 +3394,6 @@ defmodule Postgrex.Protocol do
 
   defp msg_send(s, msgs, buffer) when is_list(msgs) do
     binaries = Enum.reduce(msgs, [], &[&2 | maybe_encode_msg(&1)])
-
     do_send(s, binaries, buffer)
   end
 
--- a/lib/postgrex/protocol.ex
+++ b/lib/postgrex/protocol.ex
@@ -371,11 +371,11 @@ defmodule Postgrex.Protocol do
     if new_query = cached_query(s, query) do
       {:ok, new_query, s}
     else
-      prepare = Keyword.get(opts, :postgrex_prepare, false)
-      status = new_status(opts, prepare: prepare)
+      prepare? = Keyword.get(opts, :postgrex_prepare, false)
+      status = new_status(opts, prepare: prepare?)
 
       result =
-        case prepare do
+        case prepare? do
           true -> close_parse_describe(s, status, query)
           false -> close_parse_describe_flush(s, status, query)
         end
--- a/lib/postgrex/protocol.ex
+++ b/lib/postgrex/protocol.ex
@@ -1607,7 +1607,7 @@ defmodule Postgrex.Protocol do
   end
 
   defp parse_describe_comment_msgs(query, comment, tail) when is_binary(comment) do
-    statement = query.statement <> "/*#{comment}*/"
+    statement = [query.statement, "/*", comment, "*/"]
     query = %{query | statement: statement}
     parse_describe_msgs(query, tail)
   end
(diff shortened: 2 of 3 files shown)


josevalim:
Unfortunately this is not going to work based on how most users use Ecto+Postgres because the queries are prepared: which means the query is only sent once to the database, the first time around. The reason why it works with BEGIN is because it is a textual command that is small and it is sent every time.

dkuku:
Thanks, I did not know that. What about adding a mode that allows to send the whole query when there is a comment? Libraries in other languages send the whole sql afaik - [python example](https://github.com/google/sqlcommenter/blob/871c1c81e014830c3f5f402a9c63d83823132542/python/sqlcommenter-python/google/cloud/sqlcommenter/__init__.py#L46), there is more in other languages in this repo.

josevalim:
We can do that and say that passing a comment automatically disables query caching, but then I beleive the implementation here has to be different. I would generally try to deal with it at the protocol level though, instead of changing the binary payloads.

dkuku:
I'm happy to work on that, I just need to better understand how postgrex works.

josevalim:
You probably want to extract out the comment only in this branch: https://github.com/elixir-ecto/postgrex/pull/709/files#diff-0da854f0c1cda9486d776c72ecda6a2e595a7667b72688669bbd80d6b80f0f96R354 and pass the comment as argument to `parse_describe_flush` after the query. You can adjust other callers to pass `nil` as the comment.

And, on Postgrex module, if there is a comment, you set postgrex_prepare to false.

dkuku:
I'm considering whether it should be named differently, like `non_cacheable_comment`. That way, we could also introduce a `cacheable_comment` in the future, which could include static values like the module and function name. This would offer a good compromise by making it easier to locate where the query is defined (since we have access to that information using Ecto macros).

test/query_test.exs
```diff
@@ -1851,6 +1851,15 @@ defmodule QueryTest do
...
 
+  test "comment", context do
+    assert [[123]] = query("select 123", [], comment: "query comment goes here")
+
+    assert_raise Postgrex.Error, fn ->
+      query("select 123", [], comment: "*/ DROP TABLE 123 --")
+    end
+  end
+
+  @tag :big_binary
```
dkuku, line 1862: My laptop wan't to suck all air from the room when I have logging enabled for all queries in postgresql and this is run.

greg-rychlewski:
Sorry if i'm way off the mark, I don't know if I completely understand the ask. But would it work to add `prepare: :unnamed` to the repo operations instead? That way Postgres is not caching the query. We allow it to be specified per operation so you can pick where it's used.

dkuku:
When we have options to skip query caching would it be enough to add the comment in the query function where we call [iodata_to_binary](https://github.com/elixir-ecto/postgrex/blob/a4b85e83c074f1c49fbf9b9d38a3875dc5580fc1/lib/postgrex.ex#L293)? instead of modifying the protocol file ?

lib/postgrex.ex
```diff
@@ -289,6 +289,9 @@ defmodule Postgrex do
   @spec query(conn, iodata, list, [execute_option]) ::
           {:ok, Postgrex.Result.t()} | {:error, Exception.t()}
   def query(conn, statement, params, opts \\ []) do
+    prepare? = !Keyword.get(opts, :comment)
+    opts = Keyword.put(opts, :postgrex_prepare, prepare?)
```
josevalim, line 293: We probably don't need this here, as it defaults to false anyway?

lib/postgrex.ex
```diff
@@ -363,7 +366,8 @@ defmodule Postgrex do
           {:ok, Postgrex.Query.t()} | {:error, Exception.t()}
   def prepare(conn, name, statement, opts \\ []) do
     query = %Query{name: name, statement: statement}
-    opts = Keyword.put(opts, :postgrex_prepare, true)
+    prepare? = !Keyword.get(opts, :comment)
```
josevalim, line 369:
```suggestion
    prepare? = Keyword.get(opts, :comment) == nil
```

lib/postgrex.ex
```diff
@@ -373,7 +377,8 @@ defmodule Postgrex do
   """
   @spec prepare!(conn, iodata, iodata, [option]) :: Postgrex.Query.t()
   def prepare!(conn, name, statement, opts \\ []) do
-    opts = Keyword.put(opts, :postgrex_prepare, true)
+    prepare? = !Keyword.get(opts, :comment)
```
josevalim, line 380:
```suggestion
    prepare? = Keyword.get(opts, :comment) == nil
```

lib/postgrex/protocol.ex
```diff
@@ -343,12 +346,19 @@ defmodule Postgrex.Protocol do
...
 
   def handle_prepare(%Query{name: ""} = query, opts, s) do
-    prepare = Keyword.get(opts, :postgrex_prepare, false)
-    status = new_status(opts, prepare: prepare)
+    prepare? = Keyword.get(opts, :postgrex_prepare, false)
+    status = new_status(opts, prepare: prepare?)
 
-    case prepare do
-      true -> parse_describe_close(s, status, query)
-      false -> parse_describe_flush(s, status, query)
```
josevalim, line 351: Can you please avoid stale changes on the PR? Please keep the variable as `prepare` and using `case` instead of `if`. This makes it easier to review and debug history later. Thank you :)

lib/postgrex/protocol.ex
```diff
@@ -361,11 +371,11 @@ defmodule Postgrex.Protocol do
...
       {:ok, new_query, s}
     else
-      prepare = Keyword.get(opts, :postgrex_prepare, false)
-      status = new_status(opts, prepare: prepare)
+      prepare? = Keyword.get(opts, :postgrex_prepare, false)
+      status = new_status(opts, prepare: prepare?)
 
       result =
-        case prepare do
+        case prepare? do
```
josevalim, line 378: Please avoid all variable naming changes in this PR.

lib/postgrex/protocol.ex
```diff
@@ -1593,6 +1604,16 @@ defmodule Postgrex.Protocol do
     transaction_error(s, postgres)
   end
 
+  defp parse_describe_comment_msgs(query, comment, tail) when is_binary(comment) do
+    statement = query.statement <> "/* #{comment} */"
```
josevalim, line 1608: Are the comments meant to come before or after? Is there any general feedback here?

josevalim:
I have dropped some nitpicks. It generally looks good to me :)

lib/postgrex/protocol.ex
dkuku: For general comments it would be better to have it in the front but for sqlcommenter tags the [specification](https://google.github.io/sqlcommenter/spec/) says that it should be at the end.
```
Format

The final comment SHOULD be affixed to the final SQL statement in the format

<SQL STATEMENT> /*<ATTRIBUTE_KEY_VALUE_PAIRS>*/
```
In the examples there is also no spaces around the attributes, I will remove it from my pr too - it's possible to add it easily when you need it but no way to remove it from the caller.

dkuku: As this is part of [opentelemetry](https://cloud.google.com/blog/products/databases/sqlcommenter-merges-with-opentelemetry) maybe we should call it tracing_comment, It would be clear why the query caching is disabled for that.

josevalim: I think as long as we add it to the docs, we should be fine.

dkuku: In this case I just updated the docs. I hope it's in correct place.

dkuku:
After thinking abount it I wonder if this is exactly what I need.
My initial idea was to append the comment after the query cache - I was under the impression that it can be done in postgrex but I was wrong.
If the query can't be cached then it may be a better idea to append it in the query builder and from there mark the query as non cacheable in case the comment is dynamic or leave it as is when the comment is static (caller module and function name may be enough for most use cases).
My concern is that disabling the query cache can increase database cpu load - it may not be a problem in small projects but in bigger it may be a concern. Datadog library for example does trace sampling - maybe there is a way to know if a trace will be sampled and only include the dynamic parts on demand otherwise allow just to tag query with the basic info and have caching working.
I can prepare an rfc how I imagine it and get feedback on ecto mailing list and the observability group.

josevalim:
> If the query can't be cached then it may be a better idea to append it in the query builder and from there mark the query as non cacheable in case the comment is dynamic or leave it as is when the comment is static (caller module and function name may be enough for most use cases).

Given queries are composable, which module and function would you use? If you want to use the module and function from the repository, then it is somewhat dynamic, but it is already information that we compute anyway, so you are right we could tackle it at the builder.

lib/postgrex/protocol.ex
```diff
@@ -1593,6 +1606,16 @@ defmodule Postgrex.Protocol do
     transaction_error(s, postgres)
   end
 
+  defp parse_describe_comment_msgs(query, comment, tail) when is_binary(comment) do
+    statement = query.statement <> "/*#{comment}*/"
```
josevalim, line 1610: It seems we may no longer go wit hthis PR, but I wonder if this could be dealt as IO lists:

```suggestion
    statement = [query.statement, "/*", comment, "*/"]
```

dkuku:
So how do we proceed with that pr?
I think it would be beneficial to still have a way to force the prepare 
`Repo.all(query, comment: "repo:MyApp.Repo,app:Myapp.Web", prepare: true)`

lib/postgrex/protocol.ex
```diff
@@ -347,8 +350,17 @@ defmodule Postgrex.Protocol do
...
     case prepare do
-      true -> parse_describe_close(s, status, query)
-      false -> parse_describe_flush(s, status, query)
+      true ->
+        parse_describe_close(s, status, query)
+
+      false ->
+        comment = Keyword.get(opts, :comment)
+
+        if is_binary(comment) && String.contains?(comment, "*/") do
```
josevalim, line 359:
```suggestion
        if is_binary(comment) and String.contains?(comment, "*/") do
```

josevalim: Let's move this validation to the client. There is no need to crash the connection on bad argument. :)

josevalim (commented):
I added one tiny nitpick and we can ship it!

lib/postgrex/protocol.ex
dkuku: So should I just remove the whole `if` here?

josevalim: You should move it to the client, to the `Postgrex` module.

josevalim:

josevalim:
Thank you for leading this. I am sure people will appreciate this feature. Please give main a try in conjuction with Ecto and let us know if it works as expected.

dkuku:
Thanks, I will try it over the weekend.

dkuku:
We missed something when skipping prepare - I need to dig deeper into it, but that's what I currently have:
I created a [repo module](https://github.com/QuerySorcery/sqlcommenter/blob/master/lib/sqlcommenter/repo.ex) that has every function replaced with a macro.
Now with such module [livebook](https://github.com/dkuku/livebooks/blob/master/ecto.livemd):
```elixir
defmodule W do
  require Test.Repo
  alias Test.Repo
  import Ecto.Query

  def insert do
    weather1 = %Weather{temp_lo: 0, temp_hi: 23}
    Repo.insert!(weather1)
  end

  def get_query do
    values = [%{id: 1, name: "Zabrze"}, %{id: 2, name: "Dudley"}]
    Weather
    |> join(:inner, [w], w1 in values(values, %{id: :integer, name: :string}), on: w.id == w1.id)
    |> select([w, c], %Weather{id: w.id, city: c.name})
    |> Repo.all()
  end

  def get_schema do
    Weather
    |> select([w], %Weather{id: w.id})
    |> where([w], not is_nil(w.id))
    |> Repo.all()
  end
end
```
And calling it one after one, I'm getting the comment only on one of the queries:
```
2024-10-27 07:24:07.266 GMT,"postgres","postgres",192987,"127.0.0.1:43910",671dea97.2f1db,1,"INSERT",2024-10-27 07:24:07 GMT,8/7,0,LOG,00000,"execute ecto_insert_weather_0: INSERT INTO ""weather"" (""temp_lo"",""temp_hi"",""prcp"",""inserted_at"",""updated_at"") VALUES ($1,$2,$3,$4,$5) RETURNING ""id""","parameters: $1 = '2', $2 = '23', $3 = '0', $4 = '2024-10-27 07:24:07', $5 = '2024-10-27 07:24:07'",,,,,,,,"","client backend",,0
2024-10-27 07:24:07.275 GMT,"postgres","postgres",192984,"127.0.0.1:43882",671dea97.2f1d8,1,"SELECT",2024-10-27 07:24:07 GMT,9/7,0,LOG,00000,"execute <unnamed>: SELECT w0.""id"", v1.""name"" FROM ""weather"" AS w0 INNER JOIN (VALUES ($1::bigint,$2::varchar),($3::bigint,$4::varchar)) AS v1 (""id"",""name"") ON w0.""id"" = v1.""id""/*app='test_app',function='get_query%2F0',line='18',module='W',owner='team_c'*/","parameters: $1 = '1', $2 = 'Zabrze', $3 = '2', $4 = 'Dudley'",,,,,,,,"","client backend",,0
2024-10-27 07:24:07.278 GMT,"postgres","postgres",192981,"127.0.0.1:43868",671dea97.2f1d5,1,"SELECT",2024-10-27 07:24:07 GMT,4/6,0,LOG,00000,"execute ecto_7179: SELECT w0.""id"" FROM ""weather"" AS w0 WHERE (NOT (w0.""id"" IS NULL))",,,,,,,,,"","client backend",,0
```
as you see only one is unnamed.
When I add the option `prepare: :unnamed` to default_options then the second select is unnamed but the insert still has a name.

josevalim:
@dkuku in Ecto, we pass a cache_statement to Postgrex.query, and I think that's overriding the `prepare: false`. Around here: https://github.com/elixir-ecto/postgrex/blob/master/lib/postgrex.ex#L303

We should probably ignore the cache statement if a comment is given.

Also, you shouldn't need to replace all Ecto.Repo functions. We already support `stacktrace: true` as part of the Repo configuration. If you set it to true, you can override `def prepare_options` (or a similar callback) to build the `comment` option. :)

dkuku:
This is getting a bit complicated, I have to pass the comment around too many function and still it is not shown everywhere.
I think much easier would be to append it to the statement asap and just properly validate that it is not prepared - for example by setting `prepare: :unnamed`?

I have also a question about the stacktrace - do you have any suggestion:
When i `use Ecto.Repo` the stacktrace that I see when inspecting in `prepare_query` has not much info about the code I'm running. There is only the line where `use Ecto.Repo` - nothing about my code.
```elixir
{:all,
 #Ecto.Query<from w0 in Weather, where: not is_nil(w0.id),
  select: %Weather{id: w0.id}>,
 [
   stacktrace: [
     {Ecto.Repo.Supervisor, :tuplet, 2, [file: ~c"lib/ecto/repo/supervisor.ex", line: 179]},
     {Test.Repo, :all, 2, [file: ~c"livemd/ecto.livemd#cell:pwz2dlcaejzj2duk", line: 2]},
     {:elixir, :eval_external_handler, 3, [file: ~c"src/elixir.erl", line: 386]},
     {:erl_eval, :do_apply, 7, [file: ~c"erl_eval.erl", line: 904]},
     {:elixir, :eval_forms, 4, [file: ~c"src/elixir.erl", line: 364]},
     {Module.ParallelChecker, :verify, 1, [file: ~c"lib/module/parallel_checker.ex", line: 112]},
     {Livebook.Runtime.Evaluator, :"-eval/4-fun-0-", 3, [file: ~c"lib/livebook/runtime/evaluator.ex", line: 635]},
     {Code, :with_diagnostics, 2, [file: ~c"lib/code.ex", line: 621]},
     {Livebook.Runtime.Evaluator, :eval, 4, [file: ~c"lib/livebook/runtime/evaluator.ex", line: 629]},
     {Livebook.Runtime.Evaluator, :continue_do_evaluate_code, 6, [file: ~c"lib/livebook/runtime/evaluator.ex", line: 437]},
     {Livebook.Runtime.Evaluator, :loop, 1, [file: ~c"lib/livebook/runtime/evaluator.ex", line: 337]},
     {:proc_lib, :init_p_do_apply, 3, [file: ~c"proc_lib.erl", line: 329]}
   ]
 ]}
```
I gave up and went back to the module I made. Then the stacktrace looks like i expect - the module is on the second line. I tested it like 20 times switching between my implementation and Ecto.
```elixir
{:all,
 #Ecto.Query<from w0 in Weather, where: not is_nil(w0.id),
  select: %Weather{id: w0.id}>,
 [
   stacktrace: [
     {Ecto.Repo.Supervisor, :tuplet, 2, [file: ~c"lib/ecto/repo/supervisor.ex", line: 179]},
     {WeatherContext, :get_schema, 0, [file: ~c"livemd/ecto.livemd#cell:4h52ky3654zcawdu", line: 32]},
     {:elixir, :eval_external_handler, 3, [file: ~c"src/elixir.erl", line: 386]},
     {:erl_eval, :do_apply, 7, [file: ~c"erl_eval.erl", line: 904]},
     {:elixir, :eval_forms, 4, [file: ~c"src/elixir.erl", line: 364]},
     {Module.ParallelChecker, :verify, 1, [file: ~c"lib/module/parallel_checker.ex", line: 112]},
     {Livebook.Runtime.Evaluator, :"-eval/4-fun-0-", 3, [file: ~c"lib/livebook/runtime/evaluator.ex", line: 635]},
     {Code, :with_diagnostics, 2, [file: ~c"lib/code.ex", line: 621]},
     {Livebook.Runtime.Evaluator, :eval, 4, [file: ~c"lib/livebook/runtime/evaluator.ex", line: 629]},
     {Livebook.Runtime.Evaluator, :continue_do_evaluate_code, 6, [file: ~c"lib/livebook/runtime/evaluator.ex", line: 437]},
     {Livebook.Runtime.Evaluator, :loop, 1, [file: ~c"lib/livebook/runtime/evaluator.ex", line: 337]},
     {:proc_lib, :init_p_do_apply, 3, [file: ~c"proc_lib.erl", line: 329]}
   ]
```
The code looks like this:
image

josevalim:
I mentioned this in my previous comment:

> Also, you shouldn't need to replace all Ecto.Repo functions. We already support stacktrace: true as part of the Repo configuration. If you set it to true, you can override def prepare_options (or a similar callback) to read the :stacktrace option and use it to build the comment option. :)

It doesn't work? That should alleviate the need to pass comments. You should only need to do it on prepare_query/prepare_options/etc.

> I have also a question about the stacktrace - do you have any suggestion

Yes, don't use the shell/livebook. Those are evaluated. Stacktrace is precise on compiled code.

milmazz:
> If you set it to true, you can override `def prepare_options` (or a similar callback) to read the `:stacktrace` option and use it to build the `comment` option. :)

@josevalim This is a really nice trick that I already applied via `prepare_query/2`. However, I'm wondering how I could accomplish the same for _transactions_ without overriding those functions. Any advice?

Given that `prepare_query/3` is not called for the `:transaction` operation in Ecto, I need to do it somewhere else, but it doesn't seem right to do something like inside `MyApp.Repo`:

```elixir
defoverridable transaction: 1, transaction: 2

def transaction(fun_or_multi, opts) do
  opts = put_lazy_commit_comment(opts)

  super(fun_or_multi, opts)
end
```

josevalim:
Yes, your hack is the only option for now, unless we introduce some sort of `prepare_transaction` (which we could).

PR #710, opened 2024-10-22
Fix start_link typespec

iacutone:
The [typespec](https://github.com/elixir-ecto/postgrex/blob/master/lib/postgrex.ex#L249) for `Postgrex.start_link/1` is incorrect. This function wraps `DBConnection.start_link/2` which has a `GenServer.on_start()` [type](https://github.com/elixir-lang/elixir/blob/v1.14.1/lib/elixir/lib/gen_server.ex#L714C34-L714C84).

I'd like to introduce an accurate type for `Postgrex.start_link/1`.

### DIFF
--- a/lib/postgrex.ex
+++ b/lib/postgrex.ex
@@ -246,7 +246,7 @@ defmodule Postgrex do
       ]
 
   """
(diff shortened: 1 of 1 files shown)


wojtekmach:
Thank you!

PR #713, opened 2024-10-29
dk-remove_comment_sql_injection_validation

dkuku:
I think we can remove the check for closing `*/` in comments, postgres does not allow to have 2 statements sent so a single `;` is enough to separate the comment.
This way we don't have to build the string and traverse it. it also enables to use iolists.
In the current implementation it is unnecessary because it's not allowed to send multiple queries, postgres throws an error in this case.

### DIFF
--- a/lib/postgrex.ex
+++ b/lib/postgrex.ex
@@ -67,10 +67,6 @@ defmodule Postgrex do
   @max_rows 500
   @timeout 15_000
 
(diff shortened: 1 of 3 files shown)


dkuku:
After thinking about it for longer it may work with postgres but some other databases may actually allow for multiple statements in a single query.

PR #714, opened 2024-11-06
Raise a nice error message of params is not a list

greg-rychlewski:
This can help if the user is passing a single value as a parameter or something like that. Otherwise the function clause error can be a bit cryptic.

### DIFF
--- a/lib/postgrex.ex
+++ b/lib/postgrex.ex
@@ -297,7 +297,7 @@ defmodule Postgrex do
   """
   @spec query(conn, iodata, list, [execute_option]) ::
           {:ok, Postgrex.Result.t()} | {:error, Exception.t()}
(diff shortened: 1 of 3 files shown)


wojtekmach:
WDYT about adding is_list(params) to Postgrex.query, execute, and similar functions? We’d still get no function clause error but I believe it would be a more relevant error message. But ArgumentError is fine too though I’d consider raising it from that query and friends, I think shorter stacktrace is preferred.

greg-rychlewski:
Good point! Changed.

greg-rychlewski:
merge conflicts =/ will fix

greg-rychlewski:
ok all better now

josevalim (approved):

PR #721, opened 2024-12-19
Allow comments as iodata.

dkuku:
When I wrote the test for the original implementation, I noticed that it couldn't be exploited when there was a `;` at the end of the original statement. Postgres returned an error stating, `cannot insert multiple commands into a prepared statement.` Initially, I abandoned this idea as I didn't want to be the person who introduced SQLI into Postgrex. However, the more I think about it, the more useful it seems to remove the binary requirement for scanning. Without validation, we can use Iodata. 

I sought advice and even posted a potentially vulnerable app on a security forum. I also conducted a self-scan using SQLMap, and it seems safe. Golang PGX is using this. 

The only issue I found was at https://adeadfed.com/posts/postgresql-select-only-rce/, but I don't think it can be used after `;` looking at the example. I tried but no luck.
Also sqlcommenter specification require to escape the string.

### DIFF
--- a/lib/postgrex.ex
+++ b/lib/postgrex.ex
@@ -67,9 +67,7 @@ defmodule Postgrex do
   @max_rows 500
   @timeout 15_000
 
-  @comment_validation_error Postgrex.Error.exception(
-                              message: "`:comment` option cannot contain sequence \"*/\""
-                            )
+  @comment_validation_error Postgrex.Error.exception(message: "`:comment` not an iodata term")
 
   ### PUBLIC API ###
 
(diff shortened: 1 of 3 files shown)


dkuku:
my example app:
https://sqlcomm.fly.dev/?id=*%2Fselect+*+from+users+%2F

josevalim:
I would prefer to keep it as a binary, in case we need to do checks in the future. This is really something we don't want to get wrong and we want to have space to fix it in the future. For example, I am almost sure we should check for null bytes too. There is basically no reason to have either null bytes or */ in a comment. I think appending `;` is a good idea anyway.

dkuku:
null byte terminates the message but postgres requires it to be certain length:
```
              postgres: %{
                code: :protocol_violation,
                message: "insufficient data left in message",
                file: "pqformat.c",
```

josevalim:
Yup! It should be null byte safe today. But it is also easy for us to make changes in the future, I don't know, maybe use text for some queries, and then completely forget we assumed we are null byte safe here. :)

PR #722, opened 2024-12-19
check_for_null_in_comment

dkuku:
I'm not sure if it makes sense to precompile binary pattern for that?


josevalim:
It only makes sense to precompile if we can store it somewhere. That’s not the case here so we should be fine.

josevalim:

PR #723, opened 2025-01-11
Update dependencies with warnings

TylerWitt:
I was digging around to ensure that JSON is supported, and saw a lot of deprecations when compiling deps.

I didn't bump `decimal` even though it also has warnings because I'm not sure if we want to bump to v2 only (now that Jason supports v2, which I'm assuming is why v1 was also supported), or if we want to keep testing the v1 version of the library for some reason.

I know it's a super minor change, though. Feel free to close!


josevalim:
Thank you, although CI seems to be unhappy :)

TylerWitt:
> Thank you, although CI seems to be unhappy :)

Yeah, I'm trying to check it out. I'm assuming it has to do with the db_connection changes.

TylerWitt:
The 1.17 branch seems to be an error that might have been fixed by Elixir 1.17.2
```
15:08:50.775 [error] Failure while translating Erlang's logger event\n** (MatchError) no match of right hand side value: 
%{callback_mode: :handle_event_function, client_info: :undefined, label: {:gen_statem, :terminate}, log: [], 
modules: [DBConnection.Connection], name: #PID<0.1902.0>, postponed: [], queue: [internal: {:connect, :init}], 
reason: {:error, %ArgumentError{message: \"expected :search_path to be a list of strings, got: \\\"public, extension\\\"\"}
```

TylerWitt:
From what I can tell, [this line](https://github.com/elixir-ecto/db_connection/blob/93174d09695ef3cf6fecb9e6cf0505a62fc89434/lib/db_connection/connection.ex#L78) in db_connection makes the genstatem pattern incompatible with Elixir versions before 1.17.2 when raise is used during a callback, since it will enter `gen_statem`'s internal terminate, which generates the log format that was fixed in 1.17.2.

Assuming that's correct, raising in a gen_statem callback when the Elixir version is lower than 1.17.2 means the errors will crash the logger.

What should we do in terms of Postgrex? Do we
* Bump the Elixir version of this library to be a min of 1.17.2?
* Prevent Postgrex from using the db_connection versions that implement genstatem?
* Revert genstatem in db_connection, and continue maintaining the `connection` library?
* Patch the logger fix for Elixir into each of the Elixir minor versions?

Or is this simply not a big deal?

josevalim:
Users should update from 1.17.2, as guarantee bug compatibility everywhere else is not worth it, IMO. :)

josevalim:
To expand a bit more, other libraries with `gen_statem` will have the same issue, so the correct thing is to update Elixir. :) Which is what we should do on CI here!

mix.exs
```diff
@@ -8,7 +8,7 @@ defmodule Postgrex.Mixfile do
     [
       app: :postgrex,
       version: @version,
-      elixir: "~> 1.11",
+      elixir: "~> 1.17",
```
josevalim, line 11: To be clear, we shouldn't change this requirement. Users running on an Elixir version with a bug should upgrade their Elixir, we shouldn't force people on 1.15 (still supported) to update because of that. :) We tend to be conservative with those. :)

```suggestion
      elixir: "~> 1.11",
```

.github/workflows/ci.yml
```diff
@@ -44,14 +44,14 @@ jobs:
           - version: 14
 
         pair:
-          - elixir: 1.11.4
-            otp: 23.3.3
+          - elixir: 1.17.2
+            otp: 25.3
```
josevalim, line 48:
```suggestion
          - elixir: 1.11.4
            otp: 23.3.3
```

TylerWitt: Ah, I made this change because the test won't pass below 1.17.2

test/postgrex_test.exs
```diff
@@ -1,30 +1,34 @@
...
 
-    # invalid argument
-    Process.flag(:trap_exit, true)
-    search_path = "public, extension"
+    test "start_link/2 sets search path" do
+      # valid argument
+      search_path = ["public", "extension"]
+      {:ok, conn} = Postgrex.start_link(database: "postgrex_test", search_path: search_path)
+      %{rows: [[result]]} = Postgrex.query!(conn, "show search_path", [])
+      assert result == Enum.join(search_path, ", ")
```
greg-rychlewski, line 13: We can probably separate the happy path into its own test so it won't be ignored on lower versions.

test/postgrex_test.exs
```diff
@@ -1,30 +1,34 @@
...
-  import ExUnit.CaptureLog
 
-  test "start_link/2 sets search path" do
-    # valid argument
-    search_path = ["public", "extension"]
-    {:ok, conn} = Postgrex.start_link(database: "postgrex_test", search_path: search_path)
-    %{rows: [[result]]} = Postgrex.query!(conn, "show search_path", [])
-    assert result == Enum.join(search_path, ", ")
+  # This test fails due to a bug betweem Elixir and Erlang in earlier versions of Elixir.
+  if Version.match?(Version.parse!(System.version()), Version.parse_requirement!(">= 1.17.2")) do
```
greg-rychlewski, line 5: We have a simplified version of this in Ecto:

`if Version.match?(System.version(), ">= 1.13.0-dev") `

test/postgrex_test.exs
TylerWitt: Done! Good call.

test/postgrex_test.exs
```diff
@@ -1,30 +1,35 @@
...
     {:ok, conn} = Postgrex.start_link(database: "postgrex_test", search_path: search_path)
     %{rows: [[result]]} = Postgrex.query!(conn, "show search_path", [])
+
     assert result == Enum.join(search_path, ", ")
+  end
 
-    # invalid argument
-    Process.flag(:trap_exit, true)
-    search_path = "public, extension"
+  # This test fails due to a bug betweem Elixir and Erlang in earlier versions of Elixir.
```
josevalim, line 13: I think there are two issues here:

1. Previous version did not log through Elixir logger
2. v1.17.0 and v1.17.1 broke the logging

```suggestion
  # gen_statem reports are only captured on Elixir v1.17+
  # but a bug causes the Logger to crash on v1.17.0 and v1.17.1.
```

Does this capture it fully?

josevalim:

PR #725, opened 2025-01-18
replication connection respects reconnect_backoff option

DohanKim:
Replication Connection's {:timeout, :backoff} action does not wait for the time as reconnect_backoff option set.

### DIFF
--- a/lib/postgrex/replication_connection.ex
+++ b/lib/postgrex/replication_connection.ex
@@ -487,6 +487,7 @@ defmodule Postgrex.ReplicationConnection do
   def handle_event(type, content, state, s)
 
   def handle_event({:timeout, :backoff}, nil, @state, s) do
+    Process.sleep(s.reconnect_backoff)
     {:keep_state, s, {:next_event, :internal, {:connect, :backoff}}}
   end
 


lib/postgrex/replication_connection.ex
```diff
@@ -487,6 +487,7 @@ defmodule Postgrex.ReplicationConnection do
   def handle_event(type, content, state, s)
 
   def handle_event({:timeout, :backoff}, nil, @state, s) do
+    Process.sleep(s.reconnect_backoff)
```
josevalim, line 490: The state machine (gen_statem) has an API for sleeping, so that's where it happens. Roughly around line 500:

    {:keep_state, s, {{:timeout, :backoff}, s.reconnect_backoff, nil}}

DohanKim: Sorry, I'm not familiar with gen_statem so I'm reading the code now.
Are you referring to this?
https://github.com/erlang/otp/blob/OTP-27.2/lib/stdlib/src/gen_statem.erl#L1188

But anyway when I test after tweaking reconnect_backoff to a huge number, 
ReplicationConnection still immediately tries to reconnect without any delay.

josevalim:
Thanks @DohanKim, if it is indeed not triggering, we should address it by using the `gen_statem` APIs.

@whatyouhide can you give us a hand with this one? You have better `gen_statem` knowledge than I do.

lib/postgrex/replication_connection.ex
greg-rychlewski: Looking here https://www.erlang.org/doc/system/statem.html#state-callback

Does it have to be like this

```
{:keep_state, s, [{{:timeout, :backoff}, s.reconnect_backoff, nil}]}
```

DohanKim:
Sorry 
I confused 
- failing to make a connection with Postgres
- failing to start a replication slot

The timeout is working as expected. 
I'm closing the PR.

PR #717, opened 2024-11-12
call disconnect on protocol when reconnecting in simple connection

artemeff:
postgrex 0.17.5
oban 2.18.3

we are using oban with postgres notifiers and in a few services we have a port leaks

image

_blue line is fix hotload_

the problem is that the ports are not closed on `{:connect, :reconnect}` handle, it just drops the `protocol` state and creates a new one

```elixir
08:09:21.320006 <0.3785.0> Postgrex.SimpleConnection.handle_event(:info, {:tcp, #Port<0.84514854>,
 <<69, 0, 0, 0, 115, 83, 70, 65, 84, 65, 76, 0, 86, 70, 65, 84, 65, 76, 0, 67,
   53, 55, 80, 48, 53, 0, 77, 116, 101, 114, 109, 105, 110, 97, 116, 105, 110,
   103, 32, 99, 111, 110, 110, 101, 99, 116, 105, ...>>}, :no_state, %Postgrex.SimpleConnection{
  idle_interval: 5000,
  protocol: %Postgrex.Protocol{
    sock: {:gen_tcp, #Port<0.84514854>},
    connection_id: 3537365,
    connection_key: 904133435,
    peer: {{10, 129, 1, 100}, 5432},
    types: {Postgrex.DefaultTypes, #Reference<0.1812357926.8519733.135683>},
    null: nil,
    timeout: 15000,
    ping_timeout: 15000,
    parameters: #Reference<0.1812357926.1244921861.105622>,
    queries: nil,
    postgres: :idle,
    transactions: :naive,
    buffer: :active_once,
    disconnect_on_error_codes: [],
    scram: ...,
    disable_composite_types: false
  },
  auto_reconnect: true,
  reconnect_backoff: 500,
  state: {Oban.Notifiers.Postgres,
   %Oban.Notifiers.Postgres{
     conf: %Oban.Config{
       dispatch_cooldown: 5,
       engine: Oban.Engines.Basic,
       get_dynamic_repo: nil,
       insert_trigger: true,
       log: false,
       name: Oban,
       node: "service@172.18.0.1",
       notifier: {Oban.Notifiers.Postgres, []},
       peer: {Oban.Peers.Postgres, []},
       plugins: [{Oban.Plugins.Pruner, [max_age: 43200]}],
       prefix: "public",
       queues: [default: [limit: 10]],
       repo: Service.Repo,
       shutdown_grace_period: 15000,
       stage_interval: 1000,
       testing: :disabled
     },
     from: nil,
     key: nil,
[...]

PR #726, opened 2025-01-22
Call disconnect on protocol when reconnecting in Replication connection 

DohanKim:
### Problems
- When the ReplicationConnection encounters a socket error, it reconnects to another socket without properly closing the current one (similar to #717).
- When the ReplicationConnection fails to establish a connection, it doesn’t log any errors and simply retries indefinitely.

### Worst Case Scenario
If Postgres terminates a replication connection process due to an OOM issue or a network error:

- Postgrex creates a new ReplicationConnection without closing the previous one.
- This repeats up to 10 times. (with auto_reconnect option)
- The max_wal_senders limit (default: 10) is reached, preventing new ReplicationConnection from being created.
- The system enters a silent infinite loop of reconnecting without any logs.

This issue has occurred in my production app during periods of network instability.

Screenshot 2025-01-22 at 6 39 53 PM

## Commit Changes
- Invoke disconnect on the protocol when reconnecting to a ReplicationConnection.
- Log an error message when a ReplicationConnection fails to establish a connection.

### DIFF
--- a/lib/postgrex/replication_connection.ex
+++ b/lib/postgrex/replication_connection.ex
@@ -490,6 +490,12 @@ defmodule Postgrex.ReplicationConnection do
     {:keep_state, s, {:next_event, :internal, {:connect, :backoff}}}
   end
 
+  def handle_event(:internal, {:connect, :reconnect}, @state, %{protocol: protocol} = state)
+      when protocol != nil do
+    Protocol.disconnect(:reconnect, protocol)
+    {:keep_state, %{state | protocol: nil}, {:next_event, :internal, {:connect, :init}}}
+  end
+
   def handle_event(:internal, {:connect, _info}, @state, %{state: {mod, mod_state}} = s) do
     case Protocol.connect(opts()) do
       {:ok, protocol} ->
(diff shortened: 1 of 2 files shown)


test/.DS_Store
josevalim, line 1: Please remove this file :)

lib/postgrex/replication_connection.ex
```diff
@@ -490,12 +490,22 @@ defmodule Postgrex.ReplicationConnection do
...
+    {:keep_state, %{state | protocol: nil}, {:next_event, :internal, {:connect, :init}}}
+  end
+
   def handle_event(:internal, {:connect, _info}, @state, %{state: {mod, mod_state}} = s) do
     case Protocol.connect(opts()) do
       {:ok, protocol} ->
         maybe_handle(mod, :handle_connect, [mod_state], %{s | protocol: protocol})
 
       {:error, reason} ->
+        Logger.error(
```
josevalim, line 505: Can you please also add this exception to SImpleConnection for consistency?

DohanKim: Added

PR #728, opened 2025-02-01
Have Duration return same units as Postgrex.Interval

greg-rychlewski:
Based on our discussion

### DIFF
--- a/README.md
+++ b/README.md
@@ -43,7 +43,7 @@ iex> Postgrex.query!(pid, "INSERT INTO comments (user_id, text) VALUES (10, 'hey
 | `timestamp`        | `%NaiveDateTime{year: 2013, month: 10, day: 12, hour: 0, minute: 37, second: 14}`                                                           |
 | `timestamptz`      | `%DateTime{year: 2013, month: 10, day: 12, hour: 0, minute: 37, second: 14, time_zone: "Etc/UTC"}` (2)                                      |
 | `interval`         | `%Postgrex.Interval{months: 14, days: 40, secs: 10920, microsecs: 315}`                                                                     |
-| `interval`         | `%Duration{year: 1, month: 2, week: 5, day: 5, hour: 3, minute: 2, second: 0, microsecond: {315, 6}}` (3)                                   |
+| `interval`         | `%Duration{month: 2, day: 5, second: 0, microsecond: {315, 6}}` (3)                                   |
 | `array`            | `[1, 2, 3]`                                                                                                                                 |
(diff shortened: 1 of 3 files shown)


endoooo:
should it be reflected in [README](https://github.com/elixir-ecto/postgrex/blob/a6f20205a37ac5887cf47eaba403974a95f9cc3f/README.md?plain=1#L46)? something like:

| PostgreSQL | Elixir |
| - | - |
| `interval`| `%Duration{month: 2, day: 5, second: 0, microsecond: {315, 6}}` |

greg-rychlewski:
good call =)

josevalim (approved):

greg-rychlewski:
I will try to find the discussion when I get home. I think it was in another PR. But the gist is that Postgres wire protocol cannot support the same units as elixir duration. And we have to pick a strategy. One is do the same as the old interval struct (which we chose). Another is to fill up the bigger units first and then fill with 0. We started with the former and changed it to the latter based on a discussion with a user

greg-rychlewski:
We got lucky the discussion was easy to find https://github.com/elixir-ecto/postgrex/issues/688#issuecomment-2628584155

Wigny:
Got it, thanks! 

While I don't see an issue with the approach of returning the same units as Postgres (as `PT25M` and `PT1500S` should be the same anyway), I wonder if should we have a more friendly way of ensuring these two values are indeed the same (think of when testing insertions, for example)... Maybe having a `Duration.compare/2` would solve it, but also [rescaling the units](https://moment.github.io/luxon/api-docs/index.html#durationrescale) would suppress this needing (?). 

Suggestion on how to tackle it? is creating a custom type the way to go or do you think worth suggesting the additions of `Duration.compare/2` in the standard library?

greg-rychlewski:
It looks like the docs for duration have some suggestions for comparing so I’m not sure such a function would be accepted into the standard lib. @josevalim is Amin a better position to speak about that though.

In terms of a custom ecto type, I probably wouldn’t do it just for comparing equality. That could probably be handled with a helper function. At least I cannot think of any reason right now to do that.

Wigny:
> It looks like the docs for duration have some suggestions for comparing

Oh, nice.

> I probably wouldn’t do it just for comparing equality.

Yeah, I figured also that having a helper would be simpler. 

Many thanks.

greg-rychlewski:
No worries, thank you for raising this. In case you run into it later, the same kind of thing happens with ranges. Postgres will normalize the values you give and then you lose the original ones

PR #729, opened 2025-02-02
Search path is already supported by parameters option

greg-rychlewski:
A while ago we added a `search_path` option to start_link. But looking closer I saw it was already supported by the `parameters` option. 

I think we have to continue letting people use `search_path` so their apps don't break. But internally I'm changing it to re-use `parameters` so that we don't send an extra query unnecessarily on startup.

### DIFF
--- a/lib/postgrex.ex
+++ b/lib/postgrex.ex
@@ -52,7 +52,6 @@ defmodule Postgrex do
           | {:prepare, :named | :unnamed}
           | {:transactions, :strict | :naive}
           | {:types, module}
-          | {:search_path, [String.t()]}
           | {:disconnect_on_error_codes, [atom]}
           | DBConnection.start_option()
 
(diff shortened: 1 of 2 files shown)


josevalim:
Let's deprecate the one at the root then? And tell them to use parameters? :)

josevalim (approved):

greg-rychlewski:
@josevalim Does this look ok? I saw we used `logger.warning` instead of `io.warn` for the recent ssl deprecation so I did the same here.

josevalim (approved):

greg-rychlewski:
@josevalim I should have asked this a while ago. I see that `IO.warn` can be useful for compilation warnings and getting the stack trace. For runtime errors is there any reason to use `IO.warn` instead of `Logger.warning`?

josevalim:
Yes, IO.warn would be preferred for this style of warnings in general, since it also works with --warnings-as-errors, but in this case it doesn't matter much because it by definition happens outside of compilation, since we are starting processes. Both are fine IMO.

PR #736, opened 2025-04-11
Add example with ANY()

reisub:
I had the same question as the person in this issue: https://github.com/elixir-ecto/postgrex/issues/311

I didn't find an example in the docs, so I thought I'd add one.


wojtekmach:
Thanks, Dino!

PR #737, opened 2025-04-18
Added pgvector support

nickkaltner:
This just adds support for the pgvector extension into the types for postgrex


josevalim:
Hi @nickkaltner, in this project we only support extensions that ship as part of PostgreSQL. pgvector itself has official support for Elixir here: https://github.com/pgvector/pgvector-elixir

nickkaltner:
Ok sorry to hear that.  That extension didn't work properly so i thought i'd kill birds with one stone :) no worries, i'll submit a patch there :) Really keen to get elixir some good vector search support :)

josevalim:
A PR there would be best indeed, thanks!

PR #742, opened 2025-05-13
Prioritize explicit endpoint options over PGHOST-derived defaults

davoclavo:
Hello! First of all, thank you for this amazing library

This PR addresses a bug where setting the `PGHOST` environment variable can unintentionally override explicitly passed connection options like `:hostname`, `:socket_dir`, `:socket`, or `:endpoints`.

## Summary

- Fix: Updated `default_opts/1` to derive the host from `PGHOST` **only if** no endpoint-related options are explicitly provided (`:socket`, `:socket_dir`, `:hostname`, or `:endpoints`).
- Test: Added test coverage to ensure that explicitly provided connection options correctly take precedence over environment-derived defaults.
- Important note: This might cause breaking changes to existing users if they might be inadvertently overriding values that are currently ignored.

## Motivation

When `PGHOST` is set (e.g., to a Unix socket path), Postgrex currently prioritizes it over explicitly passed options like `:hostname`. This causes surprising behavior: users trying to connect to a remote host (e.g., localhost over TCP) may find Postgrex silently attempting to connect over a non-existent socket.

This PR brings Postgrex closer in behavior to `psql`, which properly prioritizes CLI-supplied options over `PGHOST`.

---

## Replication Instructions

You can reproduce the issue as follows:

1. Set `PGHOST` to any path:
    ```sh
    PGHOST=/whatever iex -S mix
    ```
   
2. Attempt to start a connection using valid TCP settings:

    Expected: Should connect over TCP to `localhost`
    Actual: Fails, as Postgrex attempts to connect via `/whatever/.s.PGSQL.5432` socket specified via PGHOST.

    ```elixir
    {:ok, pid} = Postgrex.start_link(
      hostname: "localhost",
      username: "postgres",
      password: "postgres",
      database: "postgres",
      port: 5432
    )
    
    Postgrex.query(pid, "SELECT 1", [])
    ```

3. Current workaround (demonstrating expected behavior with a workaround using manual override to `socket_dir: nil`):

    ```elixir
    {:ok, pid} = Postgrex.start_link(
      hostname: "localhost",
      username: "postgres",
      password: "postgres",
      database: "postgres",
      port: 5432,
      socket_dir: nil
    )
    
    Postgrex.query(pid, "SELECT 1", [])
    ```

## Behavior Comparison with `psql`

To reinforce expected behavior, here’s how `psql` treats similar inputs:

Explicit `-h localhost` takes precedence over `PGHOST`:

```sh
PGHOST=/whatever psql -d postgres -h localhost -U postgres -c "SELECT 1"
```

This should fail as no `-h` is provided and the `PGHOST` socket doesn't exist

```sh
PGHOST=/whatever psql -d postgres -U postgres -c "SELECT 1"
```

### DIFF
--- a/lib/postgrex/utils.ex
+++ b/lib/postgrex/utils.ex
@@ -84,16 +84,22 @@ defmodule Postgrex.Utils do
 
   @doc """
   Fills in the given `opts` with default options.
+  Only adds keys extracted via PGHOST if no endpoint-related keys are explicitly provided.
   """
   @spec default_opts(Keyword.t()) :: Keyword.t()
   def default_opts(opts) do
     {field, value} = extract_host(System.get_env("PGHOST"))
 
+    endpoint_keys = [:socket, :socket_dir, :hostname, :endpoints]
+    has_endpoint? = Enum.any?(endpoint_keys, &Keyword.has_key?(opts, &1))
+
     opts
     |> Keyword.put_new(:username, System.get_env("PGUSER") || System.get_env("USER"))
     |> Keyword.put_new(:password, System.get_env("PGPASSWORD"))
     |> Keyword.put_new(:database, System.get_env("PGDATABASE"))
-    |> Keyword.put_new(field, value)
+    |> then(fn opts ->
+      if has_endpoint?, do: opts, else: Keyword.put(opts, field, value)
+    end)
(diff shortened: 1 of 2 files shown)


josevalim:

PR #744, opened 2025-06-04
Add ssl_negotiation_direct option

GriffinHeart:
Adds `ssl_negotiation_direct` option

Added in Postgresql 17 libpq and client drivers that use it already support this option, it was introduced to save a round trip and also help with networks and infrastructure setups that only allow TLS traffic. More common in corporate networks or cloud environments (for example AWS Privatelinks).

Without this option it is necessary to run a proxy to deal with the upgrade of the connection to TLS before contacting the Posgres server.

This is my first contribution here and to an elixir lib, so let me know if you want me to change/add anything and if its appropriate for inclusion.


GriffinHeart:
Let me close this it needs more testing

PR #745, opened 2025-06-05
Add support for the :json library

aseigo:
This PR adds support for the recently added built-in :json library.

It does this by checking to see if `:json` is available at runtime, and then binding the correct encode/decode functions which, unfortunately, vary  between libraries.

This PR allows one to drop a dependency when using Postgrex.

### DIFF
--- a/lib/postgrex/extensions/jsonb.ex
+++ b/lib/postgrex/extensions/jsonb.ex
@@ -3,12 +3,13 @@ defmodule Postgrex.Extensions.JSONB do
   import Postgrex.BinaryUtils, warn: false
 
   def init(opts) do
-    json =
+    json_library =
       Keyword.get_lazy(opts, :json, fn ->
-        Application.get_env(:postgrex, :json_library, Jason)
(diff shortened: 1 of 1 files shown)


lib/postgrex/extensions/jsonb.ex
```diff
@@ -20,30 +21,50 @@ defmodule Postgrex.Extensions.JSONB do
...
+      # NOTE: it is not possible to see if the module is loaded, since it may
+      # not be as it is only loaded on first use. So we try to use it with apply
+      # which throws an exception if it doesn't not exist without also emitting
+      # warnings to the logger.
+      apply(&:json.encode/1, [""])
+      :json
+    rescue
+      _ -> Jason
+    end
+  end
```
Wigny, line 68: What about defining the default JSON library in compile time like done [in Cldr](https://github.com/elixir-cldr/cldr/commit/35ae66b08946b28b8b4fef0cb99a2a2a837d348d)?

PR #755, opened 2025-11-28
Make timeouts for TypeServer configurable

Th3-M4jor:
We were testing out YugabyteDB with nodes in Australia, France, and the Eastern US and found that Postgrex connections were consistently failing to startup when connecting to nodes in France or Australia if we wanted to have a pool size greater than one.

The root cause is that the bootstrap queries Postgrex makes are taking a lot longer than the 60 seconds the TypeServer allows while a connection is waiting to get the lock.

This is because the first query of a connection forces the node to cache cluster information from the leader node, which can take a long time when the leader is on the opposite side of the world.

Yugabyte has "smart" drivers that are able to work around this issue, but there isn't one in Elixir so the simplest solution would be us being able to raise this timeout to a much higher number.

### DIFF
--- a/lib/postgrex/type_server.ex
+++ b/lib/postgrex/type_server.ex
@@ -5,8 +5,6 @@ defmodule Postgrex.TypeServer do
 
   defstruct [:types, :connections, :lock, :waiting]
 
-  @timeout 60_000
-
   @doc """
   Starts a type server.
   """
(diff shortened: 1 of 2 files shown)


lib/postgrex/type_server.ex
```diff
@@ -5,7 +5,7 @@ defmodule Postgrex.TypeServer do
 
   defstruct [:types, :connections, :lock, :waiting]
 
-  @timeout 60_000
+  @timeout Application.compile_env(:postgrex, :type_server_timeout, 60_000)
```
josevalim: I don't think it needs to be a compile time setting. Let's move it to runtime. Also, move the default value to the `def application` section in the `mix.exs`, this way we can use `fetch_env!` here. Thanks!

PR #761, opened 2026-01-31
Add text query support

### LINKED ISSUE #757: Add support for text queries
### Elixir version

All

### Database and Version

All

### Postgrex Version

All

### Current behavior

I think it would be worth it to add support for the text protocol the same way Myxql does. I started thinking about this when responding to this google group question: https://groups.google.com/g/elixir-ecto/c/mooS_UwcuwU

The benefits I’m seeing:

- It will allow people to do one off queries like the google group question using the existing query functions in both Postgrex and Ecto. So no need to set up a callback module. It looks to me like both simple and extended protocol can be used on the same connection 
- It’s not completely trivial to set up a new extension for the binary protocol. You have to go into the Postgres code and copy what they do. This would let people create custom ecto types and use the formats they are already familiar with through psql 

In my head this should not be too difficult. We would just need a new struct for text queries and then the execute callback pattern matches on it and does what simple connection does today. I haven’t looked in detail though so maybe I’ll regret this once I start.

### Expected behavior

Something like myxql

greg-rychlewski:
Closes https://github.com/elixir-ecto/postgrex/issues/757

We didn't discuss this so I'm completely ok discarding it if the decision is to not add it.

The idea came from a google group message here: https://groups.google.com/g/elixir-ecto/c/mooS_UwcuwU. Basically they are trying to use Postgres's explain generic plan statement (e.g. `EXPLAIN (GENERIC_PLAN) SELECT * FROM pg_class WHERE relname = $1;`) but it does not play nice with the binary protocol because it thinks you need to supply $1 instead of it just being  a signal that it's a generic query that you want explained.

Aside from the one specific use case, they also mentioned they are using livebook. And that got me thinking it might be cool to be able to use livebook like an alternative to psql where you can just type in your queries and not worry about binary types and extensions and all that.

I didn't add query_many because it's a bit weird and I would have to think about the implementation. The reason it's weird is because it would only work for the text protocol. So I didn't want to think of a solution if we will not go this direction at all.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -4,6 +4,7 @@ on:
   push:
     branches:
       - master
+      - text_query
   pull_request:
 
 jobs:
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -4,7 +4,6 @@ on:
   push:
     branches:
       - master
-      - text_query
   pull_request:
 
 jobs:
--- a/lib/postgrex.ex
+++ b/lib/postgrex.ex
@@ -23,7 +23,7 @@ defmodule Postgrex do
   you need to start a separate (notifications) connection.
   """
 
-  alias Postgrex.Query
+  alias Postgrex.{Query, TextQuery}
 
   @typedoc """
   A connection process name, pid or reference.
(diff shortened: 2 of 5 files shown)


josevalim (approved):
I am very happy to ship it as is. If you don't want to do query_many, you just need to remove TextQueries :)

KirillKayumov:
Hi @josevalim, is it possible to release a new version in Hex that contains changes from the PR?. Last 0.20.0 release was done in January.

josevalim:
Didn’t we ship 0.22 in March?

KirillKayumov:
> Didn’t we ship 0.22 in March?

@josevalim I thought so looking at the commits in master but in [hex](https://hex.pm/packages/postgrex/versions) v0.20.0 is pushed in January. And I've just checked it once again in our project – 0.22.0 doesn't have these changes.

josevalim:
Ah, it seems v0.22 was rebased and merged on top. We can ship v0.23 later! Please ping me in a week once I am back from holidays, meanwhile give main a try :)

PR #764, opened 2026-03-12
Set process labels

nathanl:
This makes it easier to understand the process tree in tools like Observer and Phoenix LiveDashboard and gives clearer errors for crashes.

CleanShot 2026-03-12 at 16 46 18@2x

**Update** - after discussion, I updated this label to include only the database name, in case the rest is sensitive.

### DIFF
--- a/lib/postgrex/protocol.ex
+++ b/lib/postgrex/protocol.ex
@@ -229,6 +229,7 @@ defmodule Postgrex.Protocol do
     with {:ok, database} <- fetch_database(opts),
          status = %{status | types_key: if(types_mod, do: {host, port, database})},
          {:ok, ret} <- connect_and_handshake(host, port, sock_opts, timeout, s, status) do
+      Postgrex.Utils.set_label({Postgrex.Protocol, database})
       {:ok, ret}
     else
       {:error, err} ->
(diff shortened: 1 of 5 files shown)


lib/postgrex/protocol.ex
```diff
@@ -229,6 +229,7 @@ defmodule Postgrex.Protocol do
     with {:ok, database} <- fetch_database(opts),
          status = %{status | types_key: if(types_mod, do: {host, port, database})},
          {:ok, ret} <- connect_and_handshake(host, port, sock_opts, timeout, s, status) do
+      Postgrex.Utils.set_label({Postgrex.Protocol, host, port, database})
```
josevalim: Can we have anything confidential in the host? 🤔

nathanl: Hmm. Hostname could be something like `~c"localhost"` or `~c"db.example.com"` or `~c"10.0.1.42"` or `{:local, "/var/run/postgresql/.s.PGSQL.5432"}` for a Unix socket.

Host and port might both be considered sensitive, depending on the user and environment, and they could end up in crash logs on a third party service. OTOH they might be useful debugging info.

We could make the contents of the label configurable, but that's extra complexity and probably nobody will bother.

Maybe we should label this with just `{Postgrex.Protocol, database}`?

josevalim: Let’s go with the database for V1, yeah.

nathanl: Updated

nathanl:
@josevalim 👀

nathanl:
You might also be interested in seeing [Add a nested list view of application processes](https://github.com/phoenixframework/phoenix_live_dashboard/pull/487)

josevalim:

PR #765, opened 2026-03-18
Return proper error when getting tcp closed after fatal errors

v0idpwn:
When PostgreSQL sends a FATAL or PANIC ErrorResponse, it closes the connection immediately without sending ReadyForQuery.

Postgrex unconditionally waited for ReadyForQuery, which would hit tcp_closed and return a generic disconnect error, discarding the original error message.

I've included in the PR a test that would fail with the previous implementation. The query would return a
```elixir
{:error,
   %DBConnection.ConnectionError{
     message: "tcp recv: closed",
     severity: :error,
     reason: :closed
   }}
```

### DIFF
--- a/lib/postgrex/protocol.ex
+++ b/lib/postgrex/protocol.ex
@@ -3068,20 +3068,19 @@ defmodule Postgrex.Protocol do
     end
   end
 
-  defp error_ready(s, _status, %Postgrex.Error{postgres: %{severity: severity}} = err, buffer)
-       when severity in ["FATAL", "PANIC"] do
-    %{connection_id: connection_id} = s
-    {:disconnect, %{err | connection_id: connection_id}, %{s | buffer: buffer}}
-  end
-
   defp error_ready(s, status, %Postgrex.Error{} = err, buffer) do
+    %{connection_id: connection_id} = s
+
     case recv_ready(s, status, buffer) do
       {:ok, s} ->
-        %{connection_id: connection_id} = s
         {:error, %{err | connection_id: connection_id}, s}
 
       {:disconnect, _, _} = disconnect ->
-        disconnect
+        if err.postgres.severity in ["FATAL", "PANIC"] do
+          {:disconnect, %{err | connection_id: connection_id}, %{s | buffer: buffer}}
+        else
+          disconnect
+        end
     end
   end
 
--- a/lib/postgrex/protocol.ex
+++ b/lib/postgrex/protocol.ex
@@ -3068,6 +3068,12 @@ defmodule Postgrex.Protocol do
     end
   end
 
+  defp error_ready(s, _status, %Postgrex.Error{postgres: %{severity: severity}} = err, buffer)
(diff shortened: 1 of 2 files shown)


lib/postgrex/protocol.ex
```diff
@@ -3068,6 +3068,12 @@ defmodule Postgrex.Protocol do
     end
   end
 
+  defp error_ready(s, _status, %Postgrex.Error{postgres: %{severity: severity}} = err, buffer)
+       when severity in ["FATAL", "PANIC"] do
+    %{connection_id: connection_id} = s
+    {:disconnect, %{err | connection_id: connection_id}, %{s | buffer: buffer}}
+  end
+
   defp error_ready(s, status, %Postgrex.Error{} = err, buffer) do
```
v0idpwn: I don't love that this is called `error_ready` but couldn't come up with a better name for the new implementation.

v0idpwn: I think maybe instead of returning `{:disconnect, ...}` immediately, it should wait for ReadyForQuery **or** for the socket closed to be closed (which is what libpq does, if I'm not mistaken), then return the error.

v0idpwn (commented):
I think SimpleConnection may need some error handling tweak to handle this too. I'll look into that tomorrow 🙏🏻 

Edit: I think not.

test/query_test.exs
```diff
@@ -1940,6 +1940,26 @@ defmodule QueryTest do
...
+    %Postgrex.Result{connection_id: connection_id} = Postgrex.query!(pid, "SELECT 42", [])
+
+    # Start a long-running query in a separate process so we can terminate the
+    # backend while it's executing.
+    task =
+      Task.async(fn ->
+        Postgrex.query(pid, "SELECT pg_sleep(10)", [])
+      end)
+
+    Process.sleep(100)
```
josevalim: Instead of Process.sleep, you could setup a tracer and watch for Postgrex.query being returned. It should be deterministic. Would you like to give it a try?

v0idpwn: I pushed a commit using tracer to detect when the query was **sent** via tracing the `recv_bind`, but I'm still not really content with how it reads.

lib/postgrex/protocol.ex
```diff
@@ -3069,13 +3069,18 @@ defmodule Postgrex.Protocol do
...
-        %{connection_id: connection_id} = s
         {:error, %{err | connection_id: connection_id}, s}
 
       {:disconnect, _, _} = disconnect ->
-        disconnect
+        if err.postgres.severity in ["FATAL", "PANIC"] do
+          {:disconnect, %{err | connection_id: connection_id}, %{s | buffer: buffer}}
+        else
+          disconnect
+        end
```
v0idpwn, line 3083: My original implementation just skipped the `recv_ready`, but I think this would be more correct.

v0idpwn: Modified the title accordingly.

josevalim:

PR #770, opened 2026-06-17
Fix composite type loading when `record_send` is duplicated.

leandrocp:
The type-loading query does return qualified names when they are duplicated, for example when a user-defined `record_send` is present:

```
CREATE TYPE my_composite AS (a int, b text);
CREATE FUNCTION public.record_send(integer) RETURNS bytea LANGUAGE sql AS $$ SELECT ''::bytea $$;
Postgrex.query(conn, "SELECT $1::my_composite[]", [[]])
{:error, %Postgrex.QueryError{message: "type `_my_composite` can not be handled by the types module Postgrex.DefaultTypes"}}
```

This naive fix strips out the `pg_catalog.` prefix to keep only the bare name as expected by Postgrex.

--

For context, this was reported by a customer of Supabase Realtime and I tried the smallest fix possible to not cause too many changes and not cause any performance impact.

### DIFF
--- a/lib/postgrex/types.ex
+++ b/lib/postgrex/types.ex
@@ -134,7 +134,7 @@ defmodule Postgrex.Types do
     %TypeInfo{
       oid: oid,
       type: :binary.copy(type),
-      send: :binary.copy(send),
+      send: unqualify_proc(send),
       receive: :binary.copy(receive),
       output: :binary.copy(output),
       input: :binary.copy(input),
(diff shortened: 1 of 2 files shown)


josevalim:

greg-rychlewski:
This is a pretty cool issue. I’m going to show it to my team as an example of “whatever can happen will happen”

leandrocp:
> This is a pretty cool issue. I’m going to show it to my team as an example of “whatever can happen will happen”

I mean... it kinda makes sense, but duplicating a `record_send` function might not be very usual 😃

```
postgres> select t.oid, t.typsend from pg_type as t where t.typname = 'record';
+------+-------------+
| oid  | typsend     |
|------+-------------|
| 2249 | record_send |
+------+-------------+
SELECT 1

postgres> CREATE FUNCTION public.record_send(integer) RETURNS bytea LANGUAGE sql AS $$ SELECT ''::bytea $$;
CREATE FUNCTION

postgres> select t.oid, t.typsend from pg_type as t where t.typname = 'record';
+------+------------------------+
| oid  | typsend                |
|------+------------------------|
| 2249 | pg_catalog.record_send |
+------+------------------------+
SELECT 1
```

Issue #353, opened 2017-11-25
Composite types can not be safely encode or decoded


fishcakez:
If a composite type, such as a table row, changes its types then we do not update the types and oids. For encoding this means we send the wrong types to the database and error. For decoding it means we crash because oids don't match. Similarly for decoding anonymous composites it possible that it includes a type that we have not bootstrapped because the preparation of the query won't signal the missing type. In this situation we crash too. Catching any of these situations in a test is very unlikely.

I am unsure if this has been reported(https://github.com/elixir-ecto/ecto/issues/1271#issuecomment-343716476) . It would be very difficult to diagnose because of the complicated way postgrex handles types. Also I am unsure how much composites are used.

For encoding we can resolve this by looking up the composite types at parameter describe time - we don't support anonymous composites. If we require a redescribe (executing on new connection with known parameters types but unknown result types) and types changed then we will get a postgresql error because we don't lookup parameters on redescribe (1eb868). Doing a query on every composite prepare is undesirable but should work. We would be logging a warning every time this occurred. For performance people may choose not to use this but existing statements would still work.

For decoding we can partially resolve the internal types changing by treating all composites types as anonymous. This means we delay resolving the oids until decoding, or perhaps only if oids have changed fallback to the anonymous approach. However this means we need to be able to decode anonymous types.

Alternatively, or additional, we could lookup the oids at describe time (we describe results on explicit prepare and implicit redescribe). Unfortunately postgresql doesn't invalidated prepared queries when the sub-types change so we would either need to fallback to anonymous or crash if types change.

For decoding anonymous composites I am not sure. Technically if we find a type we don't know during decode we could look it up after results finish and then decode again. That'd be complex.

I feel this issue invalidates our general type handling approach somewhat and means edge cases will always exist. We may wish to consider having an additional type module that behaves like the majority of drivers in other languages: binary for built in types, text for other types. Of course we are making a trade off to get a richer type system so it might not be worthwhile changing existing behavior.

michalmuskala:
I wonder if it maybe would make sense to have some explicit API to tell postgrex that types have changed. I can see this being used when you migrate your database and then call some API in ecto to tell it about this. This could potentially give us some way of handling this - especially if we're saying we can never avoid all the edge cases.

josevalim:
@fishcakez is this now easier with DBConnection.EncodeError?

edgurgel:
[...]

PR #772, opened 2026-06-20
Fix domain over composite arrays

greg-rychlewski:
There was another gap in the way we were handling domains. This one is kind of weird but I'll do my best to paint a clear picture. This fix is for when the domain has a base type that is an array of a composite type.  For example this

```sql
CREATE TYPE composite_test AS (a text, b integer);
CREATE DOMAIN composite_array_domain AS composite_test[] CHECK (array_length(VALUE, 1) > 0);
```
When you try to encode a parameter for this you get the error "Postgrex has no clue how to encode this because we don't know the right extension" ([see here](https://github.com/elixir-ecto/postgrex/blob/39d8158060e6e34e705e37ab63f4df6a5ac5b79f/lib/postgrex/utils.ex#L167)). So why does this happen:

1. In the first bootstrap query we filter composites and arrays of composites out for efficiency ([see here](https://github.com/elixir-ecto/postgrex/blob/39d8158060e6e34e705e37ab63f4df6a5ac5b79f/lib/postgrex/types.ex#L55))
2. Our domain type does get bootstrapped as an array type (send = array_send) and its element type is the composite type that was not bootstrapped.
3. When we encode the parameter we go to fetch the type info and we see it's an array and then we say ok so what type is in the array and find a composite that has not been bootstrapped. So it errors and says Postgrex doesn't know what to do with this.

Then you might ask how are composites/arrays of composites working. And basically they get reloaded the first time they are found missing. So the fix here is to basically do the same thing: take domains with a base composite array type out of the original bootstrap so they can be reloaded.

The reason this isn't working right now is because it's only if the "top" type is missing that we reload. If a sub type is missing it just raises. And in our case we bootstrapped the top type (domain) but not the subtype (composite array)

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -3,7 +3,7 @@ name: CI
 on:
   push:
     branches:
-      - master
+      - fix_domain_2
   pull_request:
 
 jobs:
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -3,7 +3,7 @@ name: CI
 on:
   push:
     branches:
-      - fix_domain_2
+      - master
   pull_request:
 
 jobs:
--- a/lib/postgrex/type_info.ex
+++ b/lib/postgrex/type_info.ex
@@ -25,6 +25,7 @@ defmodule Postgrex.TypeInfo do
   @type t :: %__MODULE__{
           oid: Types.oid(),
           type: String.t(),
+          category: atom(),
           send: String.t(),
           receive: String.t(),
           output: String.t(),
@@ -34,5 +35,16 @@ defmodule Postgrex.TypeInfo do
           comp_elems: [Types.oid()]
         }
 
-  defstruct [:oid, :type, :send, :receive, :output, :input, :array_elem, :base_type, :comp_elems]
+  defstruct [
+    :oid,
+    :type,
+    :category,
+    :send,
+    :receive,
+    :output,
+    :input,
+    :array_elem,
+    :base_type,
+    :comp_elems
+  ]
 end
--- a/lib/postgrex/type_info.ex
+++ b/lib/postgrex/type_info.ex
@@ -15,7 +15,7 @@ defmodule Postgrex.TypeInfo do
     * `input` - The name of the "input" function (the function postgres uses
       to convert the type from its text format);
     * `array_elem` - If this is an array, the array elements' oid;
-    * `base_type` - If this is a range type, the base type's oid;
+    * `base_type` - If this is a range type, the base type's oid. If this is a domain, the underlying oid is pointed to;
     * `comp_elems` - If this is a composite type (record), the tuple
       elements' oid;
   """
--- a/lib/postgrex/types.ex
+++ b/lib/postgrex/types.ex
@@ -6,6 +6,16 @@ defmodule Postgrex.Types do
   alias Postgrex.TypeInfo
   import Postgrex.BinaryUtils
 
+  @type_categories %{
+    "b" => :base,
+    "c" => :composite,
+    "d" => :domain,
+    "e" => :enum,
+    "p" => :pseudo,
+    "r" => :range,
+    "m" => :multirange
+  }
+
   @typedoc """
   PostgreSQL internal identifier that maps to a type. See
   <https://www.postgresql.org/docs/9.4/static/datatype-oid.html>.
@@ -64,25 +74,18 @@ defmodule Postgrex.Types do
   end
 
   defp build_bootstrap_query(version, filter_oids, s) do
-    {typelem, join_domain} =
-      if version >= {9, 0, 0} do
-        {"coalesce(d.typelem, t.typelem)", "LEFT JOIN pg_type AS d ON t.typbasetype = d.oid"}
-      else
-        {"t.typelem", ""}
-      end
-
-    {rngsubtype, join_range} =
+    {base_oid, join_range} =
       cond do
         version >= {14, 0, 0} ->
-          {"coalesce(r.rngsubtype, 0)",
-           "LEFT JOIN pg_range AS r ON r.rngtypid = t.oid OR r.rngmultitypid = t.oid OR (t.typbasetype <> 0 AND r.rngtypid = t.typbasetype)"}
+          {"coalesce(r.rngsubtype, t.typbasetype)",
+           "LEFT JOIN pg_range AS r ON r.rngtypid = t.oid OR r.rngmultitypid = t.oid"}
 
         version >= {9, 2, 0} ->
-          {"coalesce(r.rngsubtype, 0)",
-           "LEFT JOIN pg_range AS r ON r.rngtypid = t.oid OR (t.typbasetype <> 0 AND r.rngtypid = t.typbasetype)"}
+          {"coalesce(r.rngsubtype, t.typbasetype)",
+           "LEFT JOIN pg_range AS r ON r.rngtypid = t.oid"}
 
         true ->
-          {"0", ""}
+          {"t.typbasetype", ""}
       end
 
     comp_oids =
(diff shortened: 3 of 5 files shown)


greg-rychlewski:
I think if we want to do this 100% properly we need to reload sub-types. But that is a pretty big change. We're only handling one level of domain nesting...i.e. you can define a new domain over an old domain. But that is consistent with how we were dealing with domain over arrays here for a long time:

```elixir
{typelem, join_domain} =
      if version >= {9, 0, 0} do
        {"coalesce(d.typelem, t.typelem)", "LEFT JOIN pg_type AS d ON t.typbasetype = d.oid"}
```

greg-rychlewski:
> I think if we want to do this 100% properly we need to reload sub-types.

Or instead of this if we had some kind of way to know if we are looking at a domain defined over a domain and trace the pointers to the end of the road.

josevalim:
Doesn't this tie back to the previous discussion that composite types are hard to support for several reasons, one because tables are composite types, and that would make bootstrap much more expensive, and another because they can be updated, which means future encodings can fail unless we start polling/watching for changes? (see #353)

greg-rychlewski:
Reading over that issue I think there are some overlapping ideas but maybe not exactly the same as this. He seems to be pointing out these two specific things:

- For named composites the underlying types can change
- For anonymous composites we might not know about a subtype and crash (similar to this issue)

The main difference I see is that anonymous composites basically can't fix the issue without trying to fix subtypes at query time because they are not in pg_type. But domains are in pg_type so you could fix it the same way we handle composite and arrays of composites. Which is basically ignore the "top" type at bootstrap time and take advantage of the existing reload functionality.

My last two messages also touch a different issue in the handling of domains. Even without composites our bootstrap assumes that when we have a domain that is an array of a normal type that we can find the array element type in one join. But if you have something like this that's true:

```sql
CREATE DOMAIN d1 AS int[];
CREATE DOMAIN d2 AS d1;
```

Basically you would nee two joins or you would crash. And this can go on for arbitrary levels. I haven't solved this issue in this PR but my gut feeling is there is something that can be done about it by treating domains as pointers to the base type.

greg-rychlewski:
> by treating domains as pointers to the base type

But actually if we did something like this then this PR wouldn't be necessary. Because it would point to the composite array which would be missing and then the top level reload would kick in

greg-rychlewski:
Would you mind if I try a PR where basically for domains we store the resolved extension to use in the types table? It might not be possible but my feeling is that there is a chance there is a nice way to do it. And if there is we don’t need this PR. It would also let us remove one of the joins from our existing bootstrap query trying to get the domain array element t

josevalim:
I don’t mind at all. The only concern is inflating the types in memory. I don’t fully remember all the details but there was something about tables also being types. We may need to make it opt-in.

greg-rychlewski:
@josevalim I actually discovered more than I anticipated trying to dig into this. But I think the result is pretty nice and actually let's us handle more than I originally thought.

There are basically two main changes:

1. When we encounter a domain type try to find the type that is eventually resolved to when bootstrapping. May be arbitrary levels of nesting
2. In the past when we bootstrapped the types we conflated two different things: (1) a necessary sub-type was not bootstrapped and (2) the type was bootstrapped but there are no extensions matching the type. 

For (2) I believe these should be separated because they are not the same thing. Missing types can be fixed by reloading and they can be fixed to arbitrary levels of nesting, i.e. a parent type and one or more of its sub types might be missing. 

For example something that already works before this change is missing arrays of composites. Neither the array type nor the underlying element type are bootstrapped. The code here essentially ensures they are bootstrapped from bottom up so that they can find the extension properly: https://github.com/elixir-ecto/postgrex/blob/c482554f25112a29cbc202bf7babc7bdecf199f2/lib/postgrex/protocol.ex#L2008

So basically all PR is doing is extending what we do for composite arrays to other things that have the same problem. This had a few other side effects:

- Now we can also handle things like ranges over composites. In the past the composite would not be bootstrapped and this would raise a "postgres can't handle this type" exception
- We could get rid of the hacks we were doing for domains in the bootstrap query. Essentially we merged composites and arrays into domains (up to one level of nesting). But now we can just handle them naturally.

The thing that is not perfect is that we essentially have to remove some types that we pulled from the bootstrap query. But I don't see any way around that because with arbitrary levels of nesting you can't really detect which ones to leave out in SQL without it being a massively complex query.

Let me know what you think.

josevalim:
@greg-rychlewski excellent work! The only remaining question is: does this affect how long the query takes and how much data we store? If it doesn't have a significant impact, we can definitely ship it. If you have an existing database you can connect to and measure, even better!!

greg-rychlewski:
Yeah definitely I can spin something up and do some tests on the old and the new.

I can give a few high level replies to your questions before as well:

1. You meant how long the bootstrap query takes right? If so then this change actually makes it simpler. We don't have to join pg_type onto itself anymore. In the past we did this to try to merge some of the underlying type information into the domain. But now we are treating domains as pointers and don't have to conflate fields
2. If you mean the data we store in the types table then we are not adding any new columns or changing anything about what was already stored. But we are storing more in the sense that now we handle arbitrary levels of nesting for domains. And to make the extension resolution faster we essentially copy the base type's extension into each of the individual domains  table record. 

For 2 there are a few things to think about:

1. Instead of copying the base type's extension into all of the domains in the chain we could essentially store pointers in them and use that to get to the base type's extension from the table. This would add more work to each query though.
2. Instead of making a chain of N domains someone could have made N domains of depth 1 all pointing to the same base type and we wouldn't be able to do anything about it now or with our old code
3. It's similar to how nesting already works in the types table. For example if you have an array over a composite over a composite the parent type would essentially store the whole extension chain. And each subtype would store its own part of the extension chain and lower. Rather than having them point to other places in the ets table.
4. I don't know for sure but I imagine the number of domains people are defining is not crazy. There is some evidence of that in the sense that we are not really handling it completely but no one has complained much over the years or raised issues about deeply nested domains.

greg-rychlewski:
I guess the TLDR of the above is there is probably only one serious decision to make. For domains do we store a pointer to the base type's table entry or store the full extension. And then the main considerations I can think of:

1. If we store a pointer then for domains we have 2 ets lookups for every query instead of 1
2. If we store the extension it is more copies of the same thing but it's similar to how nesting works today already and there might not be a tonne of domains anyway

josevalim:
:+1: for flattenning it all so the query lookup/building is faster. If the query is faster even better, let's just be double sure the flattening is not leading to much larger tables!

greg-rychlewski:
@josevalim I have done some testing and put a script here: https://github.com/greg-rychlewski/postgrex_type_size

Basically this is the test:

1. Control whether using Postgrex current master or the changes in this PR
2. Control whether we creating domains over composites types or not
3. Run the test on both branches without creating domains over composite types to see how things behave with only "normal" types 
4. Run the test on both machines creating > 100 different types of domains over composites with different levels of nested domains and arrays to see how they behave

The output of the tests are saved in different files:

1. Master branch with no domains
2. This branch with no domains
3. Master branch with domains
4. This branch with domains

The conclusions:

Without domains they are basically the same. Master has  11403words (memory) and 195 elements in the table. This branch has 11793 words (memory) and 195 elements in the table.

With domains they behave differently. Master tries to be smart doing single joins to try to find the domains during bootstrap. While this branch waits until it is used in a query to rely on reload.

So before using any queries this is how they compare:

Master:  23715 words (memory) and 355 entries
Fix: 14103 words (memory) and 195 entries

So as expected this fix is much lower because it gets the type the first time a query needs them.

Now after all the queries this is how it looks:

Fix:  26145 words (memory) and 355 entries
Master: 26626 words (memory) and 395 entries

This number of entries shot up quite a bit for this branch because of the lazy loading. It has 40 fewer entries because there are some combinations of nesting that do not work (in either this branch or master). Master counts them because it thinks it has it loaded properly but will actually crash when it tries to use it. 

So comparing equal numbers of entries it is 26145 words for this branch and 23715 for master. So something like 20kb difference I believe. But keep in mind master also cannot properly query most of these types. Any level of nesting of domains doesn't work in master. So even though the type is saved it can't be used.

I think this all seems reasonable but let me know what you think or if anything I did doesn't make sense. Sorry for the rambling.

josevalim:
Those results look great, worth giving it a try! Awesome job @greg-rychlewski!

josevalim (approved):

PR #773, opened 2026-06-22
Enforce integer bounds in type encoders

Snehil-Shah:
Started from `interval` and found quite a few types missing integer bound checks before stuffing the binary, silently wrapping the bits and corrupting data, without the user knowing.

I have added bound checks to the following types, such that it now rejects out-of-bound values like PG actually would:

- `inet`
- `interval`
- `macaddr`
- `numeric`
- `tid`
- `tsvector`

### DIFF
--- a/lib/postgrex/extensions/tsvector.ex
+++ b/lib/postgrex/extensions/tsvector.ex
@@ -4,6 +4,8 @@ defmodule Postgrex.Extensions.TSVector do
   import Postgrex.BinaryUtils, warn: false
   use Postgrex.BinaryExtension, send: "tsvectorsend"
 
+  @max_position 16383
+
   def encode(_) do
     quote location: :keep do
       values when is_list(values) ->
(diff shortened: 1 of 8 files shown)


lib/postgrex/extensions/tsvector.ex
```diff
@@ -36,6 +38,14 @@ defmodule Postgrex.Extensions.TSVector do
...
     positions =
       Enum.map(positions, fn {position, weight} ->
+        # Lexeme positions can range from 1 to 16383 with a silent ceiling at the top.
+        # Ref: https://www.postgresql.org/docs/current/datatype-textsearch.html#DATATYPE-TSVECTOR
+        if position < 1 do
+          raise DBConnection.EncodeError,
+                Postgrex.Utils.encode_msg(position, 1..@max_position)
+        end
+
+        position = min(position, @max_position)
```
Snehil-Shah, line 48: According to [postgres docs](https://www.postgresql.org/docs/current/datatype-textsearch.html#DATATYPE-TSVECTOR):

> "Position values can range from 1 to 16383; larger numbers are silently set to 16383"

Decided to follow their behavior, although I do feel that explicitly raising for max position can be a more consistent experience. WDYT?

josevalim:

PR #775, opened 2026-07-01
use private functions as report_cb

martosaur:
@hauleth [raised a good point](https://github.com/elixir-ecto/postgrex/pull/768#issuecomment-4847758237) that anonymous functions shouldn't be used as report callbacks. This PR uses private functions instead. Embarrassingly, this [isn't the first time](https://github.com/elixir-lang/elixir/pull/14380#issuecomment-2764222374) I'm making this mistake.

Anyways, I went with a private function per log message, but open to other arrangements (separate module? a single function with multiple heads matching on `log_id`?) if you think it's better.

### DIFF
--- a/lib/postgrex/simple_connection.ex
+++ b/lib/postgrex/simple_connection.ex
@@ -371,14 +371,7 @@ defmodule Postgrex.SimpleConnection do
             mod: mod,
             reason: reason
           },
-          report_cb: fn report ->
-            {"~ts (~ts) failed to connect to Postgres: ~ts",
-             [
-               inspect(report.pid_or_name),
-               inspect(report.mod),
-               Exception.format(:error, report.reason)
-             ]}
-          end
+          report_cb: &__MODULE__._format_connection_error/1
         )
 
         if state.auto_reconnect do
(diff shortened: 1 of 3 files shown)


josevalim:
@hauleth does it need to be a public function too? IIRC yes but double checking!

hauleth:
Yeah, it needs to be public function and capture needs to use remote call format.

Personally I use such approach most of the time:

```elixir
# Do not show in documentation
@doc false
# underscore functions aren't imported by default AFAIK
def __report_cb__(report) do
  # ...
end
```

That way it is public function from VM viewpoint, but it is private API from documentation viewpoint.

martosaur:
I'll make the change, but at this point I have to ask for more details. What is that quirk we're talking about and where can I read about it? It sounds like it'd be nice to document it somewhere if reports are to get any adoption.

martosaur:
Ok I think I found where it's coming from. Telemetry [recommends](https://telemetry.hexdocs.pm/telemetry.html#attach/4) remote captures for performance reasons, but looking at (this thread)[https://github.com/elixir-lang/elixir/pull/13668] it looks like the eventual conclusion was that performance difference might not be noticeable with recent OTP versions. So the only justification that landed in the docs is hot code reload. Which is still a good enough reason to use remote captures!

martosaur:
Pushed the change! I didn't spot any difference from underscores (as tested with `ModuleName.module_info(:exports)`) so didn't use them.

hauleth:
https://elixir.hexdocs.pm/Kernel.SpecialForms.html#import/2-underscore-functions

martosaur:
oof i misread "import" for "export" 🤦 ok, thanks for bearing with me folks, this new push should finally be good!

lib/postgrex/simple_connection.ex
```diff
@@ -440,6 +433,15 @@ defmodule Postgrex.SimpleConnection do
   end
 
   ## Helpers
+  @doc false
```
josevalim, line 437:
```suggestion

  @doc false
```

josevalim:

PR #776, opened 2026-07-02
Add SCRAM-SHA-256-PLUS support

Snehil-Shah:
This adds channel binding support for SCRAM auth handshake, enabling better security with servers that support `SCRAM-SHA-256-PLUS`.

Also adds a `channel_binding` knob with three modes: `:prefer`, `:require`, and `:disable`. (all self-explanatory)

I tested the flow locally on PG 16 with SSL enabled and SCRAM auth (so server advertises `SCRAM-SHA-256-PLUS`) in all three of the above modes, works as expected.

Proposal: https://groups.google.com/g/elixir-ecto/c/Flxp0Krz6hY

### DIFF
--- a/lib/postgrex.ex
+++ b/lib/postgrex.ex
@@ -124,6 +124,12 @@ defmodule Postgrex do
        - true: enable SSL with secure defaults, including peer certificate verification and hostname checking.
        - keyword list of `t::ssl.tls_client_option()/0` values: enable SSL and merge your options on top of secure defaults.
 
+    * `:channel_binding` - Controls the use of SCRAM channel binding (`SCRAM-SHA-256-PLUS`).
+      Only applies to SSL connections against servers using SCRAM authentication. Set to:
+       - `:prefer` (default): use channel binding when the server supports it.
+       - `:require`: refuse to connect unless channel binding is used.
+       - `:disable`: never use channel binding.
+
     * `:socket_options` - Options to be given to the underlying socket
       (applies to both TCP and UNIX sockets);
 
--- a/lib/postgrex.ex
+++ b/lib/postgrex.ex
@@ -48,6 +48,7 @@ defmodule Postgrex do
           | {:handshake_timeout, timeout}
           | {:ping_timeout, timeout}
           | {:ssl, boolean | [:ssl.tls_client_option()]}
+          | {:channel_binding, :prefer | :require | :disable}
           | {:socket_options, [:gen_tcp.connect_option()]}
           | {:prepare, :named | :unnamed}
           | {:transactions, :strict | :naive}
--- a/lib/postgrex/scram.ex
+++ b/lib/postgrex/scram.ex
@@ -6,15 +6,56 @@ defmodule Postgrex.SCRAM do
   @hash_length 32
   @nonce_length 24
   @nonce_rand_bytes div(@nonce_length * 6, 8)
-  @nonce_prefix "n,,n=,r="
-  @nonce_encoded_size <<byte_size(@nonce_prefix) + @nonce_length::signed-size(32)>>
+  @cb_gs2_header "p=tls-server-end-point,,"
 
-  def client_first do
-    nonce = @nonce_rand_bytes |> :crypto.strong_rand_bytes() |> Base.encode64()
-    ["SCRAM-SHA-256", 0, @nonce_encoded_size, @nonce_prefix, nonce]
+  # Chooses the SCRAM mechanism and gs2 header from the mechanisms the server offered.
+  def negotiate_channel_binding(server_mechanisms, sock, mode) do
+    offered = parse_mechanisms(server_mechanisms)
+    plus_offered? = "SCRAM-SHA-256-PLUS" in offered
+    tls? = match?({:ssl, _}, sock)
+
+    cond do
+      mode == :disable ->
+        {:ok, {"SCRAM-SHA-256", "n,,", ""}}
+
+      tls? and plus_offered? ->
+        case cert_hash(sock) do
+          {:ok, hash} ->
+            {:ok, {"SCRAM-SHA-256-PLUS", @cb_gs2_header, hash}}
+
+          :error when mode == :require ->
+            {:error,
+             %Postgrex.Error{
+               message:
+                 "channel binding is required but the server certificate has no usable hash"
+             }}
+
+          # Fallback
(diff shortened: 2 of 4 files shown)


lib/postgrex/scram.ex
```diff
@@ -6,15 +6,56 @@ defmodule Postgrex.SCRAM do
...
   @nonce_length 24
   @nonce_rand_bytes div(@nonce_length * 6, 8)
-  @nonce_prefix "n,,n=,r="
-  @nonce_encoded_size <<byte_size(@nonce_prefix) + @nonce_length::signed-size(32)>>
+  @cb_gs2_header "p=tls-server-end-point,,"
+
+  # Chooses the SCRAM mechanism and gs2 header from the mechanisms the server offered.
+  def negotiate_channel_binding(server_mechanisms, sock, mode) do
+    offered = parse_mechanisms(server_mechanisms)
+    plus_offered? = "SCRAM-SHA-256-PLUS" in offered
```
josevalim, line 14: Do we need to worry about `SCRAM-SHA-256` not being offered? We fallback to it in a few places.

lib/postgrex.ex
```diff
@@ -124,6 +124,12 @@ defmodule Postgrex do
        - true: enable SSL with secure defaults, including peer certificate verification and hostname checking.
        - keyword list of `t::ssl.tls_client_option()/0` values: enable SSL and merge your options on top of secure defaults.
 
+    * `:channel_binding` - Controls the use of SCRAM channel binding (`SCRAM-SHA-256-PLUS`).
```
josevalim, line 128: Please also update `@type start_option`!

test/login_test.exs
```diff
@@ -88,6 +88,51 @@ defmodule LoginTest do
...
+
+      error_log =
+        capture_log(fn ->
+          Postgrex.start_link(opts ++ context[:options])
+          assert_receive {:EXIT, _, :killed}
+        end)
+
+      assert error_log =~ "expected :channel_binding to be :prefer, :require or :disable"
+    end
+  end
```
josevalim, line 161: How easy it is to add a test for channel_binding: :require over non-SSL?

Snehil-Shah: Easy and obvious, I'll add it right away!

lib/postgrex.ex
Snehil-Shah: Ah yes, sorry, missed it.

test/login_test.exs
Snehil-Shah: @josevalim Added that test and also added a test for `:prefer`. Although, the `:disable` and `:prefer` tests don't really test the actual negotiation logic. It's just a sanity test. Can be removed, or we can work on finding a way to test the actual negotiation logic.. WDYT?

lib/postgrex/scram.ex
Snehil-Shah: @josevalim Tbh, we were anyway sticking to `SCRAM-SHA-256` before this and ignoring all offerings. I think it is a good enough incremental change where we are using `PLUS` when offered, and falling back otherwise to the original behavior. Less blast radius with existing behavior whilst supporting channel binding where offered. WDYT?

josevalim: Yeah, it is fine to not rock the boat too much, I am just worried in such cases the error message quality won't be great, but we can improve it later.

josevalim (approved):

lib/postgrex/scram.ex
```diff
@@ -6,15 +6,56 @@ defmodule Postgrex.SCRAM do
...
+          :error when mode == :require ->
+            {:error,
+             %Postgrex.Error{
+               message:
+                 "channel binding is required but the server certificate has no usable hash"
+             }}
+
+          # Fallback
+          :error ->
+            {:ok, {"SCRAM-SHA-256", "y,,", ""}}
```
greg-rychlewski: Should we send `n,,` when the mode is `:prefer`? Because I believe sending `y,,` will cause Postgres to reject the authentication

Snehil-Shah: @greg-rychlewski Thank you for catching this. `n,,` is the correct header here, I have pushed the change.

Thinking out loud, does it also hurt sending `n,,` instead of `y,,` when the server didn't offer `-PLUS`? Is there some backward compatibility benefit to this, given we used to always send `n,,` before? `y,,` is more semantically correct and lets the server know what's up, and I tested it on PG 10 that supports SCRAM but no channel binding, and it accepts `y,,` with no issues. Any case where this might not be backward-compatible?

greg-rychlewski: I think `y,,` is correct because it protects from downgrade attacks. In terms of backwards compatibility I'm not super familiar with this protocol if anything supporting SCRAM is supposed to be accepting `y,,`. If not this could be an issue for other databases that implement the same wire protocol like Yugabyte and Cockroach. 

If this is an issue I can think of a few things:

1. The default shouldn't be prefer it should be disable. But this is a pretty weak default.
2. We make prefer weaker so that it doesn't protect against downgrade attacks. This seems bad
3. We default prefer and let some things fail and then tell them to switch the mode if their database doesn't work with it

So all these seem kind of bad I'm hoping everything just needs to parse `y,,` out of the box. But we'd have to check. It would be good to see what @josevalim thinks too

jaqreails-tech (approved):

v0idpwn (approved):

lib/postgrex/scram.ex
Snehil-Shah: We are probably overthinking this. `y,,` is part of the standard [SCRAM grammar](https://www.rfc-editor.org/info/rfc5802/#section-7), so should be a safe (& obvious) assumption that servers would do it right. For confidence, libpq has the same semantics in their SCRAM logic, and I think it can be a good standard to follow.

greg-rychlewski: I am good with that! @josevalim im approving but wont merge just to be sure you saw this convo and are ok with the conclusion as well

greg-rychlewski (approved):

lib/postgrex/scram.ex
greg-rychlewski: @Snehil-Shah thanks for the contribution btw. This is a nice feature

josevalim:

PR #777, opened 2026-07-03
fix incorrect match in report_cb callback

martosaur:
The changes from `source` to `log_id` in https://github.com/elixir-ecto/postgrex/pull/768 left an invalid match in the code, apologies. I also made it just `inspect(__MODULE__)` so that log message is fully identical to the one in the current released version.


josevalim:

PR #778, opened 2026-07-05
Fix encoding/decoding of arrays of domains defined over arrays

greg-rychlewski:
Sorry for throwing another weird one your way about domains. This comment here basically explains it

```elixir
# A nested array is normally handled by recursing through the dimensions
  # until you hit the base element. However, there is a special case
  # where a parameter may look like a nested array but we cannot treat it
  # this way. An array whose element type is a domain that is defined
  # over an array will look like a nested array from Elixir, but the
  # Postgres protocol must treat it as a single dimensional array over
  # the domain type
```

### DIFF
--- a/lib/postgrex/extensions/array.ex
+++ b/lib/postgrex/extensions/array.ex
@@ -17,9 +17,9 @@ defmodule Postgrex.Extensions.Array do
   def encode(_) do
     quote location: :keep do
       list, [oid], [type] when is_list(list) ->
+        recurse_nested? = unquote(__MODULE__).recurse_nested_arrays(type)
         # encode_list/2 defined by TypeModule
-        encoder = &encode_list(&1, type)
-        unquote(__MODULE__).encode(list, oid, encoder)
+        unquote(__MODULE__).encode(list, recurse_nested?, oid, &encode_list(&1, type))
 
       other, _, _ ->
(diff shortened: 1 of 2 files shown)


lib/postgrex/extensions/array.ex
```diff
@@ -110,6 +110,16 @@ defmodule Postgrex.Extensions.Array do
...
 
+  # A nested array is normally handled by recursing through the dimensions
+  # until you hit the base element. However, there is a special case
+  # where a parameter may look like a nested array but we cannot treat it
+  # this way. An array whose element type is a domain that is defined
+  # over an array will look like a nested array from Elixir, but the
+  # Postgres protocol must treat it as a single dimensional array over
+  # the domain type
+  def recurse_nested_arrays({__MODULE__, _, _}), do: false
+  def recurse_nested_arrays(_), do: true
```
greg-rychlewski, line 121: The reason why this check works is because no matter how many dimensions the array is over a non-domain type, the sub type is still considered the base type. For example [[[[[int]]]]] will still just have a sub type of int. This is enforced by Postgres and not just a quirk of how we handle types. 

But [domain_over_array] will have a sub type of array. This is also enforced by postgres or else the protocol rejects it. i.e. postgres doesn't see base_type[][] the same as domain_over_array_of_base_type[].

josevalim (approved):

PR #779, opened 2026-07-07
Clarify that Duration keeps PostgreSQL's coarse interval units

nseaSeb:
Decoding an interval into a Duration stores the time part in :second without splitting it into :hour/:minute, mirroring Postgrex.Interval. This surprises users who expect 1 day 02:03:04 to decode as %Duration{day: 1, hour: 2, minute: 3, second: 4} rather than %Duration{day: 1, second: 7384} (both are the same value). Document this in Postgrex.Types.define/3 and the README, and fix the README Duration example so it matches the Postgrex.Interval row above it.


greg-rychlewski (approved):

greg-rychlewski:
Thanks :)

PR #781, opened 2026-07-10
Demonitor on Parameters delete to stop monitor leak

### LINKED ISSUE #780: Postgrex.Parameters leaks a monitor on every reconnect (no demonitor on delete)
### Elixir version

Elixir 1.18.4 (compiled with Erlang/OTP 27)

### Database and Version

PostgreSQL 16.9

### Postgrex Version

0.22.3

### Current behavior

`Postgrex.Parameters` monitors the connection process on each handshake but never demonitors on delete. Since a pooled connection reconnects in place (same pid, never dies), no `:DOWN` fires, so every reconnect orphans a monitor. On a long-running node the process accumulates monitors without bound (observed ~1M monitors across 51 live connection pids, ~70 MB, unreclaimable by GC).

`lib/postgrex/parameters.ex`:

```elixir
def handle_call({:insert, parameters}, {pid, _}, state) do
  ref = Process.monitor(pid)          # new monitor every handshake
  true = :ets.insert_new(state, {ref, parameters})
  {:reply, ref, state}
end

def handle_cast({:delete, ref}, state) do
  :ets.delete(state, ref)             # monitor never released
  {:noreply, state}
end
```

### Expected behavior

Deleting a parameters entry should release its monitor, so monitor count stays bounded by the number of live connections regardless of reconnect frequency:

```elixir
def handle_cast({:delete, ref}, state) do
  Process.demonitor(ref, [:flush])
  :ets.delete(state, ref)
  {:noreply, state}
end
```

This matches the demonitor idiom already used in `type_server.ex`, `scram/locked_cache.ex`, and `notifications.ex`; `parameters.ex` is the only module that monitors without demonitoring.

ribanez7:
Fixes #780.

`Postgrex.Parameters` monitors on insert but never demonitors on delete, so every in-place reconnect leaks a monitor. This adds the `Process.demonitor(ref, [:flush])` already used in the other monitoring modules.

### DIFF
--- a/lib/postgrex/parameters.ex
+++ b/lib/postgrex/parameters.ex
@@ -60,6 +60,7 @@ defmodule Postgrex.Parameters do
   end
 
   def handle_cast({:delete, ref}, state) do
+    Process.demonitor(ref, [:flush])
     :ets.delete(state, ref)
     {:noreply, state}
   end


josevalim:

PR #783, opened 2026-07-14
Reject auth downgrade when channel binding required

Snehil-Shah:
Following up on #776. I think we shouldn't allow downgrading to cleartext/md5 if we have channel_binding as `:require`. This aligns with libpq. I missed this the last time.


josevalim:

PR #784, opened 2026-07-22
fix: query typespec + docs on comment

edgurgel:
dialyxir was not happy about some of these options 😄


josevalim:

PR #785, opened 2026-07-24
Do not UNLISTEN while disconnected

Snehil-Shah:
We send the `UNLISTEN` query unconditionally when removing the last subscriber, which can crash the process if there is no DB connection to receive it. Similar to how we have a safety mechanism for queuing `LISTEN` queries, I think we can have a simple guard that ignores `UNLISTEN` if the connection doesn't exist anymore.


josevalim:

PR #790, opened 2026-07-31
Fix async messages handling during replication startup

### LINKED ISSUE #789: `recv_streaming/2` crashes on asynchronous messages during replication startup
### Elixir version

any

### Database and Version

any

### Postgrex Version

0.22.2

### Current behavior

`Postgrex.Protocol.recv_streaming/2` crashes if PostgreSQL sends an asynchronous protocol message after `START_REPLICATION` but before `CopyBothResponse` or `CopyOutResponse`.

PostgreSQL may send `ParameterStatus`, `NoticeResponse`, or `NotificationResponse` asynchronously. If one arrives during this window, `recv_streaming/2` raises a `CaseClauseError`, terminating the replication connection. With `auto_reconnect: true`, this may cause a reconnect/crash loop.

Repro:
this synthetic message ordering reproduces the crash
```elixir
responses =
  IO.iodata_to_binary([
    backend_message(?S, ["application_name", 0, "postgrex", 0]),
    backend_message(?N, [?S, "NOTICE", 0, ?M, "replication starting", 0, 0]),
    backend_message(?A, [<<123::32>>, "events", 0, "ready", 0]),
    backend_message(?W, <<0, 0::16>>)
  ])

state = %Postgrex.Protocol{
  sock: {TestSocket, self()},
  buffer: responses,
  parameters: %{},
  messages: []
}

Postgrex.Protocol.handle_streaming("START_REPLICATION", state)
```

Result:
```
** (CaseClauseError) no case clause matching:

    {:ok, {:msg_parameter, "application_name", "postgrex"}, ...}

    lib/postgrex/protocol.ex: Postgrex.Protocol.recv_streaming/2
```

The bug is likely difficult to reproduce with real postgres as it depends on specific message order. An example scenario where it may happen would need:

1. PostgreSQL receives `START_REPLICATION`.
2. It creates the logical decoding context before sending `CopyBothResponse`.
3. Creating that context invokes the output plugin’s `startup_cb`.
4. If that callback emits `NOTICE` or `WARNING`, PostgreSQL sends a real `NoticeResponse`.
5. Postgrex crashes `CaseClauseError`.

PostgreSQL documents that clients must accept asynchronous messages during protocol processing. `NoticeResponse` is explicitly possible because of external activity, and `ParameterStatus` can result from a configuration reload. See https://www.postgresql.org/docs/17/protocol-flow.html#PROTOCOL-ASYNC

### Expected behavior

Asynchronous messages should be processed normally, and the connection should continue waiting for `CopyBothResponse`, `CopyOutResponse`, or an error

lukaszsamson:
Fixes #789

### DIFF
--- a/lib/postgrex/protocol.ex
+++ b/lib/postgrex/protocol.ex
@@ -1363,6 +1363,11 @@ defmodule Postgrex.Protocol do
         err = Postgrex.Error.exception(postgres: fields)
         error_ready(s, status, err, buffer)
 
+      {:ok, msg, buffer} ->
+        status = new_status([], mode: :transaction)
+        s = handle_msg(s, status, msg)
+        recv_streaming(s, buffer)
+
       {:disconnect, _, _} = dis ->
         dis
     end
--- a/test/protocol_test.exs
+++ b/test/protocol_test.exs
@@ -1,0 +1,40 @@
+defmodule Postgrex.ProtocolTest do
+  use ExUnit.Case, async: true
+
+  alias Postgrex.Protocol
+
+  defmodule Socket do
+    def send(pid, data) do
+      Kernel.send(pid, {:sent, IO.iodata_to_binary(data)})
+      :ok
+    end
+  end
+
+  test "streaming startup handles asynchronous messages before the copy response" do
+    responses =
+      IO.iodata_to_binary([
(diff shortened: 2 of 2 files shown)


josevalim:

PR #791, opened 2026-08-05
Decode void type on CockroachDB

milmazz:
CockroachDB names the void type's send/output procs without underscores (`voidsend` / `voidout`), so the built-in `VoidBinary`/`VoidText` extensions never bind and `SELECT`s returning void (e.g. pg_advisory_xact_lock/1) raise `type "void" can not be handled`.

Match CRDB's proc spelling in addition to Postgres' so void decodes on both engines. The wire format is identical; only the matching changes, and the Postgres path is unaffected.


josevalim:

PR #793, opened 2026-09-03
Reject timestamptz values that do not fit the 64-bit wire format

dylanpulver:
`TimestampTZ.encode_elixir/1` writes the microsecond value into an `int64`
field without checking that it fits. Erlang truncates silently, so PostgreSQL
receives a *different* timestamp and stores that instead.

#689 removed the year cap on the grounds that "Postgres will return an error if
the timestamp is out of range". That holds for values that survive the cast, but
not for ones that wrap back into the accepted range.

Measured against PostgreSQL 17.9, inserting into a `timestamptz` column:

| input DateTime | stored |
|---|---|
| `294277-01-09 04:00:54.775807Z` | **`infinity`** (`isfinite` = false) |
| `585706-01-01` | `1151-12-14 ...` |
| `587483-01-01` | `2928-12-13 ...` |
| `294277-01-01`, `300000`, `1000000` | correctly rejected |

The first is exact rather than lucky: it encodes to 9223372036854775807, which
is the `infinity` sentinel. Sampling years 294278..2000000, 477 of 960 land back
inside the accepted range.

Reachable from ordinary arithmetic, e.g. `DateTime.add(dt, n, :microsecond)`
with a large `n`.

The sibling `timestamp` extension already bounds its input and rejects the same
year. #689 only added a test for the negative bound, since at the time the upper
bound "will crash on most elixir versions" — on Elixir 1.20 it no longer does.

Reverting the guard fails the new test. Guarding only the two sentinels also
fails it, because the wrap cases still pass through.

Assisted by Claude (`claude-opus-5`); I reviewed and verified the change.

### DIFF
--- a/test/calendar_test.exs
+++ b/test/calendar_test.exs
@@ -306,13 +306,14 @@ defmodule CalendarTest do
       # the value has to be rejected here instead.
       sentinel =
         DateTime.from_naive!(NaiveDateTime.new!(294_277, 1, 9, 4, 0, 54, {775_807, 6}), "Etc/UTC")
-  
-      wraps_into_range = DateTime.from_naive!(NaiveDateTime.new!(585_706, 1, 1, 0, 0, 0), "Etc/UTC")
-  
+
+      wraps_into_range =
+        DateTime.from_naive!(NaiveDateTime.new!(585_706, 1, 1, 0, 0, 0), "Etc/UTC")
+
       assert_raise ArgumentError, ~r/beyond the range/, fn ->
         query("SELECT $1::timestamptz", [sentinel])
       end
-  
+
       assert_raise ArgumentError, ~r/beyond the range/, fn ->
         query("SELECT $1::timestamptz", [wraps_into_range])
       end
--- a/test/calendar_test.exs
+++ b/test/calendar_test.exs
@@ -299,21 +299,23 @@ defmodule CalendarTest do
     assert message =~ "timestamp out of range"
   end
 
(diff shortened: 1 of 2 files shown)


test/calendar_test.exs
```diff
@@ -299,6 +299,26 @@ defmodule CalendarTest do
...
+      sentinel =
+        DateTime.from_naive!(NaiveDateTime.new!(294_277, 1, 9, 4, 0, 54, {775_807, 6}), "Etc/UTC")
+  
+      wraps_into_range = DateTime.from_naive!(NaiveDateTime.new!(585_706, 1, 1, 0, 0, 0), "Etc/UTC")
+  
+      assert_raise ArgumentError, ~r/beyond the range/, fn ->
+        query("SELECT $1::timestamptz", [sentinel])
+      end
+  
+      assert_raise ArgumentError, ~r/beyond the range/, fn ->
```
josevalim, line 317:
```suggestion
      sentinel =
        DateTime.from_naive!(NaiveDateTime.new!(294_277, 1, 9, 4, 0, 54, {775_807, 6}), "Etc/UTC")

      wraps_into_range =
        DateTime.from_naive!(NaiveDateTime.new!(585_706, 1, 1, 0, 0, 0), "Etc/UTC")

      assert_raise ArgumentError, ~r/beyond the range/, fn ->
        query("SELECT $1::timestamptz", [sentinel])
      end

      assert_raise ArgumentError, ~r/beyond the range/, fn ->
```

josevalim:

PR #795, opened 2026-09-06
Fix notifications lost during idle ping

### LINKED ISSUE #794: Notifications silently lost during idle connection ping
### Elixir version

any

### Database and Version

any

### Postgrex Version

0.22.2

### Current behavior

`Postgrex.SimpleConnection` can silently discard `NotificationResponse` messages while an idle ping is running. The connection remains healthy, no error is reported, and the lost notifications are never delivered.

Repro:
1. Start a `Postgrex.Notifications` process listening on a channel.
2. Configure it with a very short `:idle_interval`, such as 5 ms.
3. From another connection, issue 400 `pg_notify` calls approximately 3 ms apart.
4. Count the notifications received by the subscriber.
Observed results:
```
idle_interval: 60_000  -> sent=400 received=400 missing=0
idle_interval: 5              -> sent=400 received=378 missing=22
```
The missing sequence numbers are scattered throughout the stream, consistent with idle pings firing between notification arrivals.

Reason:
`Protocol.ping/1` calls `new_status([], mode: :transaction)`, not passing the notification callback

### Expected behavior

No notifications lost on healthy connection

lukaszsamson:
This PR adds public `Protocol.ping/2` and passes `notify` callback option to `Protocol.new_status/2` when handling idle `:timeout` event. 

Note: this new `Protocol.ping/2` is not a callback of `DbConnection` behaviour. Downside is `ping/1` called directly still defaults to `[]` options. Alternatives:
- extend the `DbConnection` callback surface
- keep using `ping/1` but instead pass `notify` as state struct property

I'm happy to make the changes

Fixes #794

### DIFF
--- a/lib/postgrex/protocol.ex
+++ b/lib/postgrex/protocol.ex
@@ -313,12 +313,17 @@ defmodule Postgrex.Protocol do
   @spec ping(state) ::
           {:ok, state}
           | {:disconnect, Postgrex.Error.t() | %DBConnection.ConnectionError{}, state}
-  def ping(%{postgres: :transaction, transactions: :strict} = s) do
+  def ping(s), do: ping(s, [])
+
+  @spec ping(state, Keyword.t()) ::
+          {:ok, state}
+          | {:disconnect, Postgrex.Error.t() | %DBConnection.ConnectionError{}, state}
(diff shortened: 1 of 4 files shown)


josevalim:

PR #796, opened 2026-09-09
Rearm SimpleConnection after query errors

lukaszsamson:
After a successful simple query, the implementation invokes `Protocol.checkin/1`, which switches the socket back into active mode. The `{:error, %Postgrex.Error{}, protocol}` branch invoked `handle_result/2` directly leaving socket in passive state. Impact: a recoverable SQL error stopped asynchronous notifications delivery on a healthy connection. It would self heal on the next successful query
This PR tries to checkin the connection on error, and if that fails disconnects.


josevalim:

PR #798, opened 2026-09-11
Ignore idle timeout while disconnected

### LINKED ISSUE #797: Notifications crash during reconnect
### Elixir version

any

### Database and Version

any

### Postgrex Version

master

### Current behavior

A failed connection attempt schedules reconnect backoff with `protocol: nil`. A callback returning `{:noreply, state}` schedules an idle timeout which calls `Protocol.ping(nil)`

Repro:
start `Postgrex.Notifications` on loopback port with:

```elixir
sync_connect: false,
auto_reconnect: true,
reconnect_backoff: 5000,
idle_interval: 20
```

`Postgrex.Notifications.listen(pid, "some_channel")` correctly returns `{:eventually, ref}` while disconnected. Around 20 ms later, the process crashes with:

```
FunctionClauseError: no function clause matching in Postgrex.Protocol.ping/1
Postgrex.Protocol.ping(nil)
Postgrex.SimpleConnection.handle_event/4, simple_connection.ex:401
```

Impact: during an outage, a `listen` operation kills the process instead of keeping the subscription alive for reconnection. It causes unnecessary process churn and loses subscription references

### Expected behavior

No crash

lukaszsamson:
This PR fixes the crash by ignoring idle timeouts while the connection is down
Fixes #797

### DIFF
--- a/lib/postgrex/simple_connection.ex
+++ b/lib/postgrex/simple_connection.ex
@@ -397,6 +397,10 @@ defmodule Postgrex.SimpleConnection do
     handle(mod, :handle_call, [msg, callback_from, mod_state], from, state)
   end
 
+  def handle_event(:timeout, nil, @state, %{protocol: nil} = state) do
+    {:keep_state, state}
+  end
+
   def handle_event(
         :timeout,
         nil,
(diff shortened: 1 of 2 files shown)


josevalim:

PR #799, opened 2026-09-18
Add pause and resume for replication streams

olivermt:
Hi!

Discovered a bug in Replicant that required an addition to Postgrex to handle.
Basically a bulk import flooded WAL faster than CDC could process it. Postgrex kept reading with no pause mechanism, so Replicant’s buffer limit was exceeded and replication stalled. Replicant could detect overload but couldn’t stop Postgrex from reading more WAL. The patch adds pause/resume so CDC can drain its queue without disconnecting or losing replication-slot ownership.

## Summary of code changes

Adds backpressure support to `Postgrex.ReplicationConnection`.

`handle_data/2` can return `{:pause, state}` or `{:pause, replies, state}` to stop socket delivery without disconnecting. `handle_info/2` and `handle_call/3` can resume delivery with equivalent `{:resume, ...}` responses.

Buffered messages retain their order, calls remain responsive, and replication feedback can still be sent while paused. Reconnecting clears buffered messages from the previous connection.

Tests cover TCP/TLS ordering, slot ownership, feedback while paused, and reconnect cleanup.


lib/postgrex/protocol.ex
```diff
@@ -1288,57 +1288,65 @@ defmodule Postgrex.Protocol do
     msg_send(s, msgs, buffer)
   end
 
-  @spec handle_copy_recv(any, Keyword.t(), state) ::
+  @spec handle_copy_recv(any, pos_integer() | nil, state) ::
```
olivermt, line 1291: this Keyword.t() seems like an obvious mis-typing as it receives max_copies? Or am I misunderstanding something here

lib/postgrex/protocol.ex
```diff
@@ -1288,57 +1288,65 @@ defmodule Postgrex.Protocol do
...
           {:ok, [binary | atom], state}
           | :unknown
           | {:error, Postgrex.Error.t(), state}
           | {:disconnect, %DBConnection.ConnectionError{}, state}
   def handle_copy_recv(msg, max_copies, s) do
+    handle_copy_recv(msg, max_copies, s, true)
+  end
+
+  @doc false
```
josevalim: It seems we don't tag anything as doc false. The whole module is private anyway.

```suggestion
```

lib/postgrex/replication_connection.ex
```diff
@@ -578,6 +616,19 @@ defmodule Postgrex.ReplicationConnection do
      ]}
   end
 
+  defp handle_data(copies, %{paused: true} = s), do: {:keep_state, %{s | pending: copies}}
```
josevalim, line 619: Shouldn't we be concatenating this? or is it guarantee this is called just once?

josevalim:
Thank you! I understand the issue and the fix, but I am not sure if this is the best solution. We are now introducing a queue and the fact that we need to respond to pings and timeouts still has complexity to the "pause" perception.

Perhaps a simpler alternative is that pause only pauses the _next_ message but any other callback or message returning something other than pause kicks it back into life. But I am not sure if that's better.

olivermt:
That works as well for  my case I believe.

If the context helps, here is how I fixed it on the Replicant end of things https://github.com/olivermt/replicant/commit/a361b9df171014e6b5d8f6655512d46f12743b2d

josevalim:
Can you please send a separate PR with the other appraoch then so we can compare them side by side?

greg-rychlewski:
There is `ALTER SUBSCRIPTION name DISABLE/ENABLE` which I think does what you want?

